PR #4, opened 2025-06-11
Update the example target in the README

eholum:
This is as far as I've gotten, and I'm proposing a method to set the install directory for a shared folder without resorting to ye-olde `get_package_share_directory`.

### DIFF
--- a/bindings/examples/utils.py
+++ b/bindings/examples/utils.py
@@ -12,8 +12,7 @@ def get_example_resources_directory():
     # TODO: Could replace this with the install location but for now it's
     #       assuming this exists in a compiled workspace.
     roboplan_path = get_package_path()
-    for _ in range(5):
-        roboplan_path = roboplan_path.parent
+    roboplan_path = roboplan_path.parent.parent.parent.parent.parent
     return (
         roboplan_path
         / "install"
--- a/roboplan_examples/examples/example_resources.hpp
+++ b/roboplan_examples/examples/example_resources.hpp
@@ -9,7 +9,7 @@ namespace roboplan_examples {
 /**
  * Provides compile time access to the resources install directory.
  */
-inline std::string get_install_prefix() {
+inline std::filesystem::path get_install_prefix() {
   // This would be a lot easier if it were an ament package, instead we use
   // dynamic linking to get the filesystem path of the example executable.
   Dl_info dl_info;
@@ -25,8 +25,8 @@ inline std::string get_install_prefix() {
  * Provides compile time access to the resources shared directory for accessing
  * robot models or other resource files.
  */
-inline std::string get_package_share_dir() {
-  return get_install_prefix() + "/share/roboplan_examples";
+inline std::filesystem::path get_package_share_dir() {
+  return get_install_prefix() / "share" / "roboplan_examples";
 }
 
 } // namespace roboplan_examples
--- a/roboplan_examples/examples/test_ik.cpp
+++ b/roboplan_examples/examples/test_ik.cpp
@@ -9,14 +9,11 @@ using namespace roboplan;
 
 int main(int /*argc*/, char* /*argv*/[]) {
 
-  const std::filesystem::path share_prefix =
-      std::filesystem::path(roboplan_examples::get_package_share_dir());
+  const auto share_prefix = roboplan_examples::get_package_share_dir();
 
   // Set up the scene
-  const std::filesystem::path urdf_path =
-      share_prefix / "ur_robot_model" / "ur5_gripper.urdf";
-  const std::filesystem::path srdf_path =
-      share_prefix / "ur_robot_model" / "ur5_gripper.srdf";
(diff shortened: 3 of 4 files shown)


roboplan_examples/examples/test_scene.cpp
```diff
@@ -1,19 +1,19 @@
 #include <filesystem>
-#include <vector>
-
 #include <roboplan/core/scene.hpp>
+#include <vector>
```
sea-bass, line 2: did your linter move this down? usually it's system headers first, then others

roboplan_examples/CMakeLists.txt
```diff
@@ -8,18 +8,26 @@ endif()
...
 find_package(roboplan REQUIRED)
 
-# Build and install the examples.
+# Build and install the examples, including the sample ur_robot_model.
 add_executable(test_scene examples/test_scene.cpp)
 target_link_libraries(test_scene
     roboplan::roboplan
 )
+target_compile_definitions(test_scene PRIVATE
+    TEST_SCENE_INSTALL_PREFIX="${CMAKE_INSTALL_PREFIX}")
```
sea-bass: sadly this won't really help with the Python bindings...

although... you could actually create a binding for this constant if you import the `test_scene.hpp` header, right?

More importantly, in the eventual future where this is installed via binaries, would this all work out? Or is it better to require users to define an environment variable / run from a specific folder or something?

eholum: Oh goodness yes I just got to the python bindings part and now I have to go read more about CMake because I don't know the answer to this question:

> More importantly, in the eventual future where this is installed via binaries, would this all work out? 

I would curse you for this but I guess it's something to learn...

roboplan_examples/examples/test_scene.cpp
eholum: Uh... yes... though I use clang format... Merits a follow up

roboplan_examples/examples/test_ik.cpp
```diff
@@ -1,23 +1,23 @@
 #include <filesystem>
 #include <vector>
 
+#include <examples/resources.hpp>
```
sea-bass: Can this be made `roboplan_examples` instead?

bindings/examples/utils.py
```diff
@@ -0,0 +1,24 @@
...
+def get_package_path():
+    roboplan_file = Path(roboplan.__file__)
+    return Path(roboplan_file.parent)
+
+
+def get_example_resources_directory():
+    # TODO: Could replace this with the install location but for now it's
+    #       assuming this exists in a compiled workspace.
+    roboplan_path = get_package_path()
+    for _ in range(5):
```
eholum: Hardcoded paths lite*

roboplan_examples/examples/example_resources.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Provides compile time access to the resources install directory.
+ */
+inline std::string get_install_prefix() {
+  // This would be a lot easier if it were an ament package, instead we use
+  // dynamic linking to get the filesystem path of the example executable.
+  Dl_info dl_info;
+  dladdr((void*)get_install_prefix, &dl_info);
+  std::filesystem::path lib_path = dl_info.dli_fname;
```
eholum, line 17: https://stackoverflow.com/questions/1642128/linux-how-to-get-full-name-of-shared-object-just-loaded-from-the-constructor/1647198#1647198

Does it work? Yes. Will it work with apt packages and source builds? Double yes. Is it good? I have no idea.

bindings/examples/utils.py
sea-bass: Oh god. Haha.

Honestly I would prefer explicitly doing `.parent.parent.parent.parent.parent`.

roboplan_examples/examples/example_resources.hpp
```diff
@@ -0,0 +1,32 @@
...
+  // <install_directory>/lib/roboplan_examples/<executable>
+  return lib_path.parent_path().parent_path().parent_path();
+}
+
+/**
+ * Provides compile time access to the resources shared directory for accessing
+ * robot models or other resource files.
+ */
+inline std::string get_package_share_dir() {
+  return get_install_prefix() + "/share/roboplan_examples";
```
sea-bass: Should we make this just return a `std::filesystem::path`? 

Nanobind also binds paths.

roboplan_examples/examples/example_resources.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <dlfcn.h>
+#include <filesystem>
+#include <string>
+
+namespace roboplan_examples {
+
+/**
+ * Provides compile time access to the resources install directory.
+ */
+inline std::string get_install_prefix() {
```
sea-bass: I think it's ok for this to return a `std::filesystem::path` too

PR #5, opened 2025-06-14
Add basic unit testing pipeline

sea-bass:
This PR adds laughably simple unit tests on the C++ and Python side, and a CI pipeline that actually runs them.

I mostly just wanted the pipeline here so that as we start adding tests, they're actually running...

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -36,10 +36,9 @@ jobs:
           file: .docker/Dockerfile
           tags: roboplan:${{ github.head_ref || github.ref_name }}
           push: false
-          outputs: type=docker,dest=./roboplan_image.tar
+          load: true
       - name: Run tests
         run: |
-          docker load --input ./roboplan_image.tar
           docker run \
             roboplan:${{ github.head_ref || github.ref_name }} \
             /bin/bash -c './src/roboplan/scripts/run_tests.bash'


roboplan/test/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+find_package(GTest REQUIRED)
+
+add_executable(
+  test_types
+  test_types.cpp
+)
+target_include_directories(test_types PRIVATE ../include)
```
sea-bass: it feels wrong to have to do this... but somehow I needed it

.docker/Dockerfile
```diff
@@ -19,7 +19,7 @@ COPY . src/roboplan
 RUN source /opt/ros/${ROS_DISTRO}/setup.bash \
     && apt-get update -y \
     && rosdep install --from-paths src --ignore-src --rosdistro ${ROS_DISTRO} -y \
-    && colcon build
+    && colcon build --cmake-args -DBUILD_TESTING=ON
```
sea-bass, line 22: for whatever ungodly reason, I was getting non-deterministic results on this variable being on depending on whether I had previously sourced the workspace. Wild.

eholum: That is dark...

roboplan/test/CMakeLists.txt
eholum: Pretty sure the includes metadata isn't being built correctly for the roboplan package:

https://github.com/sea-bass/roboplan/commit/e9107b7f99d5d108a64f6c5d02f0d628bc91f76b

.github/workflows/build_and_test.yml
```diff
@@ -29,4 +34,12 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           file: .docker/Dockerfile
+          tags: roboplan:latest
```
eholum: I don't actually know if it's possible for there to be two images on the same runner, but it there any reason to not just tag this with the ref name or something so that it never happens?

Or something?

```suggestion
          tags: roboplan:${{ github.ref }}
```

sea-bass: Good idea! It's `ref_name`, apparently.

roboplan/test/CMakeLists.txt
sea-bass: huh, interesting. I'm gonna read up what some of these options are to see if we need *everything* in that commit, or just select parts of it.

sea-bass: Indeed, it's only the `target_include_directories` changes that were needed

.docker/Dockerfile
sea-bass: Could do what Davide set up in BT.CPP?

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/d8f3b37efa24445ec0209904c677c941ba687945/CMakeLists.txt#L12

roboplan/test/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+find_package(GTest REQUIRED)
+
+add_executable(
+  test_types
+  test_types.cpp
+)
+# target_include_directories(test_types PRIVATE ../include)
```
sea-bass:
```suggestion
```

.docker/Dockerfile
eholum: I generally try to do what Davide does without success.

.github/workflows/build_and_test.yml
```diff
@@ -29,4 +34,12 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           file: .docker/Dockerfile
+          tags: roboplan:${{ github.head_ref || github.ref_name }}
           push: false
+          outputs: type=docker,dest=./roboplan_image.tar
+      - name: Run tests
+        run: |
+          docker load --input ./roboplan_image.tar
```
eholum: I didn't think saving an image as a tar file was necessary, you can just load it directly in the runners local daemon and use it in subsequent steps. Saving/loading images is expensive.

```suggestion
          load: true
      - name: Run tests
        run: |
```

https://github.com/sea-bass/roboplan/actions/runs/15681768548/job/44174695686?pr=6#step:5:1

PR #7, opened 2025-06-20
Move models to `roboplan_example_models` package

sea-bass:
This PR moves the example models to a new package named `roboplan_example_models`.

This will allow us to also include headers `roboplan_example_models` for unit testing in the existing C++ packages, which will be necessary in the stuff I'm doing next.

I also made the Python utilities for getting resources path grab the same folder as the C++ equivalent, and also live inside the `roboplan` module itself.

Please check that everything still works!

### DIFF
--- a/roboplan_example_models/src/resources.cpp
+++ b/roboplan_example_models/src/resources.cpp
@@ -4,7 +4,8 @@ namespace roboplan_example_models {
 
 std::filesystem::path get_install_prefix() {
   // This would be a lot easier if it were an ament package, instead we use
-  // dynamic linking to get the filesystem path of the example resources shared object file.
+  // dynamic linking to get the filesystem path of the example resources shared
+  // object file.
   Dl_info dl_info;
   dladdr((void*)get_install_prefix, &dl_info);
   std::filesystem::path lib_path = dl_info.dli_fname;


roboplan_example_models/include/roboplan_example_models/resources.hpp
```diff
@@ -17,16 +17,17 @@ inline std::filesystem::path get_install_prefix() {
   std::filesystem::path lib_path = dl_info.dli_fname;
 
   // Then we can just pull the relative path to the share directory
-  // <install_directory>/lib/roboplan_examples/<executable>
-  return lib_path.parent_path().parent_path().parent_path();
+  // <install_directory>/lib/roboplan_example_models/<executable>
+  return lib_path.parent_path().parent_path().parent_path().parent_path();
```
sea-bass: Interestingly enough, I had to go one level up, because as it existed, this would use the directory from the package you are currently running.

So for example, doing `ros2 run roboplan_examples test_scene`

yielded the incorrect path `$MY_WS/install/roboplan_examples/lib/roboplan_example_models/<executable>`

... this means that this solution likely is not going to work for binaries, or generally building outside the workspace, now that the packages are split up. Sorry.

roboplan_example_models/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+project(roboplan_example_models VERSION 0.0.0)
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# Find dependencies.
+
+# Build and install the library containing the resource paths.
+add_library(roboplan_example_models INTERFACE)
```
eholum: This needs to be an executable for the header to work, otherwise linking will bork.

roboplan_example_models/include/roboplan_example_models/resources.hpp
eholum: Oh ye of little faith!

I actually did this same thing while I was poking on the last PR. Sorry... should have mentioned it:

https://github.com/sea-bass/roboplan/blob/old-examples-package/roboplan_examples/example_resources/CMakeLists.txt#L8

The problem is that you've made it an interface library, so it gets included in whatever executable you add it to, rather than being installed relative the share directory as you want. If you make it it's own library like in the link, it'll load consistently as we would expect.

sea-bass: Ah, good to know -- thank you sir

.pre-commit-config.yaml
```diff
@@ -16,7 +16,6 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
-      - id: check-added-large-files
```
eholum, line 19: No lfs for you?

roboplan_example_models/src/resources.cpp
```diff
@@ -0,0 +1,21 @@
+#include <roboplan_example_models/resources.hpp>
+
+namespace roboplan_example_models {
+
+std::filesystem::path get_install_prefix() {
+  // This would be a lot easier if it were an ament package, instead we use
+  // dynamic linking to get the filesystem path of the example executable.
```
eholum:
```suggestion
  // dynamic linking to get the filesystem path of the example resources shared object file.
```

.pre-commit-config.yaml
sea-bass: The thing about LFS free tier is that it has pretty small limits. We had this on `moveit_drake`, where among 3 of us we maxed out a particular month. So I think not yet until this becomes a problem.

Doing the math, right now the UR models folder is 7 MB, so it would take a bit to exceed the 1 GB / 5 users monthly quota, but still...

PR #10, opened 2025-06-22
Collision checking functionality

sea-bass:
Inching towards an initial RRT implementation and the road is full of ~obstacles~collisions...

Adds basic collision checking functionality, both for individual configurations and along a path.

Pinocchio already has everything one might need here <3

### DIFF
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -107,15 +107,19 @@ bool Scene::hasCollisionsAlongPath(const Eigen::VectorXd& q_start, const Eigen::
   const auto distance = pinocchio::distance(model_, q_start, q_end);
 
   // Special case for short paths (also handles division by zero in the next case).
+  const bool collision_at_endpoints = hasCollisions(q_start) || hasCollisions(q_end);
   if (distance <= min_step_size) {
-    return hasCollisions(q_start) && hasCollisions(q_end);
+    return collision_at_endpoints;
   }
 
+  // In the general case, check the first and last points, then all the intermediate ones.
   const auto num_steps = static_cast<size_t>(std::ceil(distance / min_step_size)) + 1;
-  for (size_t idx = 0; idx <= num_steps; ++idx) {
+  if (collision_at_endpoints) {
+    return true;
+  }
+  for (size_t idx = 1; idx <= num_steps - 1; ++idx) {
     const auto fraction = static_cast<double>(idx) / static_cast<double>(num_steps);
-    const auto q_interp = pinocchio::interpolate(model_, q_start, q_end, fraction);
-    if (hasCollisions(q_interp)) {
+    if (hasCollisions(pinocchio::interpolate(model_, q_start, q_end, fraction))) {
       return true;
     }
   }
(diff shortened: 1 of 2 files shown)


roboplan/src/core/scene.cpp
```diff
@@ -90,14 +88,40 @@ Eigen::VectorXd Scene::randomPositions() {
...
+}
+
+bool Scene::hasCollisionsAlongPath(const Eigen::VectorXd& q_start, const Eigen::VectorXd& q_end,
+                                   const double min_step_size) {
+
+  const auto distance = pinocchio::distance(model_, q_start, q_end);
+
+  // Special case for short paths (also handles division by zero in the next case).
+  if (distance <= min_step_size) {
+    return hasCollisions(q_start) && hasCollisions(q_end);
```
eholum: Bug!!!

```suggestion
    return hasCollisions(q_start) || hasCollisions(q_end);
```

roboplan/src/core/scene.cpp
```diff
@@ -90,14 +88,40 @@ Eigen::VectorXd Scene::randomPositions() {
...
+  }
+
+  const auto num_steps = static_cast<size_t>(std::ceil(distance / min_step_size)) + 1;
+  for (size_t idx = 0; idx <= num_steps; ++idx) {
+    const auto fraction = static_cast<double>(idx) / static_cast<double>(num_steps);
+    const auto q_interp = pinocchio::interpolate(model_, q_start, q_end, fraction);
```
eholum: I would pre-allocate a `q_interp` and use the other [interpolate](https://github.com/stack-of-tasks/pinocchio/blob/master/include/pinocchio/algorithm/joint-configuration.hpp#L101), just so that you're not dynamically allocating a vector each iteration. Probably just something to generally watch out for while mucking around at the bottom layer...

roboplan/src/core/scene.cpp
```diff
@@ -90,14 +88,40 @@ Eigen::VectorXd Scene::randomPositions() {
...
   return positions;
 }
 
+bool Scene::hasCollisions(const Eigen::VectorXd& q) {
+  return pinocchio::computeCollisions(model_, model_data_, collision_model_, collision_model_data_,
+                                      q,
+                                      /* stop_at_first_collision*/ true);
+}
+
+bool Scene::hasCollisionsAlongPath(const Eigen::VectorXd& q_start, const Eigen::VectorXd& q_end,
```
eholum, line 104: Any reason to sanity check sizes on q_start or q_end?

sea-bass: Definitely a reason to check! This feeds nicely into the `tl::expected` discussion... because what should we do in the case of invalid sizes here?

I'd like to figure that out first and then come back to add error handling to all these functions, if that's okay.

roboplan/src/core/scene.cpp
sea-bass: RIP, nice one

roboplan/src/core/scene.cpp
sea-bass: I recently had a whole discussion at work about return value optimization being good enough now that outparams are "old-school".

But now I want to use this use case to benchmark the difference for a super tiny step size and/or long path

sea-bass: I just benchmarked both options and the time is pretty much the same.

eholum: > I recently had a whole discussion at work about return value optimization being good enough now that outparams are "old-school".

Sounds like someone out there knows more about compilers than yours truly. Honestly hard to believe.

roboplan/src/core/scene.cpp
eholum: It is "ok"!

PR #11, opened 2025-06-22
Test all active ROS distros

sea-bass:
As the title says.

Kilted is still on Pinocchio 3.4.0 (vs. others on 3.5.0) which requires one more explicit header include. Probably good to have regardless.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,3 +2,4 @@ Language: Cpp
 DerivePointerAlignment: false
 PointerAlignment: Left
 ColumnLimit: 100
+SpacesBeforeTrailingComments: 2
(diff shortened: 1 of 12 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -24,6 +24,10 @@ concurrency:
 jobs:
   # Build docker images
   build_and_test:
+    strategy:
+      matrix:
+        ros_distro: [humble, jazzy, kilted, rolling]
```
eholum, line 29: So it begins.

sea-bass: to be fair, it was still using ROS before, but it was only on Jazzy

PR #12, opened 2025-06-23
Add vanilla CMake workflow

sea-bass:
This PR will age me 5 years, but it's necessary. Already found some things that were tailored specifically to ROS workflows!

I bet I am doing everything wrong, but this is how you learn.

### DIFF
--- a/.docker/ubuntu/Dockerfile
+++ b/.docker/ubuntu/Dockerfile
@@ -2,39 +2,42 @@ ARG UBUNTU_VERSION=24.04
 
 FROM ubuntu:${UBUNTU_VERSION} AS roboplan_ubuntu
 
+ARG PINOCCHIO_VERSION=3.7.0
 ENV PIP_BREAK_SYSTEM_PACKAGES=1
 SHELL ["/bin/bash", "-c"]
 
 # Install core dependencies.
 RUN apt-get update -y \
     && apt install -y \
-        cmake curl libeigen3-dev libgmock-dev lsb-release python3-pip
+        cmake libeigen3-dev libgmock-dev liburdfdom-dev python3-pip
 RUN pip3 install nanobind pytest scikit-build-core
 
-# Install Pinocchio binaries.
-# https://stack-of-tasks.github.io/pinocchio/download.html
-RUN mkdir -p /etc/apt/keyrings && \
-    curl http://robotpkg.openrobots.org/packages/debian/robotpkg.asc \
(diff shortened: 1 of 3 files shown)


roboplan_example_models/src/resources.cpp
```diff
@@ -1,5 +1,5 @@
+#include <iostream>
```
sea-bass: todo: remove

scripts/build_cmake.bash
```diff
@@ -0,0 +1,45 @@
...
+export PKG_CONFIG_PATH=/opt/openrobots/lib/pkgconfig:$PKG_CONFIG_PATH
+export LD_LIBRARY_PATH=/opt/openrobots/lib:$LD_LIBRARY_PATH
+export CMAKE_PREFIX_PATH=/opt/openrobots:$CMAKE_PREFIX_PATH
+
+# Build all the packages with CMake
+# rm -rf build install  # If you want a clean build
+mkdir -p build install
+export CMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}:${PWD}/install/
+
+# TODO: Make into reusable function?
```
sea-bass, line 21: also need to make sure the exit codes are reflected properly, doesn't seem to be the case at the moment

.docker/ubuntu/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    && apt install -y \
+        cmake curl libeigen3-dev libgmock-dev lsb-release python3-pip
+RUN pip3 install nanobind pytest scikit-build-core
+
+# Install Pinocchio binaries.
+# https://stack-of-tasks.github.io/pinocchio/download.html
+RUN mkdir -p /etc/apt/keyrings && \
+    curl http://robotpkg.openrobots.org/packages/debian/robotpkg.asc \
+     | tee /etc/apt/keyrings/robotpkg.asc > /dev/null && \
+     echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/robotpkg.asc] http://robotpkg.openrobots.org/packages/debian/pub $(lsb_release -cs) robotpkg" \
```
eholum: I hate to be a Mac shill guy but everything sucks and my day is ruined, no arm binaries.

http://robotpkg.openrobots.org/packages/debian/pub/dists/noble/

sea-bass: I think this has since been replaced with your PR to use pip instead... there is no `robotpkg` on main right now!

PR #15, opened 2025-06-24
Use pip to install pinocchio and its dependencies

eholum:
because arm64.

There is definitely ugliness here. Is it better or worse? I guess better in the sense that it works on my machine, worse in the sense that we use pip installed versions of Pinocchio deps in the build (is that bad?).


.docker/ubuntu/Dockerfile
```diff
@@ -8,34 +8,56 @@ SHELL ["/bin/bash", "-c"]
 # Install core dependencies.
 RUN apt-get update -y \
     && apt install -y \
-        cmake curl libeigen3-dev libgmock-dev lsb-release python3-pip
-RUN pip3 install nanobind pytest scikit-build-core
+        cmake curl libeigen3-dev libgmock-dev lsb-release python3-pip wget
```
sea-bass: I know this is really annoying, but I have a preference for not having to install both curl and wget if they both do the same?

.docker/ubuntu/entrypoint.sh
sea-bass, line 1: If this is gone in favor of `ldconfig` we can just do away with the entrypoint altogether... but I guess it may be nice to keep it around for future modifications?

.dockerignore
```diff
@@ -1,3 +1,7 @@
+.git/
 .github
+.gitignore
 **/build/
 **/install/
+**/Dockerfile*
```
sea-bass: I don't think you need the * at the end there since both files are named just `Dockerfile`.

But also, thank you.

.docker/ubuntu/Dockerfile
```diff
@@ -8,34 +8,38 @@ SHELL ["/bin/bash", "-c"]
...
-RUN mkdir -p /etc/apt/keyrings && \
-    curl http://robotpkg.openrobots.org/packages/debian/robotpkg.asc \
-     | tee /etc/apt/keyrings/robotpkg.asc > /dev/null && \
-     echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/robotpkg.asc] http://robotpkg.openrobots.org/packages/debian/pub $(lsb_release -cs) robotpkg" \
-     | tee /etc/apt/sources.list.d/robotpkg.list && \
-    apt-get update -y && \
-    apt install -qqy robotpkg-py3*-pinocchio
-ENV CMAKE_PREFIX_PATH="/opt/openrobots/:${CMAKE_PREFIX_PATH}"
-ENV LD_LIBRARY_PATH="/opt/openrobots/lib:${LD_LIBRARY_PATH}"
+# Install pinocchio with pip as it support additional CPU architectures
```
sea-bass: This is me weaseling into your pr

```suggestion
# Install pinocchio with pip as it supports additional CPU architectures
```

.docker/ubuntu/Dockerfile
```diff
@@ -8,34 +8,38 @@ SHELL ["/bin/bash", "-c"]
...
+RUN pip3 install pin
```
sea-bass: (optional) can you also make a variable here for the pinocchio version?

so it's `pip3 install pin==3.6.0` for example?

I noticed that it was installing pinocchio 3.7.0 here, which is different from the one available in ROS at the moment (it's in 3.4.0 / 3.5.0, and 3.6.0 is around the corner)

.docker/ubuntu/Dockerfile
```diff
@@ -8,34 +8,38 @@ SHELL ["/bin/bash", "-c"]
...
-RUN cd bindings && \
+RUN . /roboplan_env.sh && \
+    cd bindings && \
     pip3 install --no-build-isolation -ve .
 
+# Add all installed packages' lib dirs to ldconfig
+RUN for PACKAGE in /roboplan_ws/install/*/lib; do \
+        echo ${PACKAGE} >> /etc/ld.so.conf.d/roboplan.conf; \
+    done && \
+    ldconfig
```
sea-bass: if you do this here, then this will only kick in for the first entrypoint on starting the service, but not if you shell in via `docker compose exec -it <service> bash`.

So I'd prefer putting this in the entrypoint for that reason.

EDIT: Same with the `PIP_PATH` and `CMAKE_PREFIX_PATH` above

.docker/ubuntu/Dockerfile
eholum: Oh my lord hopefully we're good with `3.4.0` for now...

```
5.419 ERROR: Could not find a version that satisfies the requirement pin== (from versions: 2.6.21, 2.7.0, 3.3.1, 3.4.0, 3.7.0)
```

.docker/ubuntu/Dockerfile
eholum: Oh reallllllly. Theoretically the ld cache should be part of the image once that runs... That said the cmake paths certainly aren't so I'll just revert that change and add a source to the entry point...

sea-bass: Maybe just those env vars need to change then, that's fine!

.docker/ubuntu/Dockerfile
sea-bass: lol. Cool. Just leave it at 3.7.0 then. The changes from 3.4 to 3.7 are quite minor

.docker/ubuntu/Dockerfile
eholum: I clearly don't understand how any of this works but at least it's all wrapped up in a nice neat little package. Plus I'm sure the `pip3 show pin` magic grep is robust to any and all future changes.

PR #16, opened 2025-06-24
First vanilla RRT implementation with dynotree

sea-bass:
Adds an initial implementation of "vanilla" RRT using the [`dynotree`](https://github.com/quimortiz/dynotree) library for nearest-neighbors lookup. The RRT itself is also highly inspired by the related [`dynoRRT`](https://github.com/quimortiz/dynoRRT/tree/main/include/dynoRRT) repo.

I just "vendored" the library by pasting it into this repo because it's easiest ... and credit has been given.

Don't be intimidated by the PR size! The bulk of it is copying in the `dynotree` headers into the repo.

Next up after this PR I'm going to create a visualization using Viser to see if this thing is actually working as intended.

### DIFF
--- a/roboplan_rrt/src/rrt.cpp
+++ b/roboplan_rrt/src/rrt.cpp
@@ -40,16 +40,25 @@ std::optional<JointPath> RRT::plan(const JointConfiguration& start,
   const auto& q_start = start.positions;
   const auto& q_goal = goal.positions;
 
-  kd_tree_.addPoint(q_start, 0);
-  nodes_.emplace_back(q_start, -1);
-
+  // Check whether direct connection between the start and goal are possible.
   if ((scene_->configurationDistance(q_start, q_goal) <= options_.max_connection_distance) &&
       (!scene_->hasCollisionsAlongPath(q_start, q_goal, options_.collision_check_step_size))) {
     std::cout << "Can directly connect start and goal!\n";
     return JointPath{.joint_names = scene_->getJointNames(), .positions = {q_start, q_goal}};
   }
 
+  // Initialize the trees for searching.
+  kd_tree_.addPoint(q_start, 0);
+  nodes_.emplace_back(q_start, -1);
+
   while (true) {
+    // Check loop termination criteria.
+    // TODO: Also check for max elapsed time in future?
+    if (nodes_.size() >= options_.max_nodes) {
+      std::cout << "Added maximum number of nodes (" << options_.max_nodes << ").\n";
+      break;
+    }
+
     // Sample the next node.
     const auto q_sample = (uniform_dist_(rng_gen_) <= options_.goal_biasing_probability)
                               ? q_goal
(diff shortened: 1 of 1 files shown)


bindings/CMakeLists.txt
```diff
@@ -41,9 +42,11 @@ nanobind_add_module(
   NB_STATIC
   src/roboplan_ext.cpp
 )
+set_property(TARGET roboplan PROPERTY CXX_STANDARD 20)
```
sea-bass, line 45: I did it all for the `std::map::contains()`.

roboplan_rrt/src/rrt.cpp
```diff
@@ -0,0 +1,119 @@
+#include <stdexcept>
+
+#include <roboplan_rrt/rrt.hpp>
+
+namespace roboplan {
+
+RRT::RRT(const std::shared_ptr<Scene> scene, const RRTOptions& options)
```
eholum, line 7: Discussion/food for thought topic, since some things that bothered me about a certain planning library were the giant monolithic structures that were really hard to refactor, break apart, or even mock. Do we want the scene to be at the root of all planners? Will we keep adding functions a la `PlanningScene` to make sure that new planners can get the information they need? Could we someday swap the scene with an abstraction without breaking everything?

Just thinking about it, and shamelessly stealing ideas from this [library](https://github.com/evenfurther/pathfinding/blob/main/src/directed/astar.rs#L87), one alternative is going with a more "functional" approach, wherein the planners don't actually have insight into the whole scene. What if a "context" or something was given that had a the necessary functions? Something like,

```
struct RRTContext {
  std::function<bool(const JointConfiguration&)> is_valid_pose;
  std::function<JointConfiguration()> sample;
  std::function<double(const JointConfiguration&, const JointConfiguration&)> distance;
  std::function<JointConfiguration(const JointConfiguration&, const JointConfiguration&, double)> extend;
}
```

Completely open to ideas here, and not saying this approach is wrong or bad, just want to toss the thought into the aether.

roboplan_rrt/src/rrt.cpp
```diff
@@ -0,0 +1,119 @@
...
+  nodes_.clear();
+  nodes_.reserve(options_.max_nodes);
+
+  const auto& q_start = start.positions;
+  const auto& q_goal = goal.positions;
+
+  kd_tree_.addPoint(q_start, 0);
+  nodes_.emplace_back(q_start, -1);
+
+  if ((scene_->configurationDistance(q_start, q_goal) <= options_.max_connection_distance) &&
```
eholum, line 44: Maybe put this check first before initializing the tree and stuff?

roboplan_rrt/src/rrt.cpp
```diff
@@ -0,0 +1,119 @@
...
+    // Extend from the nearest neighbor to max connection distance.
+    // TODO: In the case of RRT-Connect, we will keep extending until we cannot any longer.
+    // The collision checking will likely need to be inside this function in that case.
+    const auto nn = kd_tree_.search(q_sample);
+    const auto q_extend =
+        extend(nodes_.at(nn.id).config, q_sample, options_.max_connection_distance);
+
+    // Check that the extended node can be added to the tree, and if so whether it can directly
+    // connect to the goal node.
+    if (!scene_->hasCollisionsAlongPath(q_start, q_extend, options_.collision_check_step_size)) {
```
eholum: I don't _think_ you want to connect it all to start? Should be checking from the nearest neighbor rather than `q_start`.

roboplan_rrt/src/rrt.cpp
```diff
@@ -0,0 +1,119 @@
...
+          }
+          cur_node = nodes_.at(cur_idx);
+          path.positions.push_back(cur_node.config);
+        }
+        std::reverse(path.positions.begin(), path.positions.end());
+        return path;
+      }
+    }
+
+    if (nodes_.size() >= options_.max_nodes) {
```
eholum: Ha dumb edge case but maybe put this check before the `q_extend -> q_goal` check?

roboplan_rrt/src/rrt.cpp
sea-bass: I feel it would make it less accessible to have loose functions rather than a scene. Rust is cool, but it's not a beginner friendly language.

One saving grace here is that not every bespoke planner/solver/whatever needs to add more methods to the scene object. Because you can directly extract the underlying Pinocchio model, my preference would be to point people to that API when doing very specific things that would make the scene design clunky.

roboplan_rrt/src/rrt.cpp
sea-bass: oh wow yeah, this is 100% a bug. My bad.

roboplan_rrt/src/rrt.cpp
eholum: Are you saying you don't want to commit to a completely generic fully extensible robot motion planning platform that is also easy to use? Ridiculous! But yeah having just spent an hour debugging function signatures I hear you. 

Oh didn't know that about Pinocchio. That is super handy. I think just acknowledging that the scene is not a dumping ground of functions is saving grace enough for me!

PR #23, opened 2025-07-05
Respect Joint Limits for Start and Goal Poses and Add Planning Timeout for RRT

eholum:
Added the planning timeout test and noticed that RRT would plan to impossible positions that didn't respect joint limits, so I added a function to the `Scene` to check for valid joint poses, then added that check before added an extended point in RRT.

Worth it?

### DIFF
--- a/bindings/src/roboplan_ext.cpp
+++ b/bindings/src/roboplan_ext.cpp
@@ -83,6 +83,7 @@ NB_MODULE(roboplan, m) {
            "max_samples"_a = 1000)
       .def("hasCollisions", &Scene::hasCollisions)
       .def("hasCollisionsAlongPath", &Scene::hasCollisionsAlongPath)
+      .def("isValidPose", &Scene::isValidPose)
       .def("__repr__", [](const Scene& scene) {
         std::stringstream ss;
         ss << scene;


roboplan/src/core/scene.cpp
```diff
@@ -140,6 +140,22 @@ bool Scene::hasCollisionsAlongPath(const Eigen::VectorXd& q_start, const Eigen::
   return false;
 }
 
+bool Scene::isValidPose(const Eigen::VectorXd& q) {
+  size_t q_idx = 0;
+  for (const auto& joint_name : joint_names_) {
```
eholum, line 145: Don't have a lot of time now but I think we could vectorize this and the random sampling functions to reduce duplication...

roboplan_rrt/src/rrt.cpp
```diff
@@ -51,9 +52,19 @@ std::optional<JointPath> RRT::plan(const JointConfiguration& start,
...
+  // Record the start for measuring timeouts
+  const auto start_time = std::chrono::steady_clock::now();
+
   while (true) {
+    // Check for timeout
+    auto elapsed =
+        std::chrono::duration<double>(std::chrono::steady_clock::now() - start_time).count();
+    if (options_.max_planning_time > 0 && options_.max_planning_time <= elapsed) {
+      std::cout << "RRT timed out after " << options_.max_planning_time << " seconds.\n";
+      break;
```
eholum, line 70: `tl::expected` intensifies.

PR #25, opened 2025-07-10
Visualize RRTs with Viser

### LINKED ISSUE #18: Figure out scene visualization with Viser
For now, we'll just assume that the URDF is the "source of truth", until Pinocchio 4.x is released with its full glorious `nanobind` support we can bring in here.

This does mean that we'll need to manage the external obstacles separately in Viser and in RoboPlan, but... fine.

For ROS workflows there will be an alternative to blast RViz with MarkerArray messages.

---

EDIT: Proof

Image

sea-bass:
This PR incorporates my Viser visualizer for Pinocchio from https://github.com/stack-of-tasks/pinocchio/pull/2718 (for now, just copy pasted in), and hacks out a minimally working visualization for RRT. 

Note that we need to add obstacles for these RRTs to look even remotely interesting, so this was me reducing the max connection distance a bunch and retrying until I didn't just get a straight line path.

image

To get this example to work, you will need to `pip install pin` (or equivalent).

Closes #18

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -25,6 +25,7 @@ jobs:
 
   ubuntu_build_and_test:
     strategy:
+      fail-fast: false
       matrix:
         ubuntu_version: [22.04, 24.04]
 
@@ -52,6 +53,7 @@ jobs:
   # Build ROS docker images and run tests.
   ros_build_and_test:
     strategy:
+      fail-fast: false
       matrix:
         ros_distro: [humble, jazzy, kilted, rolling]
 


bindings/examples/example_rrt.py
```diff
@@ -36,3 +49,83 @@
...
+        dist = scene.configurationDistance(q_start, q_end)
+        num_pts = math.ceil(dist / step_size) + 1
+        for p in range(num_pts):
+            pos_start = scene.forwardKinematics(
+                scene.interpolate(q_start, q_end, p / num_pts), "tool0"
+            )[:3, 3]
+            pos_end = scene.forwardKinematics(
+                scene.interpolate(q_start, q_end, (p + 1) / num_pts), "tool0"
+            )[:3, 3]
+            start_segments.append([pos_start, pos_end])
```
sea-bass: this whole thing should probably just be a helper function on the C++ side

roboplan_rrt/include/roboplan_rrt/rrt.hpp
```diff
@@ -60,15 +61,15 @@ class RRT {
   /// @param nodes Reference to the nodes vector.
   /// @param q_init The first node to add to the tree.
   /// @param max_size The maximum size of the tree.
-  void initialize_tree(KdTree& tree, std::vector<Node>& nodes, const Eigen::VectorXd& q_init,
-                       size_t max_size = 1000);
+  void initializeTree(KdTree& tree, std::vector<Node>& nodes, const Eigen::VectorXd& q_init,
```
eholum, line 64: camelCase! That's my b.

bindings/src/roboplan/viser_visualizer.py
```diff
@@ -0,0 +1,328 @@
...
+        copy_models=False,
+        data=None,
+        collision_data=None,
+        visual_data=None,
+    ):
+        if not import_viser_succeed:
+            msg = (
+                "Error while importing the viewer client.\n"
+                "Check whether viser is properly installed "
+                "(pip install --user viser)."
```
eholum, line 47: Oooh. Needs to be installed in `--user` space, because of the numpy Debian install?

```
Is there a trick to installing viser in the containers? I have a deb install of numpy that pip3 doesn't like,

```
25.73   Attempting uninstall: numpy
25.74     Found existing installation: numpy 1.26.4
25.74 ERROR: Cannot uninstall numpy 1.26.4, RECORD file not found. Hint: The package was installed by debian.
```

Should add that the Dockerfiles though.

File: bindings/examples/example_rrt.py [hunk 4]
```diff
@@ -1,12 +1,16 @@
+import math
 from pathlib import Path
 
+import numpy as np
+import pinocchio as pin
```
eholum, line 5: Ah man as noted below maybe some more build shenanigans? This is in a fresh ros container after installing viser,

```
python3 examples/example_rrt.py 

A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.2.6 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "/workspace/roboplan_ws/src/roboplan/bindings/examples/example_rrt.py", line 5, in <module>
    import pinocchio as pin
  File "/opt/ros/jazzy/lib/python3.12/site-packages/pinocchio/__init__.py", line 19, in <module>
    from .pinocchio_pywrap_default import *
  File "/root/.local/lib/python3.12/site-packages/numpy/core/_multiarray_umath.py", line 44, in __getattr__
    raise ImportError(msg)
```

bindings/examples/example_rrt.py
eholum: Yes. Scene/Graph/Path Utils?

bindings/src/roboplan/viser_visualizer.py
sea-bass: You shouldn't have to install in user space, that's just what the warning says.

bindings/examples/example_rrt.py
sea-bass: Interesting. I may need to try the fresh Docker as it just worked ™️  for me. To qualify, I'm on 24.04 with Jazzy on my host.

bindings/src/roboplan/viser_visualizer.py
eholum: Oh... not installing in the user space yells at me with the error above... I call shenanigans!

bindings/examples/example_rrt.py
eholum: To be clear, it's definitely possible that I messed something up.

sea-bass: Wait, maybe installing viser upgraded your numpy to >=2.0 and you just need to rebuild the full Python bindings from scratch now?

sea-bass: I was able to repro this issue in the container, because there is a system python installed with version 1.26.4.

After a lot of disgusting debian installed Python shenanigans, I think I arrived at something that works. Sadly it still requires holding NumPy back to below 2.x because apt/rosdep don't let us have nice things.

sea-bass: ... and of course this kills the non-ROS workflow. Ugh, I hate ROS.

sea-bass: so yea tl;dr the ROS debians of Pinocchio are compiled with NumPy 1.x and when you get outside the ROS ecosystem and `pip install pin`, those are with NumPy 2.x.

Instead of having a hard requirement for the bindings, the only way to make this work is to force Python to have the right (read: matching whatever Pinocchio on that system has) NumPy version before and after compiling the bindings.

sea-bass: One more tip, the viser import error you see is actually a try/catch swallowing an OpenCV import error. (to verify, just try doing the `import viser` yourself in a REPL).

You can fix this with `pip3 install opencv-contrib-python-headless`

sea-bass: Anyways, I think I've managed to hack around it in the Dockerfiles. It is... not pretty, but it works.

* In the ROS container you have to hack to force the system to stay in NumPy<2.0, and by doing so you have to manually uninstall the debian installed `typing_extensions` due to https://github.com/brentyi/tyro/issues/324
* In the Ubuntu container, you can just stay with Numpy 2.x as the Lord intended. However, I couldn't just add these as dependencies to the `pyproject.toml` because it would break the ROS workflow above

bindings/src/roboplan/viser_visualizer.py
sea-bass: See my rants here for the root cause: https://github.com/sea-bass/roboplan/pull/25#discussion_r2198252877

bindings/examples/example_rrt.py
eholum: Oh man I guess Pinocchio/ros/viser users are going to run into the same issues with your upstream viser implementation? I was hoping they had some sort of magic fix.

In any case... ROS.

bindings/examples/example_rrt.py
```diff
@@ -1,12 +1,79 @@
...
+    viz: ViserVisualizer,
+    scene: Scene,
+    rrt: RRT,
+    path: JointPath,
+    frame_name: str,
+    max_step_size: float,
+) -> None:
+    """
+    Helper function to visualize the RRT path.
+    TODO: Move this to the actual Python package itself.
```
eholum, line 28: Which python package? roboplan_viz or something?

bindings/src/roboplan/viser_visualizer.py
```diff
@@ -0,0 +1,334 @@
+"""
+TEMPORARY FILE: Can be removed when this PR is landed and released.
+https://github.com/stack-of-tasks/pinocchio/pull/2718
```
eholum, line 3: 🚀🚀🚀

roboplan/src/core/path_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+  const auto distance = scene.configurationDistance(q_start, q_end);
+  const auto num_steps = static_cast<size_t>(std::ceil(distance / max_step_size)) + 1;
+
+  std::vector<Eigen::Matrix4d> frame_path;
+  frame_path.reserve(num_steps);
+
+  for (size_t idx = 0; idx <= num_steps; ++idx) {
+    const auto fraction = static_cast<double>(idx) / static_cast<double>(num_steps);
+    const auto q_interp = scene.interpolate(q_start, q_end, fraction);
+    frame_path.push_back(scene.forwardKinematics(q_interp, frame_name));
```
eholum, line 19: Out of curiosity does `framesForwardKinematics` always use the whatever link is at the root of the URDF?

bindings/examples/example_rrt.py
sea-bass: I think it can just be a manually written module within the existing `roboplan` package!

roboplan/src/core/path_utils.cpp
sea-bass: It will use the universe joint in Pinocchio I believe.

But you can just grab the transform of a different base frame and do `world_T_base.inverse() * world_T_tip`.

This is how you would implement base frame support and is something we should add to our FK wrapper.

bindings/examples/example_rrt.py
sea-bass: Between the ROS user and the non-ROS user suffering, my thumb is on ROS

PR #32, opened 2025-07-18
Initial pixi support

JafarAbdi:
Close https://github.com/sea-bass/roboplan/issues/28

See my comments here https://github.com/sea-bass/roboplan/pull/31 :D:D


.clangd
```diff
@@ -0,0 +1,30 @@
+CompileFlags:
```
JafarAbdi, line 1: Should we keep this? Happy to delete it, just to make clangd happy by default

sea-bass: This is fine!

PR #33, opened 2025-07-21
Add path shortening utils and consolidate examples

eholum:
Resolves https://github.com/sea-bass/roboplan/issues/19.

Shamelessly converting https://github.com/sea-bass/pyroboplan/blob/main/src/pyroboplan/planning/path_shortcutting.py.

The quality of the example is highly dependent on the original path, but I suppose that's to be expected. Here's a good run:

image

### DIFF
--- a/.docker/ros/Dockerfile
+++ b/.docker/ros/Dockerfile
@@ -8,7 +8,7 @@ SHELL ["/bin/bash", "-c"]
 # Install core dependencies.
 RUN apt-get update -y \
     && apt install -y libgmock-dev python3-pip
-RUN pip3 install typing_extensions==4.10.0 nanobind scikit-build-core tyro
+RUN pip3 install nanobind scikit-build-core
 
 # Create a ROS 2 workspace and copy in the source code.
 RUN mkdir -p /workspace/roboplan_ws/src/roboplan
@@ -28,7 +28,7 @@ RUN source /opt/ros/${ROS_DISTRO}/setup.bash \
 # As/if this package matures, a possible solution is to apply patches for ROS release.
 WORKDIR /workspace/roboplan_ws/src/roboplan/bindings
 RUN source /opt/ros/${ROS_DISTRO}/setup.bash \
-    && pip3 install "numpy<2.0.0" typing_extensions==4.10.0 viser opencv-contrib-python-headless \
+    && pip3 install "numpy<2.0.0" typing_extensions==4.10.0 viser opencv-contrib-python-headless tyro \
     && source /workspace/roboplan_ws/install/setup.bash \
     && pip3 install --no-build-isolation -ve .
 


roboplan/src/core/path_utils.cpp
```diff
@@ -47,4 +47,113 @@ bool hasCollisionsAlongPath(const Scene& scene, const Eigen::VectorXd& q_start,
...
+  return shortened_path;
+}
+
+std::vector<double> getNormalizedPathScaling(const Scene& scene, const JointPath& path) {
+  if (path.positions.empty()) {
+    return {};
+  }
+  if (path.positions.size() == 1) {
+    return {1.0};
+  }
```
sea-bass: I would argue that size < 2 should be an error, and that the other special case would be size = 2 simply returning the vector `{0.0, 1.0}`.

roboplan/src/core/path_utils.cpp
```diff
@@ -47,4 +47,113 @@ bool hasCollisionsAlongPath(const Scene& scene, const Eigen::VectorXd& q_start,
...
+  path_length_list.push_back(path_length);
+  for (size_t idx = 0; idx < path.positions.size() - 1; ++idx) {
+    path_length += scene.configurationDistance(path.positions[idx], path.positions[idx + 1]);
+    path_length_list.push_back(path_length);
+  }
+
+  // Normalize
+  for (auto& length : path_length_list) {
+    length /= path_length;
+  }
```
sea-bass: if you make the return type an `Eigen::VectorXd` then the normalization would be even easier.

roboplan/src/core/path_utils.cpp
```diff
@@ -47,4 +47,113 @@ bool hasCollisionsAlongPath(const Scene& scene, const Eigen::VectorXd& q_start,
...
+  JointPath shortened_path = path;
+
+  // Cannot shorten paths if they don't have enough points.
+  if (path.positions.size() < 3) {
+    return shortened_path;
+  }
+
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_real_distribution<double> dis(0.0, 1.0);
```
sea-bass: needed for this PR / deterministic testing: ability to seed this with a specific value

roboplan/src/core/path_utils.cpp
```diff
@@ -47,4 +47,113 @@ bool hasCollisionsAlongPath(const Scene& scene, const Eigen::VectorXd& q_start,
...
 
+JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_step_size,
+                       unsigned int max_iters) {
+  // Make a copy of the provided path.
+  JointPath shortened_path = path;
+
+  // Cannot shorten paths if they don't have enough points.
+  if (path.positions.size() < 3) {
+    return shortened_path;
+  }
```
sea-bass: I think you can prob just remove this if it's already in the loop. Never mind the cost of instantiating those few variables.

roboplan/src/core/path_utils.cpp
```diff
@@ -47,4 +47,113 @@ bool hasCollisionsAlongPath(const Scene& scene, const Eigen::VectorXd& q_start,
...
+
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_real_distribution<double> dis(0.0, 1.0);
+
+  for (unsigned int i = 0; i < max_iters; ++i) {
+
+    // The the path is at maximum shortcutted-ness
```
sea-bass:
```suggestion
    // The path is at maximum shortcutted-ness
```

roboplan/src/core/path_utils.cpp
```diff
@@ -47,4 +47,113 @@ bool hasCollisionsAlongPath(const Scene& scene, const Eigen::VectorXd& q_start,
...
+
+      new_positions.push_back(q_low);
+      new_positions.push_back(q_high);
+
+      for (size_t i = idx_high; i < shortened_path.positions.size(); ++i) {
+        new_positions.push_back(shortened_path.positions[i]);
+      }
+
+      shortened_path.positions = std::move(new_positions);
+    }
```
sea-bass: if is better to combine https://cplusplus.com/reference/vector/vector/erase/ and https://cplusplus.com/reference/vector/vector/insert/ instead of making a new vector?

Maybe worth doing some light benchmarking here, especially on longer paths.

eholum: Yeah the copying and allocating are a little out of control. I'll get a baseline in and then clean it up...

roboplan/src/core/path_utils.cpp
eholum: Where did we land on the `tl::expected` page?

sea-bass: I'm up for it, but as a separate PR

bindings/src/roboplan/viser_visualizer.py
```diff
@@ -65,7 +65,7 @@ def initViewer(
         open=False,
         loadModel=False,
         host="localhost",
-        port="8000",
+        port="8080",
```
eholum:
```suggestion
        port="8000",
```

bindings/src/roboplan/viser_visualizer.py
```diff
@@ -88,14 +88,24 @@ def initViewer(
 
         if open:
             import webbrowser
+            import threading
+
+            client_connected = threading.Event()
+
+            # Set up callback for when a client connects
+            @self.viewer.on_client_connect
```
eholum, line 98: Sorry to be that guy. This just times out because it'll open a browser on my Mac instead of the machine I'm ssh'd into with VS code.

Following the example from their [docs](https://github.com/nerfstudio-project/viser/blob/4a239d0eb3fc73bc10b6c41d26e48c22b4dca595/docs/source/examples/gui/modals.rst#L33)?

bindings/src/roboplan_ext.cpp
```diff
@@ -96,6 +96,12 @@ NB_MODULE(roboplan, m) {
 
   m_core.def("computeFramePath", &computeFramePath);
   m_core.def("hasCollisionsAlongPath", &hasCollisionsAlongPath);
+  m_core.def("shortcutPath", &shortcutPath, "scene"_a, "path"_a, "max_step_size"_a,
```
eholum, line 99: The fact that nanobind just supports defaults is #DopeAF

bindings/src/roboplan/viser_visualizer.py
sea-bass: Note this will go away when we use the upstream visualizer. Instead of changing the default, you're better off changing the example to specify the port 8000 as a non-default arg

bindings/src/roboplan_ext.cpp
sea-bass: Yes. We need to go through and add more of these

bindings/src/roboplan/viser_visualizer.py
eholum: Oh yeah sorry I changed and then changed it back... was just playing around since I port forward from my host to my vm...

bindings/examples/example_path_shortcutting.py
```diff
@@ -0,0 +1,96 @@
...
+    JointConfiguration,
+    JointPath,
+    Scene,
+    RRTOptions,
+    RRT,
+)
+from roboplan.viser_visualizer import ViserVisualizer
+
+
+def visualizePath(
```
sea-bass: So once the Franka PR lands, I think this is what we'll need:

* Factor out the existing `visualizePath` to two separate functions, `visualizeRRT` and `visualizePath`, which accepts the name of the thing to visualize as an input arg.
* The original example will call one of each
* This example will call `visualizeRRT` once, and `visualizePath` twice (for the raw and shortcutted paths)

sea-bass: Follow-up: Maybe we instead just want to modify the existing RRT examples and add a flag for whether we want shortcutting or not?

bindings/src/roboplan/viser_visualizer.py
sea-bass: oh. We may want to land this straight to Pinocchio then...

roboplan/include/roboplan/core/path_utils.hpp
```diff
@@ -28,4 +28,39 @@ std::vector<Eigen::Matrix4d> computeFramePath(const Scene& scene, const Eigen::V
...
                             const Eigen::VectorXd& q_end, const double max_step_size);
 
+/// @brief Attempts to shortcut the path with random sampling and checking connections.
+/// @details This implementation is based on section 3.5.3 of:
+/// https://motion.cs.illinois.edu/RoboticSystems/MotionPlanningHigherDimensions.html
+/// @param scene The scene for checking connectability between joint poses.
+/// @param path The JointPath to try to shorten.
+/// @param max_step_size Maximum step size to use in collision checking.
+/// @param max_iters Maximum number of iterations of random sampling (default 100).
+/// @param seed Seed for the random generator, if 0 will not be applied (default 0).
```
sea-bass: I think 0 is a valid RNG seed? You may need to either use signed int and use `-1` as a sentinel, or go the `std::optional` route.

bindings/examples/example_path_shortcutting.py
eholum: Done and done.

I'm assuming it's preferable just to have all of the configurable parameters directly in code rather than adding arguments or something to these example scripts to do more code consolidation?

sea-bass: Either/or. Viser uses this other package named `tyro` which makes command line args pretty nice, if you wanna check it out. The viser examples are a good reference.

eholum: Oh that is need. Do you want to tackle that here? I would probably merge the ik and rrt examples together and make everything an argument, including the robot models. Thoughts?

sea-bass: That sounds great -- up to you if you wanna split it off or club it all together.

.docker/ros/Dockerfile
```diff
@@ -8,7 +8,7 @@ SHELL ["/bin/bash", "-c"]
 # Install core dependencies.
 RUN apt-get update -y \
     && apt install -y libgmock-dev python3-pip
-RUN pip3 install nanobind scikit-build-core tyro
+RUN pip3 install typing_extensions==4.10.0 nanobind scikit-build-core tyro
```
sea-bass: Note that line 31 does installs there too. If there is at all a chance to consolidate...

eholum: Whoops, sorry. Yeah didn't see that down below but makes more sense to install this alongside viser rather than with the build deps...

eholum: I'm just here to burn CI minutes.

PR #34, opened 2025-07-23
Add Franka model and examples

### LINKED ISSUE #27: Add 7-DOF Franka Arm Model
The UR5 model has been fine and dandy, but I'd like a Franka for 2 reasons:

1. It's got 7DOF, so redundancy will be good to e.g. show off IK and such
2. Its gripper has 2 extra joints (and optionally we could make it use the mimic joint variety), and I'm interested in starting to see the effects of having joint groups with subsets of joints.

sea-bass:
image

image

Closes #27

### DIFF
--- a/bindings/examples/example_rrt_franka.py
+++ b/bindings/examples/example_rrt_franka.py
@@ -12,6 +12,7 @@ from roboplan import (
 )
 from roboplan.viser_visualizer import ViserVisualizer, visualizePath
 
+
 if __name__ == "__main__":
 
     roboplan_examples_dir = Path(get_package_share_dir())
--- a/bindings/examples/example_rrt_ur.py
+++ b/bindings/examples/example_rrt_ur.py
@@ -12,68 +12,7 @@ from roboplan import (
     RRTOptions,
     RRT,
 )
-from roboplan.viser_visualizer import ViserVisualizer
-
-
-def visualizePath(
-    viz: ViserVisualizer,
-    scene: Scene,
-    rrt: RRT,
-    path: JointPath,
-    frame_name: str,
-    max_step_size: float,
-) -> None:
-    """
-    Helper function to visualize the RRT path.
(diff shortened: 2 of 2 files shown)


roboplan_example_models/models/franka_robot_model/README.md
```diff
@@ -0,0 +1 @@
+These files were generated from the [`franka_description`](https://github.com/frankarobotics/franka_description) repository.
```
JafarAbdi, line 1: What do you think about adding https://github.com/robot-descriptions/robot_descriptions.py as dev dependency and use the models from that repo? One thing I did in ramp is https://github.com/JafarAbdi/ramp/blob/e7dda50f7bcb278ff65e8075f3c8a9a2b9bd73d6/src/ramp/pinocchio_utils.py#L193-L204, not saying we should do it here as well :D:D. Maybe something to consider in the future to keep cloning this repo fast

sea-bass: I considered something like this too!

Will it work on the C++ side?

JafarAbdi: I saw this but never used it https://github.com/mayataka/robot_descriptions.cpp

sea-bass: Interesting! I'll make a ticket to explore it. I would say we need to figure out whether it's worth the extra code generation and anyways cloning the meshes for tests/examples vs. just having a handful of robot models in the repo.

I imagine this will become more useful once our number of robot models increases.

bindings/examples/example_ik_ur.py
eholum, line 1: Since you added it to the visualizer do you want to remove the `visualizePath` impl from here?

sea-bass: Good catch! Removed it from the `example_rrt_ur.py` example.

PR #42, opened 2025-08-03
TOPP-RA path parameterization

### LINKED ISSUE #21: Incorporate an off-the-shelf time-optimal trajectory optimization algorithm
The best candidate is probably going to be TOPP-RA: https://github.com/hungpham2511/toppra. It doesn't have binaries, so this will probably start encroaching on submodule territory?

Ruckig is kinda "meh" for longer paths, and more importantly requires a proprietary version for multi-waypoint timing that isn't throttled by a cloud API.

Other less nice options would be to implement our own simple polynomial / piecewise-polynomial timing solutions that have zero (or approximated) velocities at the waypoints, without requiring external packages.

Of course, if someone is comfortable doing their own better implementation that would be cool too!

sea-bass:
This PR adds a wrapper around [TOPP-RA](https://github.com/hungpham2511/toppra), using a fork with some slight modifications to make it build with ROS 2, of course.

As this approach also requires acceleration limits, it is time to add extra stuff beyond URDF/SRDF (to be addressed in a follow on that solves #43)

Closes #21 

https://github.com/user-attachments/assets/777222a4-e45b-432d-8d15-19a36de5c354

### DIFF
--- a/.docker/ros/Dockerfile
+++ b/.docker/ros/Dockerfile
@@ -6,7 +6,8 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
 SHELL ["/bin/bash", "-c"]
 
 # Install core dependencies.
-RUN apt-get update -y \
+RUN apt update -y \
+    && apt upgrade -y \
     && apt install -y libgmock-dev python3-pip
 RUN pip3 install nanobind scikit-build-core
 
@@ -17,7 +18,6 @@ COPY . src/roboplan
 
 # Install dependencies and build RoboPlan.
 RUN source /opt/ros/${ROS_DISTRO}/setup.bash \
-    && apt-get update -y \
     && rosdep install --from-paths src --ignore-src --rosdistro ${ROS_DISTRO} -y \
     && colcon build --cmake-args -DBUILD_TESTING=ON
 
--- a/.docker/ubuntu/Dockerfile
+++ b/.docker/ubuntu/Dockerfile
@@ -7,7 +7,8 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
 SHELL ["/bin/bash", "-c"]
 
 # Install core dependencies.
-RUN apt-get update -y \
+RUN apt update -y \
+    && apt upgrade -y \
     && apt install -y \
         cmake libeigen3-dev libgmock-dev liburdfdom-dev python3-pip
 RUN pip3 install nanobind pytest scikit-build-core
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,8 @@
 .git/
 .github
 .gitignore
+.pixi/
+
 **/build/
 **/install/
 **/Dockerfile
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -106,7 +106,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-build-${{ github.sha }}
             ${{ runner.os }}-build
-      - uses: prefix-dev/setup-pixi@v0.8.3
+      - uses: prefix-dev/setup-pixi@v0.8.14
         with:
           cache: true
           cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}


bindings/examples/example_rrt.py
```diff
@@ -60,9 +62,6 @@ def main(
     viz = ViserVisualizer(model, collision_model, visual_model)
     viz.initViewer(open=True, loadModel=True, host=host, port=port)
 
-    # Optionally include path shortening
-    include_shortcutting = True
```
sea-bass, line 64: woopsie

roboplan_toppra/src/toppra.cpp
```diff
@@ -0,0 +1,145 @@
...
+      throw std::runtime_error("Continuous joints not yet supported by TOPP-RA.");
+    default:  // Prismatic, revolute, or continuous, which are single-DOF.
+      if (joint_info.limits.max_velocity.size() == 0) {
+        throw std::runtime_error("Velocity limit must be defined for joint '" + joint_name + "'.");
+      }
+      // TODO: Re-enable before this PR lands
+      // if (joint_info.limits.max_acceleration.size() == 0) {
+      //   throw std::runtime_error("Acceleration limit must be defined for joint '" + joint_name +
+      //                            "'.");
+      // }
```
sea-bass, line 42: might do this in a follow on TBH to address #43

bindings/examples/example_rrt.py
```diff
@@ -2,11 +2,13 @@
 import time
 import tyro
 
+import matplotlib.pyplot as plt
```
eholum, line 5: Add this as a dep in pixi? And possibly also install it in docker images?

bindings/examples/example_rrt.py
eholum: It's more exciting this way

external/toppra
eholum, line 1: Does this actually need to be a submodule if it's installable from pypi?

roboplan_toppra/src/toppra.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  // Extract joint velocity + acceleration limits from scene and create constraints.
+  // TODO: Extract only for specified joint group.
+  const auto num_joints = scene->getModel().nq;
+  toppra::Vector vel_lower_limit = Eigen::VectorXd::Zero(num_joints);
+  toppra::Vector vel_upper_limit = Eigen::VectorXd::Zero(num_joints);
+  toppra::Vector acc_lower_limit = Eigen::VectorXd::Zero(num_joints);
+  toppra::Vector acc_upper_limit = Eigen::VectorXd::Zero(num_joints);
+
+  const auto joint_names = scene->getJointNames();
```
eholum: Maybe worth pulling this out of the scene at request time rather than construction time? I think that's what I would intuitively expect.

roboplan_toppra/src/toppra.cpp
```diff
@@ -0,0 +1,145 @@
...
+  const auto num_pts = path.positions.size();
+  if (num_pts < 2) {
+    return tl::make_unexpected("Path must have at least 2 points.");
+  }
+  if (dt <= 0.0) {
+    return tl::make_unexpected("dt must be strictly positive.");
+  }
+
+  // Create initial cubic spline with path and random times.
+  // TODO: Support nonzero velocity at end points?
```
eholum: Will this just work out of the box if/when joint paths have velocities?

sea-bass: not unless you change those boundary conditions... so no

external/toppra
sea-bass: it unfortunately does, because the pypi version does not even use the C++ library.

If you look at the `toppra` repo, there is a pure Python implementation, and separately a C++ with bindings... at least that's how I understood it?

EDIT: Hmm... maybe I'm wrong. Seems like `toppra.cpp` is specifically a module within this package that does include the c++ bindings.

roboplan_toppra/src/toppra.cpp
sea-bass: Why? I was thinking the opposite since this constructor just grabs all the (fixed) joint limits and packages them into a vector as a one-time operation.

Were you thinking variable joint limits?

external/toppra
eholum: Oh yep. Seems like it... Funny.

sea-bass: Well, the thing is toppra began as a pure Python package and then the C++ implementation was added later. So this is likely why it works this way.

roboplan_toppra/src/toppra.cpp
eholum: > Were you thinking variable joint limits?

Yeah just thinking it would be nice to change those at request time. Alternatively could make the scene's "speed limits" the default and sometime down the road add an optional parameter to scale or set them?

sea-bass: Scaling should be easy, especially since it's all an Eigen vector.

Another argument is that if you change the scene's joints or limits, you can just construct a new instance of this object?

eholum: Yeah I'm all for scaling.

sea-bass: I can add that to this PR.

external/toppra
sea-bass: Confirmed that when you install the `toppra` package from `pip`, it has nothing in the cpp module. Sad.

roboplan_toppra/src/toppra.cpp
```diff
@@ -0,0 +1,144 @@
...
+      }
+      // TODO: Re-enable before this PR lands
+      // if (joint_info.limits.max_acceleration.size() == 0) {
+      //   throw std::runtime_error("Acceleration limit must be defined for joint '" + joint_name +
+      //                            "'.");
+      // }
+      const auto& max_vel = joint_info.limits.max_velocity[0];
+      vel_lower_limits_(idx) = -max_vel;
+      vel_upper_limits_(idx) = max_vel;
+      const auto& max_acc = 1.0;  // joint_info.limits.max_acceleration[0];
```
eholum, line 46: Looks like parsing these from yaml is arriving imminently?

sea-bass: Yea it's ready (minus pixi shenanigans) in https://github.com/open-planning/roboplan/pull/45.

Tagged @JafarAbdi as my 1 allowed reviewer to see if he knew what was happening

PR #44, opened 2025-08-07
Ensure to run apt upgrade in Dockerfiles

sea-bass:
Per https://github.com/stack-of-tasks/pinocchio/issues/2742, I finally found the silliness that was causing Rolling CI to fail... sigh.

Also, copying the Docker context took forever because we're trying to copy in pixi virtualenvs on the sly ;)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -100,7 +100,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-build-${{ github.sha }}
             ${{ runner.os }}-build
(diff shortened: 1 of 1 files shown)


.docker/ros/Dockerfile
```diff
@@ -6,7 +6,8 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
 SHELL ["/bin/bash", "-c"]
 
 # Install core dependencies.
-RUN apt-get update -y \
+RUN apt update -y \
+    && apt upgrade -y \
```
eholum, line 10: What could possibly go wrong?

PR #45, opened 2025-08-09
Specify acceleration and jerk limits through YAML config file

### LINKED ISSUE #43: Provide mechanism to specify acceleration and jerk limits
Critically, URDF/SRDF do not have any place to specify joint limits beyond velocity.

One option could be a YAML file with extra limits, but any other ideas?

sea-bass:
Closes #43

### DIFF
--- a/roboplan/CMakeLists.txt
+++ b/roboplan/CMakeLists.txt
@@ -31,10 +31,11 @@ target_link_libraries(roboplan
     Eigen3::Eigen
     pinocchio::pinocchio
 )
-target_link_libraries(roboplan
-    PRIVATE
-    yaml-cpp::yaml-cpp
-)
+if (yaml-cpp_VERSION VERSION_LESS "0.8")  # for Ubuntu 22.04
+    target_link_libraries(roboplan PRIVATE yaml-cpp)
+else()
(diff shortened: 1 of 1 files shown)


pixi.toml
```diff
@@ -190,6 +190,7 @@ libstdcxx   = ">=15.1.0,<16"
 llvm-openmp = ">=19.1.7,<20"
 mold        = ">=2.40.2,<3"
 sccache     = ">=0.10.0,<0.11"
+yaml-cpp    = ">=0.8.0,<0.9"  # TODO: Why this no work
```
sea-bass: @JafarAbdi do you know why this is still causing a linking failure where `mold` is not able to find `yaml-cpp`?

sea-bass: Update: seems I could resolve it by adding the appropriate `PRIVATE` / `PUBLIC` keywords in `target_link_libraries`. Why? I have no idea.

JafarAbdi: Oops, I didn't see the other tag as well, sorry. I'm looking into it now

JafarAbdi: Looks like you fixed it :saluting_face:

PR #46, opened 2025-08-09
Fix path shortcutting logic

sea-bass:
I realized that path shortcutting never returned a path with 2 points; it always capped out at 3, where the 1st and 2nd points were almost right next to each other.

Debugging, we kept exceeding maximum iterations at 3 points. The shortcutting was doing its best to exactly sample lengths 0 and 1, which would have worked in the limit of infinite iterations.

Turns out we needed a special case for exactly 3 points, in which you just try to remove the middle point and return the path no matter the outcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Still debating whether this should be monorepo or multi-repo...
 - `roboplan` : The core C++ library.
 - `roboplan_simple_ik` : A simple inverse kinematics (IK) solver.
 - `roboplan_rrt` : A Rapidly-exploring Random Tree (RRT) based motion planner.
+- `roboplan_toppra` : A wrapper around the TOPP-RA algorithm for trajectory timing.
 - `roboplan_example_models` : Contains robot models used for testing and examples.
 - `roboplan_examples` : Basic examples with real robot models.
 


roboplan/src/core/path_utils.cpp
```diff
@@ -60,8 +60,14 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
...
-    // The path is at maximum shortcutted-ness
     if (path_configs.size() < 3) {
+      // The path is at maximum shortcutted-ness
+      return shortened_path;
+    } else if (path_configs.size() == 3) {
+      // If the path has exactly 3 points, exclusively try to bypass the middle one
+      if (!hasCollisionsAlongPath(scene, path_configs[0], path_configs[2], max_step_size)) {
+        path_configs.erase(path_configs.begin() + 1);
+      }
       return shortened_path;
```
eholum: Same question as the other PR but I am questioning myself about it... Those cases where,

`A -------> B -------> C`

where `A` and `B` are not connectable but there's some intermediate `B'` that shortens this, to

`A -> B' -> C`

As written I _think_ this could find that intermediate point but maybe the likelihood of just constantly hitting the case on line 92 makes it not even worth it.

sea-bass: Yep, I think you're right! Fixed.

sea-bass: I also just added a boolean so that you're not constantly checking the same 3-point path repeatedly. The new special case only triggers once per unique 3-point path (if I did it right...)

PR #49, opened 2025-08-12
Fix IK example

sea-bass:
CI is busted right now, so please ignore that, but also please test!

This PR makes some fixes to the IK example, namely:
1. Fixes syntax to also include the new YAML config file for the model
2. Ensures IK always solves from the previous joint configuration, not always the same starting configuration (also renamed `start_pose` to `q_start` because that was misleading)
3. Returns in the IK solver if the velocity ends up becoming `NaN` (this is what crashed the example when I was showing you!)

### DIFF
--- a/bindings/examples/example_ik.py
+++ b/bindings/examples/example_ik.py
@@ -89,9 +89,7 @@ def main(
 
     @reset_button.on_click
     def reset_position(_):
-        print(f"Start positions: {start.positions}")
         fk_tform = scene.forwardKinematics(start.positions, goal.tip_frame)
-        print(f"FK:\n{fk_tform}")
         controls.position = fk_tform[:3, 3]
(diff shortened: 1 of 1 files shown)


bindings/examples/example_ik.py
```diff
@@ -89,7 +89,9 @@ def solve_ik(_):
 
     @reset_button.on_click
     def reset_position(_):
+        print(f"Start positions: {start.positions}")
```
sea-bass:
```suggestion
```

bindings/examples/example_ik.py
```diff
@@ -89,7 +89,9 @@ def solve_ik(_):
...
         fk_tform = scene.forwardKinematics(start.positions, goal.tip_frame)
+        print(f"FK:\n{fk_tform}")
```
sea-bass:
```suggestion
```

Issue #41, opened 2025-08-03
Implement notion of joint groups


sea-bass:
The key issue here is that if you have e.g. a 16 DOF robot with 2 arms and grippers, you may only want to plan for a single arm and no gripper.

Doesn't necessarily have to be a formal joint group concept, but at least a way to tell planners, IK solvers, etc. to only deal with specific indices.

However, SRDF already contains this concept so that tends to help.

PR #50, opened 2025-08-15
Add an example model with two fr3s

eholum:
Will be useful for https://github.com/open-planning/roboplan/issues/41, if we want it.

I added a scene constructor to be able to load urdfs from xml strings rather in addition to from files. As noted in the discussions and comments... this has some issues with nanobind.

TODO: Planning groups:

https://github.com/user-attachments/assets/d464563e-c260-4dad-bff3-b7db39763d21

### DIFF
--- a/bindings/examples/example_ik.py
+++ b/bindings/examples/example_ik.py
@@ -47,7 +47,7 @@ def main(
     urdf_xml = xacro.process_file(model_data.urdf_path).toxml()
     srdf_xml = xacro.process_file(model_data.srdf_path).toxml()
 
-    # TODO: How can we identify the right constructor? Or are argument name differences good enough?
+    # Specify argument names to distinguish overloaded Scene constructors from python.
     scene = Scene(
         "test_scene",
         urdf=urdf_xml,
@@ -79,7 +79,7 @@ def main(
     start.positions = np.array(model_data.starting_joint_config)
     goal = CartesianConfiguration()
     goal.base_frame = model_data.base_link
-    goal.tip_frame = model_data.ee_name
+    goal.tip_frame = model_data.ee_names[0]
     solution = JointConfiguration()
 
     # Create an interactive marker.
--- a/bindings/examples/example_rrt.py
+++ b/bindings/examples/example_rrt.py
@@ -58,8 +58,8 @@ def main(
     urdf_xml = xacro.process_file(model_data.urdf_path).toxml()
     srdf_xml = xacro.process_file(model_data.srdf_path).toxml()
 
-    # TODO: How can we identify the right constructor? Or are argument name differences good enough?
-    scene = Scene.from_xml(
+    # Specify argument names to distinguish overloaded Scene constructors from python.
+    scene = Scene(
         "test_scene",
         urdf=urdf_xml,
         srdf=srdf_xml,
@@ -109,8 +109,8 @@ def main(
     # Visualize the tree and path
     print(path)
     viz.display(start.positions)
-    visualizePath(viz, scene, path, model_data.ee_name, 0.05)
-    visualizeTree(viz, scene, rrt, model_data.ee_name, 0.05)
+    visualizePath(viz, scene, path, model_data.ee_names, 0.05)
+    visualizeTree(viz, scene, rrt, model_data.ee_names, 0.05)
 
     if include_shortcutting:
         print("Shortcutted path:")
@@ -119,7 +119,7 @@ def main(
             viz,
             scene,
             shortcut_path,
-            model_data.ee_name,
+            model_data.ee_names,
             0.05,
             (0, 100, 0),
             "/rrt/shortcut_path",
--- a/bindings/examples/common.py
+++ b/bindings/examples/common.py
@@ -21,7 +21,7 @@ class RobotModelConfig:
     urdf_path: Path
     srdf_path: Path
     yaml_config_path: Path
-    ee_name: str
+    ee_names: List[str]
     base_link: str
     starting_joint_config: List[float]
 
@@ -34,7 +34,7 @@ MODELS = {
         urdf_path=ROBOPLAN_EXAMPLES_DIR / "ur_robot_model" / "ur5_gripper.urdf",
         srdf_path=ROBOPLAN_EXAMPLES_DIR / "ur_robot_model" / "ur5_gripper.srdf",
         yaml_config_path=ROBOPLAN_EXAMPLES_DIR / "ur_robot_model" / "ur5_config.yaml",
-        ee_name="tool0",
+        ee_names=["tool0"],
         base_link="base",
         starting_joint_config=[0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
     ),
@@ -44,7 +44,7 @@ MODELS = {
         yaml_config_path=ROBOPLAN_EXAMPLES_DIR
         / "franka_robot_model"
         / "fr3_config.yaml",
-        ee_name="fr3_hand",
+        ee_names=["fr3_hand"],
         base_link="fr3_link0",
         starting_joint_config=[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
     ),
@@ -54,7 +54,7 @@ MODELS = {
         yaml_config_path=ROBOPLAN_EXAMPLES_DIR
         / "franka_robot_model"
         / "dual_fr3_config.yaml",
-        ee_name="left_fr3_hand",  # Default for convenience, but either tip is valid
+        ee_names=["left_fr3_hand", "right_fr3_hand"],
         base_link="left_fr3_link0",
         starting_joint_config=[0.0] * 18,
     ),
--- a/bindings/examples/example_dual_arm.py
+++ b/bindings/examples/example_dual_arm.py
@@ -16,5 +16,8 @@ if __name__ == "__main__":
     urdf = xacro.process_file(urdf_path).toxml()
     srdf = xacro.process_file(srdf_path).toxml()
 
-    scene = roboplan.Scene.from_xml("dual_arm_scene", urdf, srdf, package_paths)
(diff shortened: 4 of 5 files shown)


roboplan_example_models/models/franka_robot_model/fr3.urdf
```diff
@@ -1,351 +1,9 @@
...
-    <origin rpy="0 0 3.141592653589793" xyz="0 0 0.0584"/>
-    <axis xyz="0 1 0"/>
-    <limit effort="100" lower="0.0" upper="0.04" velocity="0.2"/>
-    <mimic joint="fr3_finger_joint1"/>
-    <dynamics damping="0.3"/>
-  </joint>
+<?xml version='1.0' encoding='utf-8'?>
+<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="fr3">
+
+  <xacro:include filename="fr3.xacro"/>
```
eholum, line 4: Do we love or hate Xacro in open-planning?

sea-bass: We must embrace it, regardless of my opinions, lol...

bindings/src/roboplan_ext.cpp
```diff
@@ -132,6 +132,21 @@ NB_MODULE(roboplan, m) {
            "name"_a, "urdf_path"_a, "srdf_path"_a,
            "package_paths"_a = std::vector<std::filesystem::path>(),
            "yaml_config_path"_a = std::filesystem::path())
+      // There's an ambiguity issue due to file paths vs strings in python, so if you're using xml
+      // we use a factory method for XMLs. We could alternatively change the function signatures, or
+      // or else figure out some other way to distinguish these.
```
sea-bass: I think file paths become `pathlib.Path` so is the ambiguity just that Python doesn't have a concept of function overloads because of abominations such as `args` and `kwargs`?

bindings/examples/example_ik.py
```diff
@@ -39,10 +39,24 @@ def main(
         print(f"Invalid model requested: {model}")
         sys.exit(1)
 
-    urdf_path, srdf_path, ee_name, base_link, start_pose = MODELS[model]
+    urdf_path, srdf_path, yaml_config_path, ee_name, base_link, q_start, is_xacro = MODELS[model]
```
sea-bass: I was gonna comment before, but this return tuple is becoming unsightly... I think this probably could benefit from a dataclass or something.

As for the `is_xacro` or not, maybe just always default to have this example helper dump out the strings for URDF/SRDF?

roboplan/src/core/scene.cpp
```diff
@@ -45,6 +45,36 @@ Scene::Scene(const std::string& name, const std::filesystem::path& urdf_path,
...
+
+  pinocchio::urdf::buildGeom(model_, std::istringstream(urdf), pinocchio::COLLISION,
+                                    collision_model_, package_paths_str);
+  collision_model_.addAllCollisionPairs();
+  pinocchio::srdf::removeCollisionPairsFromXML(model_, collision_model_, srdf);
+
+  initialize(yaml_config_path);
+}
+
+void Scene::initialize(const std::filesystem::path& yaml_config_path) {
```
sea-bass: hah this is funny, almost exactly how we have it working in our work stack for model parsing.

However, you can reduce duplication further by also absorbing the model building into `initialize()` and simply having the original constructor read files into strings and call the other one.

bindings/examples/example_ik.py
eholum: Hard agree. I'll clean this up...

I need to look more at Pinocchio's constructors/parsers, too. The temporary file is... irritating, so it would be nice if we could do more for the viz model with raw xml everywhere.

bindings/src/roboplan_ext.cpp
eholum: Another issue is some kind of nightmare memory management issue with static factory methods like this... Not sure exactly where the problem is but if I construct a scene with the factory method then I get random segfaults down the line. I'm guessing python is doing some cleanup of things that it has ownership of that it shouldn't be. I vote no factory method.

In answer to your question though, seems like the problem is it is implicitly converting strings to file paths so you end up with this fun error:

```
  File "/home/parallels/ROS/seabass/roboplan/bindings/examples/example_rrt.py", line 151, in <module>
    tyro.cli(main)
  File "/home/parallels/ROS/seabass/roboplan/.pixi/envs/default/lib/python3.11/site-packages/tyro/_cli.py", line 230, in cli
    return run_with_args_from_cli()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/parallels/ROS/seabass/roboplan/bindings/examples/example_rrt.py", line 69, in main
    scene = Scene("test_scene", urdf_xml, srdf_xml, package_paths=package_paths, yaml_config_path=model_data.yaml_config_path)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: filesystem error: status: File name too long [<?xml version="1.0" ?>
<!-- =================================================================================== -->
<!-- |    This document was autogenerated by xacro from /home/florent/devel/isae/install/share/ur_description/urdf/ur5_joint_limited_robot.urdf.xacro | -->
```

eholum: The cool thing is that it grabs the first matching constructor, so 

```

      .def(nanobind::init<const std::string&, const std::string&,
                          const std::string&, const std::vector<std::filesystem::path>&,
                          const std::filesystem::path&>(),
           "name"_a, "urdf"_a, "srdf"_a,
           "package_paths"_a = std::vector<std::filesystem::path>(),
           "yaml_config_path"_a = std::filesystem::path())
      .def(nanobind::init<const std::string&, const std::filesystem::path&,
                          const std::filesystem::path&, const std::vector<std::filesystem::path>&,
                          const std::filesystem::path&>(),
           "name"_a, "urdf_path"_a, "srdf_path"_a,
           "package_paths"_a = std::vector<std::filesystem::path>(),
           "yaml_config_path"_a = std::filesystem::path())
```

will work for string xml, but not with pathlib.Paths, which seem to resolve to strings, too. 

I think it's like Apollo 13, it doesn't matter what the constructor was designed to do, it matters what it can do.

bindings/src/modules/core.cpp
```diff
@@ -87,6 +87,14 @@ void init_core_scene(nanobind::module_& m) {
            "name"_a, "urdf_path"_a, "srdf_path"_a,
            "package_paths"_a = std::vector<std::filesystem::path>(),
            "yaml_config_path"_a = std::filesystem::path())
+      // There's an ambiguity issue due to file paths vs strings in python. So to use this
```
eholum, line 90: Is this better or worse?

As noted in the other comment, I think avoiding static factory methods in nanobind is a good idea.

eholum: Ok actually went through and read the docs (ha). Nanobind has really great handling of objects,

https://nanobind.readthedocs.io/en/latest/ownership.html#unique-ptr

So just making sure that the returned object is correctly owned prevents data in the pinnochio model from being cleaned up when it shouldn't be.

So we have choices.

bindings/examples/example_rrt.py
```diff
@@ -50,15 +51,30 @@ def main(
...
-    urdf_path, srdf_path, yaml_config_path, ee_name, _, _ = MODELS[model]
+    model_data = MODELS[model]
     package_paths = [ROBOPLAN_EXAMPLES_DIR]
 
-    scene = Scene("test_scene", urdf_path, srdf_path, package_paths, yaml_config_path)
+    # Pre-process with xacro. This is not necessary for raw URDFs.
+    urdf_xml = xacro.process_file(model_data.urdf_path).toxml()
+    srdf_xml = xacro.process_file(model_data.srdf_path).toxml()
+
+    # TODO: How can we identify the right constructor? Or are argument name differences good enough?
```
sea-bass: Does this todo still hold?

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -18,6 +18,11 @@
 #include <roboplan/core/types.hpp>
 
 namespace roboplan {
+
+/// @brief Returns the contents of a file as a string.
+/// @param name path The path to the file.
+std::string readfile(const std::filesystem::path& path);
```
sea-bass: Call it `readFile` for style consistency?

roboplan_example_models/models/franka_robot_model/dual_fr3_config.yaml
```diff
@@ -0,0 +1,45 @@
+# RoboPlan configuration file for the Franka Research 3 (FR3) robot model.
```
sea-bass: Update comment to point to dual arms

bindings/examples/example_rrt.py
eholum: Yeah... this should probably still be a draft PR but I decided that was meaningless. In any case, I wanted your feedback on constructors for complex python objects.

I guess all that matters is do you want the constructor with args to differentiate between them, or the static factory method?

This:

```
    scene = Scene.from_xml(
        "test_scene",
        urdf_xml,
        srdf_xml,
        package_paths=package_paths,
        yaml_config_path=model_data.yaml_config_path,
    )
```

vs

```
    scene = Scene(
        "test_scene",
        urdf=urdf_xml,  # Can differentiate from the file path constructor with args
        srdf=srdf_xml,
        package_paths=package_paths,
        yaml_config_path=model_data.yaml_config_path,
    )
```

sea-bass: I would opt for whatever is most consistent between the C++ and Python APIs.

In other words, I'd do the latter if we're sticking with constructors, but the former if the C++ side also operates via static factory methods.

PR #51, opened 2025-08-16
Organize Python bindings

### LINKED ISSUE #13: Figure out how to better organize Python bindings by folder and module
Right now it's all flat. It shouldn't be as the library grows.

sea-bass:
trust me bro, it builds

Closes #13

### DIFF
--- a/bindings/include/modules/example_models.hpp
+++ b/bindings/include/modules/example_models.hpp
@@ -2,6 +2,10 @@
 
 #include <nanobind/nanobind.h>
 
+namespace roboplan {
+
 /// @brief Initializes Python bindings for example models module.
 /// @param m The nanobind core module.
 void init_example_models(nanobind::module_& m);
+
+}  // namespace roboplan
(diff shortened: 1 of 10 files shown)


bindings/include/modules/example_models.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <nanobind/nanobind.h>
+
+/// @brief Initializes Python bindings for example models module.
+/// @param m The nanobind core module.
+void init_example_models(nanobind::module_& m);
```
eholum, line 9: Is there any particular reason the models aren't namespaced? I'm not even sure how that plays with nanobind but it doesn't seem to make a difference.

sea-bass: The reason was that the source code used a different namespace of `roboplan_example_models` instead of `roboplan`. I can just fix that for consistency tbh

eholum: Meh. You'd think I would remember these things but here we are.

PR #58, opened 2025-08-20
Create map of frame names to IDs in Scene 

### LINKED ISSUE #26: Create map of frame names to IDs in Scene
I recently learned that Pinocchio's `getFrameId()` function, which gives you the index of a specific frame by name, actually is not implemented with a map; it's a linear search.

Since we rely a lot on this function to make the API use frame names instead of indices, we should internally store our own map by calling `getFrameId()` once on model load and stashing it. Then things like calling forward kinematics repeatedly won't keep doing needless linear-time searches.

Again, this isn't really needed in Pinocchio itself, because the strategy there would be:

1. Get the index using `getFrameId()`
2. In whatever tight loop you have, just use the resulting index directly

catarina-p:
Added two new methods to `Scene`:

-  `createFrameMap` – builds a map of frame names to their IDs.
- `getFrameMapId` – retrieves a frame ID from the map by inputting the frame name.

Replaced calls to `pinocchio::getFrameId` with `Scene::getFrameMapId` across the codebase (except in `test_constraints.cpp`).

 Added a new unit test in `test_scene.cpp` for these methods.

Verified the new methods by updating and running `example_ik.py`.

Closes #26

### DIFF
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -216,13 +216,7 @@ std::ostream& operator<<(std::ostream& os, const Scene& scene) {
 
 void Scene::createFrameMap(const pinocchio::Model& model) {
   frame_map_.clear();  // Clear existing map if needed
-
-  if (model.nframes <= 1) {
-    std::cout << "Warning: Model has no frames to process" << std::endl;
-    return;
-  }
-
-  for (size_t i = 1; i < model.frames.size(); ++i) {
+  for (size_t i = 1; i < model.nframes; ++i) {
     const auto& frame = model.frames[i];
     frame_map_[frame.name] = model.getFrameId(frame.name);
   }


roboplan/include/roboplan/core/scene.hpp
```diff
@@ -100,6 +100,10 @@ class Scene {
   /// @brief Prints basic information about the scene.
   friend std::ostream& operator<<(std::ostream& os, const Scene& scene);
 
+  /// @brief Helper funtion to get the pinocchio ID of a frame through its name
+  /// @return The pinocchio frame ID
```
sea-bass: You'll need to add a 

```
@param name The name of the frame to look up.
```

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -129,6 +133,14 @@ class Scene {
 
   /// @brief The current state of the model (used to fill in partial states).
   JointConfiguration cur_state_;
+
+  /// @brief Maps each frame name to each respective frame ID
+  std::unordered_map<std::string, pinocchio::FrameIndex> frame_map_;
+
+  /// @brief Helper funtion to create a map of the robot's frame ID's
```
sea-bass:
```suggestion
  /// @brief Helper function to create a map of the robot's frame IDs.

File: roboplan/src/core/scene.cpp [hunk 3]
```diff
@@ -208,4 +212,26 @@ std::ostream& operator<<(std::ostream& os, const Scene& scene) {
...
+  }
+
+  for (size_t i = 1; i < model.frames.size(); ++i) {
+    const auto& frame = model.frames[i];
+    frame_map_[frame.name] = model.getFrameId(frame.name);
+  }
+}
+
+pinocchio::FrameIndex Scene::getFrameMapId(const std::string &name) const {
+  auto it = frame_map_.find(name);
```
sea-bass: Since we're compiling with C++20 you can instead do

```cpp
if (frame_map_.contains(name))

File: roboplan/src/core/scene.cpp [hunk 4]
```diff
@@ -208,4 +212,26 @@ std::ostream& operator<<(std::ostream& os, const Scene& scene) {
...
+  if (it == frame_map_.end()) {
+      throw std::runtime_error("Frame name '" + name + "' not found in frame_map_");
```
sea-bass: Rather than throwing an error, maybe let's make this use `tl::expected`? Check out some of the other examples in the code base that use that!

roboplan/test/test_scene.cpp
```diff
@@ -100,4 +101,24 @@ TEST_F(RoboPlanSceneTest, CollisionCheckAlongPath) {
...
+  EXPECT_EQ(scene_->getFrameMapId("wrist_2_link"), model.getFrameId("wrist_2_link"));
+  EXPECT_EQ(scene_->getFrameMapId("wrist_3_link"), model.getFrameId("wrist_3_link"));
+  EXPECT_EQ(scene_->getFrameMapId("ee_link"), model.getFrameId("ee_link"));
+  EXPECT_EQ(scene_->getFrameMapId("base"), model.getFrameId("base"));
+  EXPECT_EQ(scene_->getFrameMapId("tool0"), model.getFrameId("tool0"));
+  EXPECT_EQ(scene_->getFrameMapId("world"), model.getFrameId("world"));
+}
+
+
+} // namespace roboplan
```
sea-bass: Linter will yell at you when CI runs, but you need to leave a newline at end of files

roboplan/src/core/scene.cpp
sea-bass: Also generally it's preferable to check error cases first. So like

If the frame name is NOT in the map, return an error immediately.

Otherwise, do the main thing.

That is

```cpp
if (!frame_map_.contains(name)) {
  return tl::make_unexpected("blah");
}

return frame_map_.at(name);
```

roboplan/include/roboplan/core/scene.hpp
catarina-p: Thanks for catching that, and the typos!

roboplan/src/core/scene.cpp
catarina-p: Thanks! I looked a bit into the `tl:expected` examples in the repo and what it's doing and have now replaced it in the method as well

bindings/src/roboplan_ext.cpp
```diff
@@ -144,6 +144,7 @@ NB_MODULE(roboplan, m) {
       .def("isValidPose", &Scene::isValidPose)
       .def("interpolate", &Scene::interpolate)
       .def("forwardKinematics", &Scene::forwardKinematics)
+      .def("getFrameMapId", &Scene::getFrameMapId)
```
sea-bass: sorry, I had missed this because my first review was on my phone... but can we just call this `getFrameId` and leave out the "map" part? It's shorter and users don't need to know it's a map.

roboplan/src/core/scene.cpp
sea-bass: Did you by any chance forget to push your changes, by the way? From the above comment it seems you've applied the changes, but they're not visible on here.

bindings/src/roboplan_ext.cpp
catarina-p: Yeah, sure I can change this

roboplan/src/core/scene.cpp
catarina-p: Yes, I hadn't pushed since I still had to tie up some loose ends, but now there is a new push

bindings/test/test_scene.py
```diff
@@ -78,3 +79,23 @@ def test_collision_check_along_path(test_scene: Scene) -> None:
...
+
+
+def test_create_frame_map(test_scene: Scene) -> None:
+    roboplan_examples_dir = Path(get_package_share_dir())
+    urdf_path = roboplan_examples_dir / "ur_robot_model" / "ur5_gripper.urdf"
+    package_paths = [roboplan_examples_dir]
+    model, _, _ = pin.buildModelsFromUrdf(
+        urdf_path, package_dirs=package_paths
+    )
+    assert test_scene.getFrameId("base_link") == model.getFrameId("base_link")
```
sea-bass: Lot of copy pasting here! Would highly recommend doing this all in a for loop using the list of frame names

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -100,6 +102,11 @@ class Scene {
   /// @brief Prints basic information about the scene.
   friend std::ostream& operator<<(std::ostream& os, const Scene& scene);
 
+  /// @brief Helper function to get the pinocchio ID of a frame through its name.
+  /// @param name The name of the frame to look up.
+  /// @return The pinocchio frame ID.
```
sea-bass: This if successful, else a string describing the error.

roboplan/test/test_scene.cpp
```diff
@@ -100,4 +101,24 @@ TEST_F(RoboPlanSceneTest, CollisionCheckAlongPath) {
...
 }
 
-}  // namespace roboplan
+// Testing the function for issue #26
+TEST_F(RoboPlanSceneTest, GetFrameMapReturnsCorrectMapping) {
+  // Act
+  pinocchio::Model model = scene_->getModel();
+  
+  // Verify the frame IDs are correct
+  EXPECT_EQ(scene_->getFrameId("base_link"), model.getFrameId("base_link"));
```
sea-bass: Same here re: for loops

bindings/include/utils/expected.hpp
```diff
@@ -38,3 +38,12 @@ template <typename Ret, typename Err, typename... Args>
 auto unwrap_expected(tl::expected<Ret, Err> (*method)(Args...)) {
   return [method](Args... args) -> Ret { return handle_expected((*method)(args...)); };
 }
+
+/// @brief Wrapper function binding tl::expected return types in const fucntions with nanobind.
```
sea-bass: I think this is a rebase conflict, as it seems like a duplicate of the above

bindings/examples/example_rrt.py
```diff
@@ -131,4 +131,4 @@ def main(
 
 
 if __name__ == "__main__":
-    tyro.cli(main)
+    tyro.cli(main)
```
sea-bass: Seems something broke formatting here

bindings/src/roboplan/viser_visualizer-old.py
```diff
@@ -0,0 +1,342 @@
+"""
```
sea-bass: This file shouldn't be here?

roboplan/test/test_path_utils.cpp
```diff
@@ -101,4 +101,4 @@ TEST_F(RoboPlanPathUtilsTest, testShortcutPath) {
   ASSERT_EQ(shortcut_path.positions[1], test_path.positions[3]);
 }
 
-}  // namespace roboplan
+}  // namespace roboplan
```
sea-bass: Again, ensure you don't break the code formatting

roboplan_examples/examples/example_scene.cpp
```diff
@@ -27,4 +27,4 @@ int main(int /*argc*/, char* /*argv*/[]) {
   std::cout << "Random positions (seeded):\n" << random_positions.transpose() << "\n";
 
   return 0;
-}
+}
```
sea-bass: And here

bindings/include/utils/expected.hpp
```diff
@@ -32,6 +32,16 @@ auto unwrap_expected(tl::expected<Ret, Err> (Class::*method)(Args...)) {
...
 
+/// @brief Wrapper function binding tl::expected return types in const class functions with
+/// nanobind.
+/// @return The unwrapped value, or throw a runtime_error.
+template <typename Class, typename Ret, typename Err, typename... Args>
+auto unwrap_expected(tl::expected<Ret, Err> (Class::*method)(Args...) const) {
+  return [method](const Class& self, Args... args) -> Ret {
+    return handle_expected((self.*method)(args...));
+  };
+}
```
sea-bass, line 43: ok I guess this was actually needed for `const` class methods!

roboplan/src/core/scene.cpp
```diff
@@ -51,6 +53,8 @@ Scene::Scene(const std::string& name, const std::filesystem::path& urdf_path,
   YAML::Node yaml_config;
   if (!yaml_config_path.empty() && !std::filesystem::is_directory(yaml_config_path)) {
     yaml_config = YAML::LoadFile(yaml_config_path);
+
+    createFrameMap(model_);
```
sea-bass: oh, this should definitely not be called inside of this block! It means if you don't specify a YAML config, this frame map won't get filled. Will fix this.

roboplan/src/core/scene.cpp
```diff
@@ -181,7 +185,11 @@ Eigen::Matrix4d Scene::forwardKinematics(const Eigen::VectorXd& q,
   // TODO: I recently learned recently that Pinocchio's getFrameId() actually does a linear-time
   // search! So we should put together a map.
```
sea-bass: this TODO can now go away

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -129,6 +136,13 @@ class Scene {
 
   /// @brief The current state of the model (used to fill in partial states).
   JointConfiguration cur_state_;
+
+  /// @brief Maps each frame name to each respective frame ID
+  std::unordered_map<std::string, pinocchio::FrameIndex> frame_map_;
+
+  /// @brief Helper function to create a map of the robot's frame IDs.
+  /// @param model The Pinocchio model.
+  void createFrameMap(pinocchio::Model model);
```
sea-bass: this needs to be const ref!

roboplan/test/test_scene.cpp
```diff
@@ -100,4 +100,16 @@ TEST_F(RoboPlanSceneTest, CollisionCheckAlongPath) {
   EXPECT_TRUE(hasCollisionsAlongPath(*scene_, q_start_free, q_end_coll, 0.05));
 }
 
+TEST_F(RoboPlanSceneTest, GetFrameMapReturnsCorrectMapping) {
+  // Act
```
sea-bass: ?

PR #59, opened 2025-08-24
Add Kinova + Robotiq model, initial limited support for continuous and mimic joints

### LINKED ISSUE #53: Add Kinova Gen3 + Robotiq 2F-85 model
This will also require supporting continuous and mimic joints which will be fun.

sea-bass:
This PR
* Add a Kinova Gen3 + Robotiq 2F-85 model, including continuous and mimic joints
* Adds basic (and sometimes incomplete/incorrect) support for continuous and mimic joints in scenes, IK, and RRT.

Closes #53

### DIFF
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -11,6 +11,7 @@
 
 namespace {
 
+/// @brief Map from Pinocchio joint model short names to RoboPlan joint type enums.
 const std::map<std::string, roboplan::JointType> kPinocchioJointTypeMap = {
     {"JointModelPX", roboplan::JointType::PRISMATIC},
     {"JointModelPY", roboplan::JointType::PRISMATIC},
@@ -27,6 +28,10 @@ const std::map<std::string, roboplan::JointType> kPinocchioJointTypeMap = {
     {"JointModelFreeFlyer", roboplan::JointType::FLOATING},
     {"JointModelMimic", roboplan::JointType::UNKNOWN},
 };
+
+/// @brief Tolerance for the norm of continuous joint values (in the form cos(theta), sin(theta))
+/// to be on the unit circle.
+constexpr double kUnitCircleTol = 1.0e-6;
 
 }  // namespace
 
@@ -240,7 +245,11 @@ bool Scene::isValidPose(const Eigen::VectorXd& q) const {
     switch (info.type) {
     // TODO: Validate multi-DOF joints.
     case JointType::CONTINUOUS:
-      q_idx += 2;  // Unbounded so always valid.
+      // Unbounded so always valid, but check whether the representation is on the unit circle.
+      if (std::abs(std::pow(q(q_idx), 2) + std::pow(q(q_idx + 1), 2) - 1.0) > kUnitCircleTol) {
+        return false;
+      }
+      q_idx += 2;
       break;
     default:
       for (size_t idx = 0; idx < info.num_position_dofs; ++idx) {
--- a/roboplan_toppra/src/toppra.cpp
+++ b/roboplan_toppra/src/toppra.cpp
@@ -30,21 +30,8 @@ PathParameterizerTOPPRA::PathParameterizerTOPPRA(const std::shared_ptr<Scene> sc
     case JointType::PLANAR:
       throw std::runtime_error("Multi-DOF joints not yet supported by TOPP-RA.");
     case JointType::CONTINUOUS:
-      // TODO: Should convert to single-DOF limit.
-      // Basically, right now continuous joint velocity limits are wrong.
-      // The solution should be to squash the position vectors so continuous
-      // joints are also represented as single-DOF.
-      vel_lower_limits_(q_idx) = -joint_info.limits.max_velocity[0];
-      vel_lower_limits_(q_idx + 1) = -joint_info.limits.max_velocity[0];
-      vel_upper_limits_(q_idx) = joint_info.limits.max_velocity[0];
-      vel_upper_limits_(q_idx + 1) = joint_info.limits.max_velocity[0];
-      acc_lower_limits_(q_idx) = -joint_info.limits.max_acceleration[0];
-      acc_lower_limits_(q_idx + 1) = -joint_info.limits.max_acceleration[0];
-      acc_upper_limits_(q_idx) = joint_info.limits.max_acceleration[0];
-      acc_upper_limits_(q_idx + 1) = joint_info.limits.max_acceleration[0];
-      q_idx += 2;
-      break;
-    default:  // Prismatic, revolute, or continuous, which are single-DOF.
+      throw std::runtime_error("Continuous joints not yet supported by TOPP-RA.");
+    default:  // Prismatic or revolute, which are single-DOF.
       if (joint_info.limits.max_velocity.size() == 0) {
         throw std::runtime_error("Velocity limit must be defined for joint '" + joint_name + "'.");
       }


roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -3,15 +3,16 @@
 namespace roboplan {
 
 SimpleIk::SimpleIk(const std::shared_ptr<Scene> scene, const SimpleIkOptions& options)
-    : scene_{scene}, options_{options} {};
+    : scene_{scene}, options_{options} {
+  data_ = pinocchio::Data(scene_->getModel());
```
sea-bass, line 7: incidental fix, but creating the data as a member variable will make the IK way less slow lol, no need to allocate a new one every time

roboplan/src/core/scene.cpp
```diff
@@ -148,13 +184,32 @@ Eigen::VectorXd Scene::randomPositions() {
...
+      for (size_t idx = 0; idx < info.num_position_dofs; ++idx) {
+        const auto& lo = info.limits.min_position[idx];
+        const auto& hi = info.limits.max_position[idx];
+        positions(q_idx) = std::uniform_real_distribution<double>(lo, hi)(rng_gen_);
+        ++q_idx;
+      }
     }
   }
+
+  // TODO: Set mimic relationships.
```
sea-bass: oh yeah, I need to do this...

bindings/examples/common.py
```diff
@@ -58,4 +58,36 @@ class RobotModelConfig:
...
+    "kinova": RobotModelConfig(
+        urdf_path=ROBOPLAN_EXAMPLES_DIR / "kinova_robot_model" / "kinova_robotiq.urdf",
+        srdf_path=ROBOPLAN_EXAMPLES_DIR / "kinova_robot_model" / "kinova_robotiq.srdf",
+        yaml_config_path=ROBOPLAN_EXAMPLES_DIR
+        / "kinova_robot_model"
+        / "kinova_robotiq_config.yaml",
+        ee_names=["robotiq_85_base_link"],
+        base_link="base_link",
+        starting_joint_config=[
+            1.0,
```
eholum, line 70: Joints on joints on joints on joints.

roboplan_example_models/models/kinova_robot_model/kinova_robotiq.urdf
```diff
@@ -0,0 +1,497 @@
...
+  <joint name="end_effector" type="fixed">
+    <origin rpy="3.14159265358979 1.09937075168372E-32 0" xyz="0 0 -0.0615250000000001"/>
+    <parent link="bracelet_link"/>
+    <child link="end_effector_link"/>
+    <axis xyz="0 0 0"/>
+  </joint>
+  <link name="robotiq_85_base_link">
+    <visual>
+      <geometry>
+        <mesh filename="package://robotiq_description/meshes/visual/robotiq_base.dae"/>
```
eholum: Do you possibly have the robotiq description installed on your machine? I think these should be

```
filename="package://kinova_robot_model/meshes/visual/robotiq_base.dae"/>
```

Otherwise it can't find the mesh.

roboplan/src/core/scene.cpp
```diff
@@ -148,13 +184,31 @@ Eigen::VectorXd Scene::randomPositions() {
...
   for (const auto& joint_name : joint_names_) {
     const auto& info = joint_info_.at(joint_name);
-    for (size_t idx = 0; idx < info.num_position_dofs; ++idx) {
-      const auto& lo = info.limits.min_position[idx];
-      const auto& hi = info.limits.max_position[idx];
-      positions(q_idx) = std::uniform_real_distribution<double>(lo, hi)(rng_gen_);
-      ++q_idx;
+
+    // Mimic joints will get updated later.
+    if (info.mimic_info) {
```
eholum, line 196: Well, I think shenanigans might abound (more shenanigans?). Works great with the UR5, but the franka has a mimic joint and there are likely some indexing errors that are making things pretty unhappy.

```
  File "/home/parallels/ROS/seabass/roboplan/bindings/examples/example_rrt.py", line 158, in <module>
    tyro.cli(main)
  File "/home/parallels/ROS/seabass/roboplan/.pixi/envs/default/lib/python3.11/site-packages/tyro/_cli.py", line 230, in cli
    return run_with_args_from_cli()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/parallels/ROS/seabass/roboplan/bindings/examples/example_rrt.py", line 102, in main
    path = rrt.plan(start, goal)
           ^^^^^^^^^^^^^^^^^^^^^
IndexError: vector::_M_range_check: __n (which is 18446744073628644616) >= this->size() (which is 1)
(viser) Server stopped
```

The problem is that `model_.nq` for the franka is now `8` since we've lost the mimic joint. And that's all fine and dandy for pinnochio, but consumers of random joint poses are missing the extra joint. So we should probably have,

```
Eigen::VectorXd positions(model_.nq + model_.mimicked_joints.size());
```

But then of course we'd need to remove the mimicked indexes when we pass it to pinnichio for any kinematics or collision checking... otherwise we get,

```
  File "/home/parallels/ROS/seabass/roboplan/bindings/examples/example_rrt.py", line 158, in <module>
    tyro.cli(main)
  File "/home/parallels/ROS/seabass/roboplan/.pixi/envs/default/lib/python3.11/site-packages/tyro/_cli.py", line 230, in cli
    return run_with_args_from_cli()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/parallels/ROS/seabass/roboplan/bindings/examples/example_rrt.py", line 93, in main
    start.positions = scene.randomCollisionFreePositions()
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: wrong argument size: expected 8, got 9
hint: The configuration vector is not of right size
```

FWIW, I think think it's best if every _external_ consumer of `Scene` gets every joint position (including mimics), and we should just handle the nightmare that these are in here...

roboplan/src/core/scene.cpp
```diff
@@ -190,6 +244,29 @@ bool Scene::isValidPose(const Eigen::VectorXd& q) const {
...
+
+    const auto& joint_info = joint_info_.at(mimicking_joint_name);
+    const auto& mimic_info = joint_info.mimic_info.value();
+    if (joint_info.type == JointType::CONTINUOUS) {
+      const auto mimicked_angle = std::atan2(q(mimicked_idx_q + 1), q(mimicked_idx_q));
+      const auto mimicking_angle = mimicked_angle * mimic_info.scaling + mimic_info.offset;
+      q(mimicked_idx_q) = std::cos(mimicking_angle);
+      q(mimicked_idx_q + 1) = std::sin(mimicking_angle);
+    } else {  // Prismatic or revolute, which are single-DOF.
+      q(mimicked_idx_q) = q(mimicking_idx_q) * mimic_info.scaling + mimic_info.offset;
```
eholum: I'm not 100% sure the assignments here are correct... can you convince me that they're not backwards?

roboplan/src/core/scene.cpp
eholum: On that note we should probably just have some RRT integration test that just include all the freaking models...

eholum: Could we just keep a list of all the non mimic joints and index out the full joint state vectors when we pass things to Pinocchio? Just started playing around with it but didn't get far yet.

roboplan_example_models/models/kinova_robot_model/kinova_robotiq.urdf
sea-bass: woopsie, good catch

roboplan/src/core/scene.cpp
sea-bass: gah, thanks for looking into that and agreed re: tests.

Guess we about to blow up the PR size with things that we should have probably done beforehand lol

roboplan/src/core/scene.cpp
sea-bass: I cannot convince you of that as I seem to be a fool

sea-bass: also something that a test would have caught. However, I am gonna hold off on testing this until the joint groups thing because I have a plan.

roboplan/src/core/scene.cpp
sea-bass: Fun question: Because Franka uses xacro... how on earth do you parse that in pure C++ code?

Edit: it appears one cannot. So just need to drop the pre-parsed urdfs into the repo? Or only test in Python?

bindings/examples/common.py
sea-bass: would be 17 instead of 12 if we didn't have `mimic = True`!

roboplan/src/core/scene.cpp
eholum: Oh my christ yeah I'll fix that...

I think it's probably easiest to just make a system call wrapper that will process it for you... Maybe not in the Scene though. Possibly in the example models folder directly?

roboplan_rrt/src/rrt.cpp
```diff
@@ -11,25 +11,39 @@ RRT::RRT(const std::shared_ptr<Scene> scene, const RRTOptions& options)
...
+  size_t q_idx = 0;
+  for (const auto& joint_name : joint_names) {
     const auto& joint_info = scene_->getJointInfo(joint_name);
     switch (joint_info.type) {
     case JointType::FLOATING:
     case JointType::PLANAR:
       throw std::runtime_error("Multi-DOF joints not yet supported by RRT.");
     case JointType::CONTINUOUS:
-      throw std::runtime_error("Continuous joints not yet supported by RRT.");
+      // TODO: Use "SO2" state space instead.
```
eholum, line 29: Would these paths actually be valid before we handle this `TODO`? Assuming Pinocchio handles distance/interpolation correctly it _should_ work, right? I'm breaking my brain thinking about what's actually happening that makes this almost work. Or possibly an issue in the viz itself?

https://github.com/user-attachments/assets/76ee9e12-2a2e-4406-b274-b7f9bf9fb56f

roboplan/src/core/scene.cpp
```diff
@@ -178,18 +232,76 @@ bool Scene::isValidPose(const Eigen::VectorXd& q) const {
...
+    if (info.mimic_info) {
       ++q_idx;
+      continue;  // Skip over mimic joints since we validate their parent.
+    }
+
+    switch (info.type) {
+    // TODO: Validate multi-DOF joints.
+    case JointType::CONTINUOUS:
+      q_idx += 2;  // Unbounded so always valid.
+      break;
```
eholum: But if we're representing continuous joints as angles on the unit circle we should probably check that they're actually on the unit circle, right? Even if our sampling is correct other things could produce invalid joint states.

roboplan/src/core/scene.cpp
```diff
@@ -145,16 +186,29 @@ void Scene::setRngSeed(unsigned int seed) { rng_gen_ = std::mt19937(seed); }
...
-      const auto& hi = info.limits.max_position[idx];
-      positions(q_idx) = std::uniform_real_distribution<double>(lo, hi)(rng_gen_);
-      ++q_idx;
+    if (info.mimic_info) {
+      continue;  // Skip mimic joints as they are set later.
+    }
+
+    const auto q_idx = model_.idx_qs[model_.getJointId(joint_name)];
+    if (info.type == JointType::CONTINUOUS) {
+      // Special case for continuous joints, since the format is [sin(theta), cos(theta)].
```
eholum: Right?

```suggestion
      // Special case for continuous joints, since the format is [cos(theta), sin(theta)].
```

roboplan/src/core/scene.cpp
```diff
@@ -178,18 +232,76 @@ bool Scene::isValidPose(const Eigen::VectorXd& q) const {
...
+    }
+  }
+
+  applyMimics(q_out);
+  return q_out;
+}
+
 Eigen::VectorXd Scene::interpolate(const Eigen::VectorXd& q_start, const Eigen::VectorXd& q_end,
                                    const double fraction) const {
   return pinocchio::interpolate(model_, q_start, q_end, fraction);
```
eholum, line 316: Ah, so Pinocchio does all the arithmetic for unbounded joints using the Lie group under the hood? Neat. And this will "just work" because we're using the non-mimic model under the hood, right?

roboplan_rrt/src/rrt.cpp
eholum: Ok no, I think issue is just with interpolation for large jumps on continuous joints when path shortcutting. I vote that gets solved later, separate from this PR.

roboplan/src/core/scene.cpp
sea-bass: That's a good idea! Will add

roboplan/src/core/scene.cpp
sea-bass: it does!

roboplan_rrt/src/rrt.cpp
sea-bass: so the display looks fine with the shortcutting turned off?

sea-bass: oh yeah that's confusing... I may need to dig further into what's happening.

eholum: Ah, in regards to the viz problem, it's very subtle but it's actually broken with both shortcut and non-shortcut paths. It's just way more obvious that there are some impossible configurations when there are only two points in the trajectory. 

I'm guessing the problem is with the spline in toppra, which isn't actually handling continuous joints correctly... I think we might have to unwrap continuous joints before running it as mentioned in the todo.

Still. If you want to move on with your life from this PR I think just throw an exception for continuous joints in toppra now and fix it in post.

eholum: Yeah, I threw in a very ugly vibe coded conversion to 1-dof for continuous joints in toppra and it seems to resolve the issue, breaks other things though so 🚀🚀🚀

https://github.com/user-attachments/assets/0fef5459-ae64-492d-a23a-ba3cc59b9636

roboplan_toppra/src/toppra.cpp
```diff
@@ -15,22 +15,35 @@ PathParameterizerTOPPRA::PathParameterizerTOPPRA(const std::shared_ptr<Scene> sc
...
-    const auto& joint_name = joint_names_.at(idx);
+  size_t q_idx = 0;
+  for (const auto& joint_name : joint_names_) {
     const auto& joint_info = scene->getJointInfo(joint_name);
     switch (joint_info.type) {
     case JointType::FLOATING:
     case JointType::PLANAR:
       throw std::runtime_error("Multi-DOF joints not yet supported by TOPP-RA.");
     case JointType::CONTINUOUS:
-      throw std::runtime_error("Continuous joints not yet supported by TOPP-RA.");
```
eholum, line 33: Might be worth just leaving this in for now and handling continuous joints for TOPP-RA in a separate PR? I think there's work needed in either the scene or types classes that should be added to just make these conversions easier. There are probably more tools than just topp-ra that would benefit from the single dof representation of continuous joints?

sea-bass: I'll put this back in for now.

sea-bass: BTW I created this issue to track it: https://github.com/open-planning/roboplan/issues/65

PR #64, opened 2025-09-13
Support joint groups

### LINKED ISSUE #41: Implement notion of joint groups

sea-bass:
This PR adds support for joint groups by parsing the SRDF file and writing joint group information into the scene.

To do this, we bring in `tinyxml2` so I can parse the SRDF file directly with ... not Boost (which is what Pinocchio uses). This contributes to the bulk of the PR.

You can then do something like:

C++
```cpp
const auto arm_group = scene.getJointGroupInfo("arm");
if (arm_group.has_value()) {
  const auto arm_q = full_q(arm_group.value().q_indices);
}
```

Python
```python
arm_group = scene.getJointGroupInfo("arm")
if arm_q is not None:
    arm_q = full_q[arm_group.q_indices];
```

TODO:
- [x] Bring in `tinyxml2` dependency
- [x] Create `JointGroupInfo` class
- [x] Parse the SRDF to get joint group info
- [x] Make IK work with joint groups
- [x] Make RRT work with joint groups
- [x] Make path shortcutting work with joint groups 
- [x] Make TOPP-RA work with joint groups
- [x] Ensure Python bindings are all up to date

Closes #41

### DIFF
--- a/bindings/examples/example_rrt.py
+++ b/bindings/examples/example_rrt.py
@@ -66,7 +66,6 @@ def main(
         package_paths=package_paths,
         yaml_config_path=model_data.yaml_config_path,
     )
-    q_full = scene.getCurrentJointPositions()
     q_indices = scene.getJointGroupInfo(model_data.default_joint_group).q_indices
 
     # Create a redundant Pinocchio model just for visualization.
@@ -92,16 +91,17 @@ def main(
     options.rrt_connect = rrt_connect
     rrt = RRT(scene, options)
 
+    q_full = scene.randomCollisionFreePositions()
+    scene.setJointPositions(q_full)
+
     start = JointConfiguration()
-    start.positions = scene.randomCollisionFreePositions()[q_indices]
+    start.positions = q_full[q_indices]
     assert start.positions is not None
 
     goal = JointConfiguration()
     goal.positions = scene.randomCollisionFreePositions()[q_indices]
     assert goal.positions is not None
 
-    q_full[q_indices] = start.positions
-    scene.setJointPositions(q_full)
     path = rrt.plan(start, goal)
     assert path is not None
 


roboplan/CMakeLists.txt
```diff
@@ -23,9 +23,20 @@ target_include_directories(expected
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
     $<INSTALL_INTERFACE:include>)
 
+# ROS already has a TinyXML2 vendor, so add this only in non-ROS contexts.
+find_package(tinyxml2_vendor QUIET)
+find_package(TinyXML2 REQUIRED)
```
sea-bass: This should also be `QUIET`

roboplan_example_models/models/franka_robot_model/dual_fr3.srdf
```diff
@@ -162,4 +162,10 @@
   <disable_collisions link1="right_fr3_link6" link2="right_fr3_rightfinger" reason="Never"/>
   <disable_collisions link1="right_fr3_link7" link2="right_fr3_rightfinger" reason="Never"/>
   <disable_collisions link1="right_fr3_link8" link2="right_fr3_rightfinger" reason="Never"/>
+
+  <!-- Dual arm-->
```
sea-bass: nit: add space

roboplan/src/core/scene_utils.cpp
```diff
@@ -0,0 +1,137 @@
...
+  tinyxml2::XMLElement* robot = doc.FirstChildElement("robot");
+  if (robot == nullptr) {
+    throw std::runtime_error("No <robot> tag found in the SRDF file!");
+  }
+
+  // Loop through all the "group" elements.
+  for (tinyxml2::XMLElement* group = robot->FirstChildElement("group"); group != nullptr;
+       group = group->NextSiblingElement("group")) {
+    const char* name;
+    group->QueryStringAttribute("name", &name);  // TODO: Check success
```
eholum: Definitely don't want to be in the business of too much format checking for SRDFs, but we could replace all of these queries with something simple that would give a _slightly_ more informative error message if it were missing something? Something like,

```
auto getAttribute = [](const tinyxml2::XMLElement* element, const char* attr_name) -> std::string {
    const char* value = nullptr;
    if (element->QueryStringAttribute(attr_name, &value) != tinyxml2::XML_SUCCESS) {
        throw std::runtime_error("Missing attribute in SRDF: " + std::string(attr_name));
    }
    return std::string(value);
};

const auto name = getAttribute(group, "name");
```

roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -19,12 +32,21 @@ bool SimpleIk::solveIk(const CartesianConfiguration& goal, const JointConfigurat
   }
   auto frame_id = frame_id_result.value();
```
eholum, line 33: Worth checking that the frame_id is part of the joint group?

roboplan/include/roboplan/core/scene.hpp
eholum, line 1: Should `randomCollisionFreePositions` and `randomPositions` be updated to support a joint group? In the dual rrt config planning with the right/left arm still samples the entire joint space.

sea-bass: Yeah... this question extends pretty globally to lots of functionality. We basically have 2 choices:

1. Expand all partial states to full and only call the one unambiguous function
2. Have overloads that accept a group and internally does the necessary expanding/collapsing

Any preference?

I think 2 is better from a usability standpoint, but 1 is better from an efficiency standpoint. That is, encapsulating a lot of the creation/copying of Eigen vectors in functions will cause a lot of unnecessary allocations.

roboplan_simple_ik/src/simple_ik.cpp
sea-bass: I don't think this is necessary. If you're solving IK for a fingertip but only using the arm joints, such a check would fail but IK should still be valid?

roboplan/src/core/scene_utils.cpp
sea-bass: yeah this is a good idea, will do

roboplan_simple_ik/src/simple_ik.cpp
eholum: Yeah and if the tip you choose is just plain wrong that's just user error... WFM.

roboplan_rrt/src/rrt.cpp
```diff
@@ -51,8 +59,30 @@ tl::expected<JointPath, std::string> RRT::plan(const JointConfiguration& start,
...
+  auto q_goal = scene_->getCurrentJointPositions();
+  if (goal.positions.size() == model.nq) {
+    q_goal = goal.positions;  // full joint positions
+  } else if (goal.positions.size() == q_indices.size()) {
+    q_goal(q_indices) = goal.positions;  // group joint positions
+  } else {
+    throw std::runtime_error(
+        "Goal positions is size " + std::to_string(goal.positions.size()) +
+        " which is incompatible with the selected group or full Pinocchio model.");
+  }
```
eholum: This is tied to the question about the scene's API and how joint groups are passed. Basically, users can specify the full joint configuration for both the start + goal poses, but also request that only joints from the specified group are used in planning? But if the start and goal positions have differences in joints that aren't in that group then planning is guaranteed to time out, right? Would it more intuitive for users to only specify positions that are in the group?

roboplan/include/roboplan/core/scene.hpp
eholum: Ugh. I honestly don't know but I always tend to lean towards making things easier for the user? In the case of 2) and reducing copying/creating vectors I feel like a lot of that can be handled with just always using indexed views of the full position vector? Though in the case of 1) who does the expanding when calling `randomCollisionFreePositions`? 

I think my source of confusion comes down to how I should be interacting with the scene. Do I (as the user) ALWAYS use full set of joint states + a group or should I ONLY use the joints from the "active" joint group? Does that make sense? There are some weird in-between things (like my comment down in RRT) that are mildly confusing to me, and I think setting some rule in stone, whichever it is, might be helpful.

Though I am about as bright as a half-watt bulb in a power outage so it's possible that I'm interpreting this wrong... If that's the case please educate me!

roboplan_rrt/src/rrt.cpp
sea-bass: Yeah, you're right...

roboplan/include/roboplan/core/scene.hpp
sea-bass: Nah, you're right. I think leaning towards UX is probably good, and that includes the zen of Python like "there should be one unambiguous way to do it".

roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -37,16 +37,7 @@ bool SimpleIk::solveIk(const CartesianConfiguration& goal, const JointConfigurat
...
-  if (start.positions.size() == model.nq) {
-    q = start.positions;  // full joint positions
-  } else if (start.positions.size() == q_indices.size()) {
-    q(q_indices) = start.positions;  // group joint positions
-  } else {
-    throw std::runtime_error(
-        "Start positions is size " + std::to_string(start.positions.size()) +
-        " which is incompatible with the selected group or full Pinocchio model.");
-  }
+  auto q = scene_->toFullJointPositions(options_.group_name, start.positions);
```
eholum, line 40: Yes.

bindings/examples/example_rrt.py
```diff
@@ -66,6 +66,8 @@ def main(
         package_paths=package_paths,
         yaml_config_path=model_data.yaml_config_path,
     )
+    q_full = scene.getCurrentJointPositions()
```
eholum: I would recommend randomly generating start positions and setting them here, so that you don't end up with invalid configurations.

```
q_full = scene.randomCollisionFreePositions()
```

And then just set the `start.positions` to `q_full` down below. That way things are consistent.

sea-bass: done!

PR #71, opened 2025-09-25
Ensure the sampled points are actually connectable in path shortcutting

### LINKED ISSUE #69: Path shortcutting doesn't seem to avoid collisions consistently
@eholum as I was working on my joint groups PR, I noticed that in some cases the path shortcutting can cause collisions, especially seen in the Dual franka setup.

Notes:
* Confirming this does not happen without shortcutting; the raw paths do not collide in my tests
* I did check whether this was something I did wrong in my PR, and went back to `main` to verify. The following video is on `main`.

https://github.com/user-attachments/assets/584ff7cf-bfd9-421a-921e-4460cf14e1fd

eholum:
Resolves #69 

Tracked the issue down to cases where the sampled shortcut configurations are connectable to each other, but not actually to the later configurations in the path. See [here](https://github.com/open-planning/roboplan/pull/71#discussion_r2382530397).

FWIW I added a seed to the RRT that seemed to produced consistent "bad" shortcut paths so that I could reproduce this.

```
    scene.setRngSeed(1235)
    start = JointConfiguration()
    start.positions = scene.randomCollisionFreePositions()
    assert start.positions is not None

    goal = JointConfiguration()
    goal.positions = scene.randomCollisionFreePositions()
    assert goal.positions is not None

    scene.setRngSeed(123)
    path = rrt.plan(start, goal)
    assert path is not None
```

Before:

https://github.com/user-attachments/assets/48d728f7-ea55-4381-a0cb-a0c3fd292328

After:

https://github.com/user-attachments/assets/9c369bf2-256e-43ef-839d-5fb0d7d7e663

### DIFF
--- a/roboplan/src/core/path_utils.cpp
+++ b/roboplan/src/core/path_utils.cpp
@@ -74,6 +74,9 @@ JointPath PathShortcutter::shortcut(const JointPath& path, double max_step_size,
   std::uniform_real_distribution<double> dis(std::numeric_limits<double>::epsilon(), 1.0);
 
   q_full_ = scene_->getCurrentJointPositions();
+  auto q_start = q_full_;
+  auto q_end = q_full_;
+
   const auto& q_indices = joint_group_info_.q_indices;
   for (unsigned int i = 0; i < max_iters; ++i) {
     if (path_configs.size() < 3) {
@@ -116,14 +119,12 @@ JointPath PathShortcutter::shortcut(const JointPath& path, double max_step_size,
     // However, if  `q_start` and `q_low` or `q_high` and `q_end` are very close to each other,
     // it doesn't make sense to add new configurations. If this is the case, use the existing
     // configuration as the sample.
-    auto q_start = q_full_;
     q_start(q_indices) = path_configs[idx_low - 1];
     if (scene_->configurationDistance(q_start, q_low) < max_step_size) {
       q_low = q_start;
       idx_low--;  // Remove the existing configuration
     }
 
-    auto q_end = q_full_;
     q_end(q_indices) = path_configs[idx_high];
     if (scene_->configurationDistance(q_high, q_end) < max_step_size) {
       q_high = q_end;


roboplan/src/core/path_utils.cpp
```diff
@@ -91,21 +91,44 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
         getConfigurationFromNormalizedPathScaling(scene, shortened_path, path_scalings, low);
     const auto [q_high, idx_high] =
         getConfigurationFromNormalizedPathScaling(scene, shortened_path, path_scalings, high);
-    if (idx_low == idx_high) {
+
+    // TODO: There's an issue with adjacent shortcutting in path reconstruction that needs to be
```
eholum: @sea-bass You're right that we shouldn't need the extra checks. There are some really annoying off by one and edge cases for these that we weren't handling correctly, so unchecked connections (particularly between `path_configs[idx_high]` and `q_high` that were being added. That's how the collisions got there... 

I was poking at this for a bit today but haven't worked out the edge cases yet but I'll get there eventually...

roboplan/src/core/path_utils.cpp
```diff
@@ -91,21 +91,44 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
...
+    if (idx_high <= idx_low + 1) {
       continue;
     }
 
-    // Check if the sampled segment is collision free. If it is, shortcut the path by updating
-    // the configs vector in place. We connect the low and high configurations directly, and
-    // erase the intermediate nodes (if any).
-    if (!hasCollisionsAlongPath(scene, q_low, q_high, max_step_size)) {
-      path_configs[idx_low] = q_low;
-      path_configs[idx_high] = q_high;
```
eholum: This was wrong, if `idx_low + 1 == idx_high` we need to be inserting a point.

roboplan/src/core/path_utils.cpp
```diff
@@ -91,21 +83,43 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
...
-      if (idx_high > idx_low + 1) {
-        path_configs.erase(path_configs.begin() + idx_low + 1, path_configs.begin() + idx_high);
+    // If the indexes are adjacent then we will be adding a segment to the path which will increase
+    // the number of configurations in the path, and could could potentially increase the overall
+    // path length. To ensure that it is worth adding the additional point, we must check that it is
+    // a valid shortcut. For now, this means that the shortcut must be at least a 10% improvement
+    // over the existing segment. This helps to ensure "pointless" shortcuts are not taken.
+    if (idx_high == idx_low + 1) {
```
eholum: @sea-bass this is the kicker from the shortcutting algorithm (I think). Basically it's two things, 1) path shortcutting can add points to a path, and 2) I'm not sure that we actually want the special case for 3 points, because the triangle inequality is real.

Take the path, ` A - B - C - D`, before if we had sampled at points `B'` and `C'` (between `A-B` and `B-C`), then the previous algorithm would drop in  - `A - B' - C' - D`, but `C' - D` is not necessarily a valid segment! Instead what we would want is `A - B' - C' - C - D`, which has more points in the path but should* be shorter because `B' - C'` has to be shorter than `B' - B - C'`.

The issue now is that the number of points in the path could increase (even though the overall length is shorter). Really we should do some kind of decimation or something to ensure that points that are very close to existing points aren't added, but for now I just dropped in a check to make sure that shortcutting is _actually_ worth it and the inequality is greater than a 10% improvement.

eholum: Or turning to my chicken scratch diagram
PXL_20250926_135910482
s

roboplan/src/core/path_utils.cpp
```diff
@@ -91,21 +83,43 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
         getConfigurationFromNormalizedPathScaling(scene, shortened_path, path_scalings, low);
     const auto [q_high, idx_high] =
         getConfigurationFromNormalizedPathScaling(scene, shortened_path, path_scalings, high);
-    if (idx_low == idx_high) {
+
+    // If the indexes are the same then they are on the same path segment so shortening would have
+    // no effect.
```
eholum:
```suggestion
    // Samples are on the same segment so shortening would have no effect.
```

roboplan/src/core/path_utils.cpp
```diff
@@ -91,21 +83,43 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
...
     }
+
+    // We want the new path to be:
+    //
+    // path_configs[idx_low - 1] - > q_low -> q_high -> path_configs[idx_high]
+    //
+    // Because q_low and q_high should be on the valid paths: path_configs[idx_low - 1] ->
+    // path_configs[idx_low] and path_configs[idx_high - 1] -> path_configs[idx_high], we shouldn't
+    // need to check those connections! Assuming that q_low and q_high are directly connectable, we
+    // _should_ just be able to construct a new shorter path.
```
eholum:
```suggestion
    // We want the new path to be:
    //
    // path_configs[idx_low - 1] - > q_low -> q_high -> path_configs[idx_high]
    //
    // Because q_low and q_high should be on the valid paths: 
    // path_configs[idx_low - 1] -> path_configs[idx_low], and
    // path_configs[idx_high - 1] -> path_configs[idx_high], respectively, we shouldn't
    // need to check those connections! We only need to ensure that q_low and q_high
    // are directly connectable.
```

roboplan/src/core/path_utils.cpp
sea-bass: Cool flex with the NASA stationery

roboplan/src/core/path_utils.cpp
```diff
@@ -59,22 +59,14 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
   std::mt19937 gen(seed < 0 ? seed : rd());
   std::uniform_real_distribution<double> dis(std::numeric_limits<double>::epsilon(), 1.0);
 
-  bool path_changed = true;
   for (unsigned int i = 0; i < max_iters; ++i) {
     if (path_configs.size() < 3) {
       // The path is at maximum shortcutted-ness
       return shortened_path;
-    } else if (path_changed && (path_configs.size() == 3)) {
```
sea-bass, line 83: I don't think we need to get rid of this. The logic now is:

1. If path length is 3, try shorten it to 2 if you can.
2. If successful, return the length 2 path (good!)
3. If not, go to a next iteration where you could still add a 4th point to shorten further (per your diagram)

roboplan/src/core/path_utils.cpp
```diff
@@ -91,21 +83,45 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
...
+    // If the indexes are adjacent then we will be adding a segment to the path which will increase
+    // the number of configurations in the path, and could could potentially increase the overall
+    // path length. To ensure that it is worth adding the additional point, we must check that it is
+    // a valid shortcut. For now, this means that the shortcut must be at least a 10% improvement
+    // over the existing segment. This helps to ensure "pointless" shortcuts are not taken.
+    if (idx_high == idx_low + 1) {
+      const auto low_to_existing = scene.configurationDistance(q_low, path_configs[idx_low]);
+      const auto existing_to_high = scene.configurationDistance(path_configs[idx_low], q_high);
+      const auto new_distance = scene.configurationDistance(q_low, q_high);
+      if (low_to_existing + existing_to_high < new_distance * 1.1) {
```
sea-bass: Can you make this a parameter?

What may be needed in this case is to also create a path shortcutting options struct that contains the group name and the fraction.

That or just plain old constructor args. Up to you tbh

roboplan/src/core/path_utils.cpp
```diff
@@ -91,21 +83,45 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
...
 
-    // Check if the sampled segment is collision free. If it is, shortcut the path by updating
-    // the configs vector in place. We connect the low and high configurations directly, and
-    // erase the intermediate nodes (if any).
-    if (!hasCollisionsAlongPath(scene, q_low, q_high, max_step_size)) {
-      path_configs[idx_low] = q_low;
-      path_configs[idx_high] = q_high;
-      if (idx_high > idx_low + 1) {
-        path_configs.erase(path_configs.begin() + idx_low + 1, path_configs.begin() + idx_high);
+    // If the indexes are adjacent then we will be adding a segment to the path which will increase
```
sea-bass:
```suggestion
    // If the indices are adjacent then we will be adding a segment to the path which will increase

File: roboplan/src/core/path_utils.cpp [hunk 9]
```diff
@@ -91,21 +83,45 @@ JointPath shortcutPath(const Scene& scene, const JointPath& path, double max_ste
...
+    // the number of configurations in the path, and could could potentially increase the overall
```
sea-bass:
```suggestion
    // the number of configurations in the path, and could potentially increase the overall

File: roboplan/src/core/path_utils.cpp (hunk 3 above)
Comment (eholum) [reply roboplan/src/core/path_utils.cpp]: Everything i do i do for NASA swag.

File: roboplan/src/core/path_utils.cpp (hunk 6 above)
Comment (sea-bass) [reply roboplan/src/core/path_utils.cpp:83]: I re-added it BTW

File: roboplan/src/core/path_utils.cpp [hunk 10]
```diff
@@ -112,21 +116,47 @@ JointPath PathShortcutter::shortcut(const JointPath& path, double max_step_size,
...
-      path_configs[idx_low] = q_low(q_indices);
-      path_configs[idx_high] = q_high(q_indices);
-      if (idx_high > idx_low + 1) {
-        path_configs.erase(path_configs.begin() + idx_low + 1, path_configs.begin() + idx_high);
+    // If the indices are adjacent, then we will be adding a segment to the path which will increase
+    // the number of configurations in the path, and could potentially increase the overall path
+    // length. To ensure that it is worth adding the additional point, we must check that it is a
+    // valid shortcut. For now, this means that the shortcut must be an improvement over the
+    // existing segment by a factor of `min_shorten_fraction`.
+    // This helps to ensure "pointless" shortcuts are not taken.
```
eholum: Ah man, I think this might still be a dumb bandaid fix and the "right" thing to do is to not add configurations that are very close to each other. That could probably just be the max_step_size or rename the min_shorten_fraction to something else... I'll try that quickly...

roboplan/src/core/path_utils.cpp
eholum: I think if we just put a check at the start to see if the first and last points are directly connectable (like in RRT) then that would cover this case? Is that crazy talk?

roboplan/src/core/path_utils.cpp
```diff
@@ -108,25 +95,50 @@ JointPath PathShortcutter::shortcut(const JointPath& path, double max_step_size,
...
-      path_changed = true;
+    // We generally want the new path to be:
+    //
+    // q_start - > q_low -> q_high -> q_end
+    //
+    // Because q_low and q_high exist on valid segments, we do not need to check the preceding
+    // and following connections. We ONLY need to ensure that q_low and q_high are directly
+    // connectable!
+    //
+    // However, if  `q_start` and `q_low` or `q_high` and `q_end` are very close to each other,
```
eholum, line 119: Ok. Do this and everything just works. I consistently get down to two points, but if we wanted the check right at the start to see if the start/end configurations are visible we could also throw that in.

bindings/examples/example_rrt.py
```diff
@@ -109,7 +109,10 @@ def main(
     if include_shortcutting:
         shortcutter = PathShortcutter(scene, model_data.default_joint_group)
         shortened_path = shortcutter.shortcut(
-            path, options.collision_check_step_size, 1000
+            path,
+            max_step_size=options.collision_check_step_size,
+            max_iters=1000,
+            min_shorten_fraction=0.9,
```
eholum:
```suggestion
```

roboplan/src/core/path_utils.cpp
```diff
@@ -112,48 +95,42 @@ JointPath PathShortcutter::shortcut(const JointPath& path, double max_step_size,
...
+    // and following connections. We ONLY need to ensure that q_low and q_high are directly
+    // connectable!
+    auto q_start = q_full_;
+    q_start(q_indices) = path_configs[idx_low - 1];
+    if (scene_->configurationDistance(q_start, q_low) < max_step_size) {
+      q_low = q_start;
+      idx_low--;  // Remove the existing configuration
+    }
+
+    auto q_end = q_full_;
```
sea-bass: if we're using `q_start` and `q_full` in a loop, makes sense to allocate them up front. I can change that

PR #73, opened 2025-09-28
Support continuous joints in RRT and TOPP-RA

### LINKED ISSUE #65: Add utility to collapse and expand positions for continuously rotating joints
We found that TOPP-RA doing trajectory timing for continuous joints requires "collapsing" the joint angle to a single-DOF before uncollapsing it again.

This issue tracks creating such a function, which could be something like:

```cpp
Eigen::VectorXd collapseContinuousJointPositions(const Scene& scene, const Eigen::VectorXd& q_orig);

Eigen::VectorXd expandContinuousJointPositions(const Scene& scene, const Eigen::VectorXd& q_collapsed);
```

This would affect all joints with continuous rotation, which include continuous and (the rotation DOF of) planar joints.

sea-bass:
This was painful.

[kinova_rrt.webm](https://github.com/user-attachments/assets/633b6c7d-355d-4dbc-877e-f89a64910551)

[kinova_rrt_shortcut.webm](https://github.com/user-attachments/assets/300e95b7-87f3-43a1-9d1b-69135ca19a1b)

Note: Sometimes it appears that shortcutting decides to "take the long way around". Probably because of the way distances are measured. But I think this is OK for now.

Closes #65

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -120,3 +120,7 @@ jobs:
           pixi run install_all
           pixi run build_tests
           pixi run test_all
+      # TODO: Decide on a method for automated benchmark performance testing.
+      - name: Run benchmarks
+        run: |
+          pixi run test_benchmarks
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@
 # pixi environments
 .pixi
 *.egg-info
+
+# Benchmarking results
+pytest_benchmarks.json
--- a/benchmarks/test_benchmark_rrt.py
+++ b/benchmarks/test_benchmark_rrt.py
@@ -1,0 +1,95 @@
+import pytest
+import sys
+import xacro
+
+from roboplan import (
+    JointConfiguration,
+    Scene,
+    RRTOptions,
+    RRT,
+)
+
+# We don't build the bindings examples, so we just include the relative
+# directory manually.
+from pathlib import Path
+
+examples_dir = Path(__file__).parent.parent / "bindings" / "examples"
+sys.path.insert(0, str(examples_dir))
+
+from common import MODELS, ROBOPLAN_EXAMPLES_DIR
+
+
+def solve(scene: Scene, rrt: RRT, seed: int = 1234):
+    """
+    Runs an RRT test by sampling random, collision-free joint configurations
+    then attempting to plan a path between them.
+
+    Returns 1 if planning was successful, 0 otherwise.
+    """
+    scene.setRngSeed(seed)
+
+    start = JointConfiguration()
+    start.positions = scene.randomCollisionFreePositions()
+    assert start.positions is not None
+
+    goal = JointConfiguration()
+    goal.positions = scene.randomCollisionFreePositions()
+    assert goal.positions is not None
+
+    path = rrt.plan(start, goal)
+    return 0 if path is None else 1
+
+
+def solve_many(scene: Scene, rrt: RRT, iterations: int = 10):
+    """
+    Runs the specified number of iterations of RRT with a random seed.
+
+    Returns the number of successful solves.
+    """
+    successes = 0
+    for i in range(iterations):
+        successes += solve(scene, rrt, 11235 + i)
+    return successes
+
+
+def create_scene(model_name: str) -> Scene:
+    model_data = MODELS[model_name]
+    package_paths = [ROBOPLAN_EXAMPLES_DIR]
+
+    urdf_xml = xacro.process_file(model_data.urdf_path).toxml()
+    srdf_xml = xacro.process_file(model_data.srdf_path).toxml()
+
+    scene = Scene(
+        f"{model_name}_benchmark_scene",
+        urdf=urdf_xml,
+        srdf=srdf_xml,
+        package_paths=package_paths,
+        yaml_config_path=model_data.yaml_config_path,
+    )
+    return scene
+
+
+@pytest.fixture(scope="session", params=["kinova", "ur5", "franka", "dual"])
+def scene(request):
+    return create_scene(request.param)
+
+
+def test_benchmark_rrt(benchmark, scene):
+    options = RRTOptions()
+    options.max_nodes = 100000
+    options.max_planning_time = 10.0
+    rrt = RRT(scene, options)
+
+    success_rate = benchmark(solve_many, scene, rrt, 10)
+    assert success_rate >= 0.95
+
+
+def test_benchmark_rrt_connect(benchmark, scene):
+    options = RRTOptions()
+    options.max_nodes = 100000
+    options.rrt_connect = True
+    options.max_planning_time = 10.0
+    rrt = RRT(scene, options)
+
+    success_rate = benchmark(solve_many, scene, rrt, 10)
+    assert success_rate >= 0.95
--- a/pixi.lock
+++ b/pixi.lock
@@ -284,6 +284,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pthread-stubs-0.4-hb9d3cd8_1002.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pugixml-1.15-h3f63f65_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pulseaudio-client-17.0-hac146a9_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/py-cpuinfo-9.0.0-pyhd8ed1ab_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/py-opencv-4.12.0-qt6_py311h5956852_600.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/pycparser-2.22-pyh29332c3_1.conda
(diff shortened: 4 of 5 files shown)


bindings/examples/example_rrt.py
```diff
@@ -113,31 +118,28 @@ def main(
...
             "/rrt/shortcut_path",
         )
-        path = shortened_path
-
-    # Set up TOPP-RA path parameterization
-    dt = 0.01
-    toppra = PathParameterizerTOPPRA(scene, model_data.default_joint_group)
-    traj = toppra.generate(path, dt)
+    else:
+        visualizePath(viz, scene, traj, model_data.ee_names, 0.05)
```
sea-bass, line 148: This is an important note: Sometimes you can get mismatches if you display `path` vs. `traj` (with continuous joints), so ensuring to plot the trajectory always.

But also, note to self to plot both `path` and `traj` to help isolate the source of issues

roboplan/src/core/scene_utils.cpp
```diff
@@ -123,7 +124,18 @@ std::map<std::string, JointGroupInfo> createJointGroupInfo(const pinocchio::Mode
       for (int dof = 0; dof < joint.nv(); ++dof) {
         v_indices.push_back(v_idx + dof);
       }
+
+      // Check for any continuous degrees of freedom.
+      const auto joint_type = kPinocchioJointTypeMap.at(joint.shortname());
+      if (joint_type == JointType::CONTINUOUS || joint_type == JointType::PLANAR) {
```
eholum, line 131: Wait, why do we care about planar joints here?

roboplan/src/core/scene_utils.cpp
```diff
@@ -143,9 +155,106 @@ std::map<std::string, JointGroupInfo> createJointGroupInfo(const pinocchio::Mode
       .joint_names = std::vector<std::string>(model.names.begin() + 1, model.names.end()),
       .joint_indices = all_joint_indices,
       .q_indices = Eigen::VectorXi::LinSpaced(model.nq, 0, model.nq - 1),
-      .v_indices = Eigen::VectorXi::LinSpaced(model.nv, 0, model.nv - 1)};
+      .v_indices = Eigen::VectorXi::LinSpaced(model.nv, 0, model.nv - 1),
+      .has_continuous_dofs = false,
```
eholum: I think you want this to be true if any of the groups above have a continuous joint?

roboplan_toppra/src/toppra.cpp
```diff
@@ -7,37 +7,47 @@
...
 
 #include <roboplan/core/path_utils.hpp>
+#include <roboplan/core/scene_utils.hpp>
 #include <roboplan_toppra/toppra.hpp>
 
 namespace roboplan {
 
 PathParameterizerTOPPRA::PathParameterizerTOPPRA(const std::shared_ptr<Scene> scene,
-                                                 const std::string& group_name) {
+                                                 const std::string& group_name)
```
eholum, line 16: Maybe the default for this should just be `""` so that it uses the `all` joint group out of the box?

roboplan_rrt/src/rrt.cpp
```diff
@@ -28,21 +34,22 @@ RRT::RRT(const std::shared_ptr<Scene> scene, const RRTOptions& options)
...
     case JointType::FLOATING:
     case JointType::PLANAR:
       throw std::runtime_error("Multi-DOF joints not yet supported by RRT.");
     case JointType::CONTINUOUS:
-      // TODO: Use "SO2" state space instead.
-      // The solution should be to squash the position vectors so continuous
-      // joints are also represented as single-DOF.
-      state_space_names.push_back("Rn:2");
+      // The solution squashes the continuous position vectors to be used as an SO(2).
+      state_space_names.push_back("SO2");
```
eholum, line 49: Assuming the kd-tree is actually using `SO2` for calculations then aren't the actual limits -PI to PI? It looks like it from the implementation of `SO2<Scalar>`, https://github.com/open-planning/roboplan/blob/main/roboplan_rrt/include/dynotree/StateSpace.h#L227

Definitely a PITA to deal with different representations of the configuration space... Someday it'll probably be worth making a different KD graph that just works with Pinocchios representation. For now I think it'll definitely be worth putting a comment in that mentions what version is being used where:

1. Scene - Uses full joint configurations with continuous joints in the `[cos(theta), sin(theta)]` repr.
2. KD-Tree - Uses exclusively the joint group's joints with `theta` in SO(2).

So anytime we're querying the search graph we need to make sure we're using collapsed joint group states and anytime we're querying the scene we need to use exploded full joint states. Does that sum it up accurately?

roboplan_toppra/src/toppra.cpp
```diff
@@ -94,10 +104,16 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
   std::vector<double> steps;
   steps.reserve(num_pts);
   double s = 0.0;
-  for (const auto& pos : path.positions) {
-    path_pos_vecs.push_back(pos);
+  for (size_t idx = 0; idx < path.positions.size(); ++idx) {
+    const auto& pos = path.positions.at(idx);
+    const auto maybe_collapsed_pos = collapseContinuousJointPositions(*scene_, group_name_, pos);
```
eholum: Ok I dug into your comment about path shortcutting because it's not good if things disagree about which directions to go for continuous joints. First off, I Pinocchio's interpolation definitely seems to interpolate correctly and go the "short" way. So we are correctly determining that two points are connectable in the path shortcutter.

The issue (I _think_) is that this is going to do weird things to the spline below. Right? Using random seed `1237` for joint sampling I get this path and shortcut,

Image

Here's the path:

```
Planning...
 Found goal with 19 sampled nodes!
Joint Path with 8 points:
  1:  -0.31368 -0.949529   1.00145  0.954418 -0.298473  -1.71814 -0.999742 0.0227032  -1.72083  0.992094 -0.125499
  2: -0.626147 -0.779705   1.16756   0.75141 -0.659835  -1.12632  -0.98782  0.155601  -1.44535  0.939417  0.342776
  3: -0.672116 -0.740446   1.15016  0.290117 -0.956991  -1.27894 -0.670955  0.741498   -1.0649  0.970491  0.241135
  4: -0.862757 -0.505619  0.970608  0.331264 -0.943538  -1.66557 -0.573895  0.818929  -1.15267  0.472733  0.881206
  5: -0.968488 -0.249061   1.20722 0.0381521 -0.999272  -0.87018 -0.646358  0.763034 -0.941736  0.192308  0.981335
  6:  -0.999644  0.0266746    1.44382  -0.258356   -0.96605 -0.0747921  -0.713408   0.700748  -0.730805  -0.105168   0.994454
  7: -0.953827  0.300356   1.68043 -0.531866 -0.846829  0.720595 -0.774484  0.632594 -0.519873 -0.393319  0.919402
  8: -0.913062  0.407821   1.77817 -0.632298 -0.774725   1.04916 -0.797848  0.602859 -0.432739  -0.50355  0.863966

Shortcutted path:
Joint Path with 2 points:
  1:  -0.31368 -0.949529   1.00145  0.954418 -0.298473  -1.71814 -0.999742 0.0227032  -1.72083  0.992094 -0.125499
  2: -0.913062  0.407821   1.77817 -0.632298 -0.774725   1.04916 -0.797848  0.602859 -0.432739  -0.50355  0.863966
```

So just looking at joint 1 we have `atan2(-0.949, -0.313) = -1.88` to `atan2(0.407, -0.913) = 2.722`. The problem is that pinocchio went the short way (wrapping around -PI) in its collision checking, but the spline doesn't take the wrapping into account so it just goes through `0`. 

So we need to do one of,

1. Tell the TOPP-RA that continuous joints are wrappable (if it supports it but doubt)
2. Unwrap continuous joints and check distances to preceding points and wrap them ourselves before passing to the spline
3. Add non-zero velocity support for intermediate poses and interpolate the trajectories ourselves before passing to TOPP-RA (this might be a good choice period)
4.  Something else?

eholum: Ok I tried it in https://github.com/open-planning/roboplan/commit/b0548db288699f557e17caa5530088a33e289471 and added the[ rng seed arg](https://github.com/open-planning/roboplan/commit/94d1d63ff847a956252f8c9c120323ca0f48c905) to the example rrt. FWIW it seems to do the thing, but I'm still not sure what the "right answer" is.

```
python3 bindings/examples/example_rrt.py --host ubuntu-linux.shared --no-rrt-connect --include-shortcutting  --model kinova --rng-seed 1237
```

image

roboplan/src/core/scene_utils.cpp
sea-bass: Planar joints also have a continuous dof for the rotation part, it's `[x, y, cos(theta), sin(theta)]`.

I've also incorporated this into the expand/collapse functions.

roboplan_rrt/src/rrt.cpp
sea-bass: It does sum it up correctly!

Far as I know limits are ignored in dynotree SO2 spaces

roboplan_toppra/src/toppra.cpp
sea-bass: I had something like this in before, but didn't do the appropriate analysis. This seems fine to add.

sea-bass: The only difference is that I pre-computed the indices for continuous joints once at init time to make the computation more efficient

PR #75, opened 2025-09-30
Add basic unit test for continuous and mimic joints

### LINKED ISSUE #66: Add unit tests that include mimics and continuous joints
Just using the kinova would be a great choice.

As part of this we also need to to throw in a utility function for parsing xacro for C++. Honestly I think just wrapping the system call might be the easiest/dumbest thing to do... it's the same thing we do on the python side.

eholum:
Resolves #66.

Just covering the lowest level scene functions here for mimic/continuous joints. I think between this and the benchmarks we can be reasonably certain that things are working "as expected".


roboplan/test/CMakeLists.txt
```diff
@@ -20,6 +20,15 @@ target_link_libraries(
   roboplan_example_models::roboplan_example_models
 )
 
+add_executable(test_joints test_joints.cpp)
```
sea-bass, line 23: I think you will also need to add this below line 45

roboplan/test/test_joints.cpp
```diff
@@ -0,0 +1,128 @@
...
+  Eigen::VectorXd pos(4);
+  pos << 0.7071067, -0.7071067, 0.5, 0.5;  // -45 degrees
+
+  // Collapse the continuous joint
+  Eigen::VectorXd expected_collapsed(3);
+  expected_collapsed << -0.785398, 0.5, 0.5;
+  const auto collapsed = collapseContinuousJointPositions(*scene_, "", pos);
+  ASSERT_TRUE(collapsed.has_value());
+  ASSERT_EQ(collapsed->size(), 3);
+  EXPECT_FLOAT_EQ((*collapsed)(0), expected_collapsed(0));
```
sea-bass: What if you do e.g. `collapsed->at(0)` etc.? And same throughout the rest of the file

eholum: what even is a vector?

roboplan/test/CMakeLists.txt
eholum: lol yep.

roboplan/test/test_joints.cpp
sea-bass: Ah crap this is an Eigen vector isn't it? My comment may not be valid

roboplan/test/test_joints.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+TEST_F(RoboPlanJointTest, ExpandCollapse) {
+  Eigen::VectorXd pos(4);
+  pos << 0.7071067, -0.7071067, 0.5, 0.5;  // -45 degrees
+
+  // Collapse the continuous joint
+  Eigen::VectorXd expected_collapsed(3);
+  expected_collapsed << -0.785398, 0.5, 0.5;
+  const auto collapsed = collapseContinuousJointPositions(*scene_, "", pos).value();
```
sea-bass: This test may crash with bad expected access.

Worth doing an

```cpp
ASSERT_TRUE(maybe_collapsed.has_value()) << maybe_collapsed.error();
const auto& collapsed = maybe_collapsed.value();

File: roboplan/test/CMakeLists.txt [hunk 4]
```diff
@@ -36,8 +45,12 @@ find_package(ament_cmake_gtest QUIET)
 if ( ament_cmake_gtest_FOUND )
     ament_add_gtest_test(test_types)
     ament_add_gtest_test(test_scene)
+    ament_add_gtest_test(test_joints)
     ament_add_gtest_test(test_path_utils)
 else()
     include(GoogleTest)
-    gtest_discover_tests(test_types test_scene test_path_utils)
```
eholum, line 42: Ah, `pixi run test roboplan` was only picking up the types test. This was interpreting the later tests as arguments to `test_types`,

https://cmake.org/cmake/help/latest/module/GoogleTest.html#command:gtest_discover_tests

eholum: Ok now it's actually running all the tests,

https://github.com/open-planning/roboplan/actions/runs/18133642807/job/51606539987?pr=75#step:5:237

sea-bass: Oh dang. This may be needed in other places but I can take a look

eholum: I checked I I think it's all good,

```
$ grep -r gtest_discover_tests *
roboplan/test/CMakeLists.txt:    gtest_discover_tests(test_types)
roboplan/test/CMakeLists.txt:    gtest_discover_tests(test_scene)
roboplan/test/CMakeLists.txt:    gtest_discover_tests(test_joints)
roboplan/test/CMakeLists.txt:    gtest_discover_tests(test_path_utils)
roboplan_rrt/test/CMakeLists.txt:    gtest_discover_tests(test_rrt)
roboplan_toppra/test/CMakeLists.txt:    gtest_discover_tests(test_toppra)
```

PR #77, opened 2025-10-02
Remove prints in RRT source and add them to example

sea-bass:


bindings/examples/example_rrt.py
```diff
@@ -99,6 +99,8 @@ def main(
 
     q_full = scene.randomCollisionFreePositions()
     scene.setJointPositions(q_full)
+    viz.display(q_full)
+    time.sleep(0.1)
```
sea-bass, line 103: this also makes it so the arm displays _before_ a (potentially long) RRT planning cycle

PR #80, opened 2025-10-12
Support collision objects in scene

### LINKED ISSUE #40: Add functionality to add, remove, and update obstacles
This involves taking geometric primitives and placing them in the scene relative to some parent object (robot or world).

Includes also adding a few obstacles to the robot arm examples to really showcase RRT.

Specific to parenting objects to the robot, this requires also an ability to specify allowable collisions. For example, attaching a tool to the end effector should exclude the end effector from its collision pairs.

The "add" part exists in Python here for reference: https://github.com/sea-bass/pyroboplan/blob/eb1fa4a40c16dbb85a82f4037ac80d213e5f6973/src/pyroboplan/models/panda.py#L82, whereas updating and deleting are probably easier to do

sea-bass:
Also took the liberty of repositioning the dual Franka URDFs to be less awkward.

Closes #40

### DIFF
--- a/roboplan_example_models/models/franka_robot_model/fr3.xacro
+++ b/roboplan_example_models/models/franka_robot_model/fr3.xacro
@@ -369,6 +369,120 @@
       <dynamics damping="0.3"/>
     </joint>
 
+  <!-- ros2_control setup for mock hardware -->
+  <ros2_control name="${prefix}name" type="system">
+    <hardware>
+      <plugin>mock_components/GenericSystem</plugin>
+      <param name="mock_sensor_commands">false</param>
+      <param name="state_following_offset">0.0</param>
+      <param name="calculate_dynamics">true</param>
+    </hardware>
+    <joint name="${prefix}fr3_joint1">
+      <command_interface name="position"/>
+      <command_interface name="velocity"/>
+      <state_interface name="position">
+        <param name="initial_value">0.0</param>
+      </state_interface>
+      <state_interface name="velocity">
+        <param name="initial_value">0.0</param>
+      </state_interface>
+      <state_interface name="effort">
+        <param name="initial_value">0.0</param>
+      </state_interface>
+    </joint>
+    <joint name="${prefix}fr3_joint2">
+      <command_interface name="position"/>
+      <command_interface name="velocity"/>
+      <state_interface name="position">
+        <param name="initial_value">${-pi/4}</param>
+      </state_interface>
+      <state_interface name="velocity">
+        <param name="initial_value">0.0</param>
+      </state_interface>
+      <state_interface name="effort">
+        <param name="initial_value">0.0</param>
+      </state_interface>
+    </joint>
+    <joint name="${prefix}fr3_joint3">
+      <command_interface name="position"/>
+      <command_interface name="velocity"/>
+      <state_interface name="position">
+        <param name="initial_value">0.0</param>
+      </state_interface>
+      <state_interface name="velocity">
+        <param name="initial_value">0.0</param>
+      </state_interface>
+      <state_interface name="effort">
+        <param name="initial_value">0.0</param>
+      </state_interface>
+    </joint>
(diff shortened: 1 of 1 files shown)


roboplan/include/roboplan/core/scene.hpp
```diff
@@ -148,10 +149,29 @@ class Scene {
...
   /// @brief Prints basic information about the scene.
   friend std::ostream& operator<<(std::ostream& os, const Scene& scene);
 
 private:
+  /// @brief Adds a Pinocchio geometry object to the scene.
+  /// @details This is private for now, but can be made the sole public entrypoint once
+  /// Pinocchio and Coal have working nanobind bindings compatible with this library.
+  /// @param geom_obj The geometry object instance to add.
+  /// @return Void if successful, else a string describing the error.
+  tl::expected<void, std::string> addGeometry(const pinocchio::GeometryObject& geom_obj);
```
eholum, line 180: Ah yes so this will be SO much easier once it is supported in the underlying library? Although boxes should always be sufficient, as Mr. Joe always says.

bindings/include/utils/expected.hpp
```diff
@@ -12,7 +12,11 @@
 /// @throw std::runtime_error if the result is an error.
 template <typename Expected> typename Expected::value_type handle_expected(Expected&& result) {
   if (result.has_value()) {
-    return result.value();
+    if constexpr (std::is_void_v<typename Expected::value_type>) {
+      return;  // Return nothing since it's a void type
+    } else {
+      return result.value();
+    }
```
eholum, line 19: It's brilliant holmes.

roboplan/include/roboplan/core/scene.hpp
sea-bass: I also added spheres because we must always assume a spherical cow. But I'm gonna leave it at that for now.

roboplan/src/core/scene.cpp
```diff
@@ -353,6 +353,95 @@ Eigen::VectorXi Scene::getJointPositionIndices(const std::vector<std::string>& j
...
+  collision_geom.placement = pinocchio::SE3(tform);
+  return {};
+}
+
+tl::expected<void, std::string> Scene::removeGeometry(const std::string& name) {
+  auto it = collision_geometry_map_.find(name);
+  if (it == collision_geometry_map_.end()) {
+    return tl::make_unexpected("Could not find object '" + name + "' to remove.");
+  }
+  collision_model_.removeGeometryObject(name);
```
eholum, line 439: I'm assuming this will handle removing all of the collision pairs that were added to the model?

bindings/examples/example_rrt.py
```diff
@@ -79,6 +81,13 @@ def main(
     visual_model = pin.buildGeomFromUrdfString(
         model, urdf_xml, pin.GeometryType.VISUAL, package_dirs=package_paths
     )
+
+    # Optionally add obstacles
+    if include_obstacles:
+        for obstacle in model_data.obstacles:
+            obstacle.addToScene(scene)
+            obstacle.addToPinocchioModels(model, collision_model, visual_model)
```
eholum, line 91: Might be worth just adding something like this to remind users that are two models at play here. The comment above won't necessarily be obvious and if you're removing objects you have to remember to remove it from both,

```suggestion
            # Add the obstacle as both a collision object in the scene's model AND to the viz models for rendering.
            obstacle.addToScene(scene)
            obstacle.addToPinocchioModels(model, collision_model, visual_model)
```

roboplan/src/core/scene.cpp
sea-bass: Yes, it does.

https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/structpinocchio_1_1GeometryModel.html#aaafa899316a8f4bbb722e40bc169a2b8

PR #85, opened 2025-10-15
Allow setting collision pairs

### LINKED ISSUE #52: Add functionality to enable/disable collision pairs
Basically the C++ versions of something like this:

https://github.com/sea-bass/pyroboplan/blob/eb1fa4a40c16dbb85a82f4037ac80d213e5f6973/src/pyroboplan/core/utils.py#L379-L476

sea-bass:
My call tonight was canceled so I actually had time to see this through.

Closes #52.

image

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -29,7 +29,7 @@ jobs:
       matrix:
         ubuntu_version: [22.04, 24.04]
 
-    runs-on: self-hosted  #ubuntu-latest
+    runs-on: ubuntu-latest
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
@@ -64,7 +64,7 @@ jobs:
       matrix:
         ros_distro: [humble, jazzy, kilted, rolling]
 
-    runs-on: self-hosted  #ubuntu-latest
+    runs-on: ubuntu-latest
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
@@ -97,11 +97,10 @@ jobs:
       SCCACHE_DIR: ${{ github.workspace }}/.sccache
     strategy:
       fail-fast: false
-      # TODO: Re-enable when we do not have self-hosted runners in the public repo
-      # matrix:
-      #   ubuntu_version: [22.04, 24.04]
+      matrix:
+        ubuntu_version: [22.04, 24.04]
 
-    runs-on: self-hosted  #ubuntu-${{ matrix.ubuntu_version }}
+    runs-on: ubuntu-${{ matrix.ubuntu_version }}
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


bindings/examples/common.py
```diff
@@ -101,17 +108,26 @@ class RobotModelConfig:
         obstacles=[
             ObstacleConfig(
                 name="test_box",
-                geom=hppfcl.Box(1.0, 1.0, 0.5),
+                geom=hppfcl.Box(0.5, 0.5, 0.5),
```
eholum, line 111: https://www.youtube.com/watch?v=yD67ClenBEY

roboplan/src/core/scene.cpp
```diff
@@ -221,12 +221,26 @@ std::optional<Eigen::VectorXd> Scene::randomCollisionFreePositions(size_t max_sa
...
   pinocchio::updateGeometryPlacements(model_, model_data_, collision_model_, collision_model_data_,
                                       q);
-  return pinocchio::computeCollisions(model_, model_data_, collision_model_, collision_model_data_,
-                                      q,
-                                      /* stop_at_first_collision*/ true);
+  const auto result =
+      pinocchio::computeCollisions(model_, model_data_, collision_model_, collision_model_data_, q,
+                                   /* stop_at_first_collision*/ !debug);
+
+  if (debug) {
```
eholum, line 231: Someday, not now, but someday we should wrap `computeDistances`. Maybe when python bindings are released?

roboplan/src/core/scene.cpp
```diff
@@ -442,6 +456,62 @@ tl::expected<void, std::string> Scene::removeGeometry(const std::string& name) {
...
+  for (const auto& body1_id : body1_collision_geom_ids) {
+    for (const auto& body2_id : body2_collision_geom_ids) {
+      const auto pair = pinocchio::CollisionPair(body1_id, body2_id);
+      if (enable) {
+        collision_model_.addCollisionPair(pair);
+      } else {
+        collision_model_.removeCollisionPair(pair);
+      }
+    }
+  }
```
eholum, line 510: Any issues adding duplicates? Will Pinocchio reject it and not do any work if these pairs already exist/do not exist?

sea-bass: I think Pinocchio throws if you add a duplicate geometry/frame name...

bindings/examples/common.py
sea-bass: why shrink this?

roboplan/src/core/scene.cpp
sea-bass: oh I am reading this again -- yeah I think it's fine.

bindings/examples/common.py
sea-bass: oh this is just the diff haha

roboplan/src/core/scene.cpp
sea-bass: yes -- as and when we need it.

Would be readily handy for debugging, but also for collision-avoiding IK/servoing stuff down the line.

bindings/examples/common.py
eholum: Oh yeah sorry I just had to get my Simpsons quote in because I'm a millennial.

PR #86, opened 2025-10-21
Add collision checking, random restarts, and max time to simple IK solver

### LINKED ISSUE #74: Validate IK by collision checking and add random restarts
Right now, the simple IK does not return collision-free results.

The simplest possible approach is the "KDL" style, which does this via rejection sampling. In other words:
1. Solve IK to completion
2. Check for collisions; if no collisions return, if collisions, try again from a random restart
3. Repeat for several random restart attempts / for some time budget

sea-bass:
The random restarts aren't super cleanly seedable in a thread-safe way because it relies on the scene's RNG instead of some other optional input, but this is fine ™️  for now.

Closes #74

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -29,7 +29,7 @@ jobs:
       matrix:
         ubuntu_version: [22.04, 24.04]
 
-    runs-on: self-hosted  #ubuntu-latest
+    runs-on: ubuntu-latest
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
@@ -64,7 +64,7 @@ jobs:
       matrix:
         ros_distro: [humble, jazzy, kilted, rolling]
 
-    runs-on: self-hosted  #ubuntu-latest
+    runs-on: ubuntu-latest
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -39,31 +43,51 @@ bool SimpleIk::solveIk(const CartesianConfiguration& goal, const JointConfigurat
...
-      solution.positions = q(q_indices);
-      return true;
-    }
+      const auto error = pinocchio::log6(goal_tform.actInv(data_.oMf[frame_id])).toVector();
 
-    pinocchio::computeFrameJacobian(model, data_, q, frame_id, pinocchio::ReferenceFrame::LOCAL,
-                                    full_jacobian_);
-    jacobian_ = full_jacobian_(Eigen::placeholders::all, v_indices);
+      if (error.norm() <= options_.max_error_norm) {
+        if (options_.check_collisions && !scene_->hasCollisions(q)) {
```
eholum: This is wrong since we want it to set if collision checking is disabled. This broke my brain for second but I _think_ this is it?

```suggestion
        if (!options_.check_collisions || !scene_->hasCollisions(q)) {
```

roboplan_simple_ik/include/roboplan_simple_ik/simple_ik.hpp
```diff
@@ -24,6 +30,9 @@ struct SimpleIkOptions {
   /// @brief The maximum error norm.
   /// TODO: Separate into linear and angular?
   double max_error_norm = 0.001;
+
+  /// @brief Whether to check collisions.
+  bool check_collisions = true;
```
eholum, line 35: It's worth adding these options to the example too!

roboplan_simple_ik/src/simple_ik.cpp
sea-bass: I am a dishonor to my bloodline

roboplan_simple_ik/include/roboplan_simple_ik/simple_ik.hpp
sea-bass: done

PR #88, opened 2025-11-01
Fix applyMimics binding by copying and returning a value

eholum:
Ran into this in playing with the python bindings. 

Modifying the array in place isn't guaranteed because nanobind will copy the underlying data into a C object. I think it's probably safer to just copy and return an updated value.

https://nanobind.readthedocs.io/en/latest/eigen.html#python-c

Updated the tests to be 100% sure because I was going a little crazy over this.

### DIFF
--- a/bindings/test/test_scene.py
+++ b/bindings/test/test_scene.py
@@ -211,7 +211,7 @@ def test_set_collisions(test_scene: Scene) -> None:
 
 
 def test_mimics() -> None:
-    # Equivalend to the C++ test, but the updated joint state is returned as a new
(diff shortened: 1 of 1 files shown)


bindings/test/test_scene.py
```diff
@@ -165,3 +208,13 @@ def test_set_collisions(test_scene: Scene) -> None:
         "Frame name 'nonexistent_link' not found in frame_map_."
     )
     assert str(exc_info.value) == expected_error
+
+
+def test_mimics() -> None:
+    # Equivalend to the C++ test, but the updated joint state is returned as a new
```
sea-bass: Typo: equivalenT

PR #89, opened 2025-11-06
Update to open-planning fork of toppra

sea-bass:
Updates toppra from a personal fork to a branch that we actually would like to land upstream: 

https://github.com/open-planning/toppra/tree/colcon-workaround

https://github.com/hungpham2511/toppra/pull/266

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -19,7 +19,6 @@ configure = { cmd = [
   "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON",
   "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
   "-DBUILD_TESTING=ON",
-  "-DPYTHON_BINDINGS=OFF",
   # Use clang as the compiler
   "-DCMAKE_C_COMPILER=clang",
   "-DCMAKE_CXX_COMPILER=clang++",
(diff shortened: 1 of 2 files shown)


scripts/build_cmake.bash
```diff
@@ -2,6 +2,8 @@
 
 # Helper script that builds all the packages using vanilla CMake.
 
+set -e  # Needed to ensure cmake failures propagate up
```
sea-bass, line 5: this was why our "vanilla cmake" workflows were swallowing build errors! 🤦🏻

eholum: lol derp :/

pixi.toml
```diff
@@ -19,6 +19,7 @@ configure = { cmd = [
   "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON",
   "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
   "-DBUILD_TESTING=ON",
+  "-DPYTHON_BINDINGS=OFF",
```
eholum: Wait... For topp-ra specifically?

sea-bass: Yes. It doesn't (yet) hurt other packages to have it be a blanket argument, but we have other options such as

1. Special case this arg just for toppra (would make the pixi stuff even more complicated)
2. Default this flag to OFF in our fork of toppra (what was happening prior to this PR)

sea-bass: actually, I found a way to do 1. that isn't terribly offensive.

PR #90, opened 2025-11-07
Add initial ReadTheDocs setup

### LINKED ISSUE #83: Figure out a documentation workflow
This involves something like a readthedocs site, plus autogenerated API docs (C++ and Python, somehow?)

sea-bass:
This PR adds a ReadTheDocs based docs page, which also includes Doxygen generated API stuff using breathe: https://breathe.readthedocs.io/en/latest/index.html

It doesn't look great, but it's a start.

See https://roboplan--90.org.readthedocs.build/en/90/index.html

Punch list:

- [x] Set up ReadTheDocs
- [x] Move design philosophy in README to docs
- [x] Move installation steps in README to docs
- [x] Add Doxyfiles for each package with C++ code
- [x] Automate stub generation for bindings (https://nanobind.readthedocs.io/en/latest/typing.html#stubs) and also add to docs
- [x] Add link to Discord group

Closes #83

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -1,7 +1,7 @@
 Design Philosophy
 =================
 
-RoboPlan is a motion planning framework, similarly to tools like `MoveIt <https://github.com/moveit/moveit2>`_.
+RoboPlan is a motion planning framework, similar to tools like `MoveIt <https://github.com/moveit/moveit2>`_.
 
 The key idea is the use of the `Pinocchio <https://github.com/stack-of-tasks/pinocchio>`_ library to represent robots.
 Pinocchio's strength is loading robot models from standard formats such as URDF and MJCF, and being able to represent rigid-body kinematics and dynamics.
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -19,7 +19,7 @@ Vanilla CMake
 
 One of the design points of this library is that it should be portable, and therefore compiles with "vanilla" CMake.
 
-__We do not recommend using this workflow;__ it's more of an exercise in making sure the software can compile without any dependencies that lock it into a particular ecosystem.
+**We do not recommend using this workflow;** it's more of an exercise in making sure the software can compile without any dependencies that lock it into a particular ecosystem.
 
 If you do want to use regular CMake, you should take a look at the Dockerfile under ``.docker/ubuntu``.
 Alternatively, you can try it for yourself.


roboplan/docs/Doxyfile
eholum, line 1: Do you want to move this to `docs/` and adjust the inputs so that we can pull all the docs from a single Doxyfile? Otherwise I think each package will need its own. Something I need to play with though...

sea-bass: I kinda wanted each package to have its own doxyfile?

bindings/src/modules/core.cpp
```diff
@@ -22,15 +22,17 @@ void init_core_types(nanobind::module_& m) {
 
   nanobind::class_<JointConfiguration>(m, "JointConfiguration")
       .def(nanobind::init<>())  // Default constructor
-      .def(nanobind::init<const std::vector<std::string>&, const Eigen::VectorXd&>())
+      .def(nanobind::init<const std::vector<std::string>&, const Eigen::VectorXd&>(),
+           "joint_names"_a, "positions"_a)
```
sea-bass, line 26: This is just an example of arg definitions that need to be done more broadly so the generated stubs don't just have `arg0`, `arg1`, etc. as their arg names.

I didn't go through everything here, but was more playing around with the code

.docker/ubuntu/Dockerfile
```diff
@@ -20,7 +20,8 @@ RUN pip3 install \
     pin==${PINOCCHIO_VERSION} \
     opencv-contrib-python-headless \
     tyro \
-    "viser>=1.0.1"
+    "viser>=1.0.1" \
+    xacro
```
sea-bass, line 24: Found this was missing when I tried to run an example in this docker container.

Btw found some issues with the example models mesh paths in running the examples, will debug later. Probably subtle since the unit tests still work.

bindings/CMakeLists.txt
```diff
@@ -79,6 +79,17 @@ nanobind_add_module(
   ${ROBOPLAN_EXT_SOURCES}
 )
 
+# Generate Python stub files from the bindings.
```
eholum, line 82: Interesting... Do the generated stubs actually need to be committed or is it something that should just go into the install directory?

docs/source/design.rst
```diff
@@ -0,0 +1,40 @@
+Design Philosophy
+=================
+
+RoboPlan is a motion planning framework, similarly to tools like `MoveIt <https://github.com/moveit/moveit2>`_.
```
eholum:
```suggestion
RoboPlan is a motion planning framework, similar to tools like `MoveIt <https://github.com/moveit/moveit2>`_.
```

docs/source/getting_started.rst
```diff
@@ -0,0 +1,161 @@
...
+    docker compose run ubuntu bash
+
+Inside the shell, you can try different commands, such as.
+
+::
+
+    ./build/roboplan_examples/example_scene
+    python3 bindings/examples/example_ik.py
+
+To run the unit tests, you can do:
```
eholum, line 45: Nit but periods or colons? Are the commands part of the sentence?


```suggestion
To run the unit tests, you can do.
```

docs/source/getting_started.rst
```diff
@@ -0,0 +1,161 @@
...
+There are currently 3 supported ways to work with RoboPlan.
+
+---
+
+Vanilla CMake
+-------------
+
+One of the design points of this library is that it should be portable, and therefore compiles with "vanilla" CMake.
+
+__We do not recommend using this workflow;__ it's more of an exercise in making sure the software can compile without any dependencies that lock it into a particular ecosystem.
```
eholum: Formatting here is off, do you want it bold?

```suggestion
**We do not recommend using this workflow;** it's more of an exercise in making sure the software can compile without any dependencies that lock it into a particular ecosystem.
```

Image

eholum (approved):
It's beautiful! Great baseline to start from. I didn't test all the procedures but read through things and have no serious blockers... So stamping this and merge/update as you see fit.

bindings/CMakeLists.txt
sea-bass: They have to be committed purely because the docs look for the .pyi files.

I guess you also wouldn't need this if the docs CI job actually went through and built all of roboplan, which... it doesn't. So instead it snags this stuff from the source tree.

Not needed for the code/IDE integration side of things, though I haven't quite gotten that to work either lol

docs/source/getting_started.rst
sea-bass: Nice, good catch. Markdown to rst brain fog here

docs/source/getting_started.rst
sea-bass: Out-nitting you, but I guess with a period it would have to be a full sentence like "Run the following command."

PR #94, opened 2025-11-09
More consts in the Scene member accessors and python tests

### LINKED ISSUE #92: Make Python bindings tests run with `colcon test`
As of #76, the Python bindings build with colcon, but the tests don't run with `colcon test`.

eholum:
The model getter is the big one that I'm using in places in the bindings, but I think all of these should probably return const references? Copying can either be done by the consumer or we could add non-const access if we decide we need it.

EDIT: While I'm at it resolves #92, but let me know if you want separate PRs.


sea-bass (approved):
Two birds, one scone!

PR #95, opened 2025-11-11
Organize examples

### LINKED ISSUE #91: Put all examples (C++ and Python) in the `roboplan_examples` folder/package
What it says on the tin, it will help with discoverability.

Also involves updating docs to point to how to run examples.

sea-bass:
This PR does 2 main things:

1. Moves the Python bindings examples in the `roboplan_examples` package for better organization
2. Adds a C++ RRT example which acts equivalently to the Python one that already existed

Everything else is small fixes as I found them.

Closes #91

### DIFF
--- a/roboplan_example_models/src/resources.cpp
+++ b/roboplan_example_models/src/resources.cpp
@@ -1,5 +1,5 @@
 #include <dlfcn.h>
-#include <iostream>
+
 #include <roboplan_example_models/resources.hpp>
 
 namespace roboplan::example_models {
--- a/roboplan_examples/python/example_rrt.py
+++ b/roboplan_examples/python/example_rrt.py
@@ -7,6 +7,7 @@ import xacro
 
 import matplotlib.pyplot as plt
 import pinocchio as pin
+
 from common import MODELS
 from roboplan.core import JointConfiguration, PathShortcutter, Scene
 from roboplan.example_models import get_package_share_dir


docs/source/getting_started.rst
```diff
@@ -40,7 +40,7 @@ Inside the shell, you can try different commands, such as.
 ::
 
     ./build/roboplan_examples/example_scene
```
sjahr: I think this needs to be updated as well here and in one other location
```suggestion
    ./build/roboplan_examples/cpp/example_scene
```

sjahr (commented):
In general this looks great! One nitpick comment.
I run into the issue that I wasn't able to run the cpp example because of this error:
```
./build/roboplan_examples/cpp/example_scene
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Mesh package://roboplan_example_models/models/ur_robot_model/meshes/collision/base.stl could not be found.
  ```
  
  When parsing the URDF it looks like the mesh path is not properly found in the pixi environment. Am I missing anything other than following the getting started steps to set everything up?

File: roboplan_example_models/src/resources.cpp [hunk 2]
```diff
@@ -1,5 +1,5 @@
 #include <dlfcn.h>
-
+#include <iostream>
```
sea-bass: Oops

```suggestion

```

roboplan_example_models/src/resources.cpp
```diff
@@ -1,5 +1,4 @@
 #include <dlfcn.h>
```
sea-bass, line 1:
```suggestion
#include <dlfcn.h>

```

roboplan_examples/python/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+install(
+  PROGRAMS
+  common.py
```
eholum, line 3: Bad news bears... Have you tested this with colcon?

```
  File "/home/parallels/ROS/seabass/roboplan-ros/install_jazzy/roboplan_examples/lib/roboplan_examples/example_ik.py", line 11, in <module>
    from common import MODELS, ROBOPLAN_EXAMPLES_DIR
ImportError: cannot import name 'ROBOPLAN_EXAMPLES_DIR' from 'common' (/home/parallels/ROS/seabass/roboplan-ros/roboplan/roboplan_examples/python/common.py). Did you mean: 'ROBOPLAN_MODELS_DIR'?
[ros2run]: Process exited with failure 1
```

I think that `common.py` isn't actually importable if it's installed in the manner... this might have to actually be a python package... Does it work with pixi?

sea-bass: It does work, I just removed that constant and forgot to update it here... one sec

sea-bass: ... try again now? (and thank you)

PR #97, opened 2025-11-11
Upgrade to Pinocchio 3.9

sea-bass:
Now that Pinocchio 3.8 is out in all available forms, we can upgrade and use the upstream Viser visualizer.

EDIT: 3.8 had bugs with visualizing DAE files, so pushing to 3.9 which fixes them.

- [x] available on `conda-forge` (for Pixi workflow)
- [x] available on PyPi
- [x] available on ROS 2
  - [x] Humble
  - [x] Jazzy
  - [x] Kilted
  - [x] Rolling


eholum (commented):
Yay! But you have to update the imports in the examples since they all still point to the roboplan viser visualizer...

sea-bass:
Yeah it shouldn't be that far off. Will stash this PR for now.

sea-bass:
Just 1 jazzy sync away from this PR now...

Issue #96, opened 2025-11-11
Make a reusable function for visualizing joint trajectories


sea-bass:
Right now, there is a TODO in the `example_rrt.py` file saying to do this. So we should do this.

PR #98, opened 2025-11-15
Extract visualize_joint_trajectory in from example_rrt.py

sjahr:
Address #96

### DIFF
--- a/bindings/src/roboplan/visualization.py
+++ b/bindings/src/roboplan/visualization.py
@@ -1,3 +1,4 @@
+from matplotlib.figure import Figure
 import matplotlib.pyplot as plt
 import numpy as np
 
@@ -124,8 +125,8 @@ def visualizeTree(
 
 
 def visualizeJointTrajectory(
-    trajectory: JointTrajectory, scene: Scene, plot_title: str = "Trajectory"
-):
+    trajectory: JointTrajectory, scene: Scene, plot_title: str = "Joint Trajectory"
+) -> Figure:
     """
     Visualize a joint trajectory as a plot of joint positions over time.
 
@@ -145,11 +146,11 @@ def visualizeJointTrajectory(
 
     dof_names = []
     for name in trajectory.joint_names:
-        for idx in range(scene.getJointInfo(name).num_position_dofs):
-            if idx > 0:
-                dof_names.append(f"{name}:{idx}")
-            else:
-                dof_names.append(name)
+        nq = scene.getJointInfo(name).num_position_dofs
+        if nq == 1:
+            dof_names.append(name)
+        else:
+            dof_names.extend(f"{name}:{idx}" for idx in range(nq))
 
     plt.legend(dof_names)
 


bindings/examples/common.py
```diff
@@ -249,3 +250,31 @@ class RobotModelConfig:
         ],
     ),
 }
+
+
+def visualize_joint_trajectory(trajectory, joint_names: List[str], scene: Scene):
```
sea-bass: I think there is a `bindings/src/roboplan/visualization.py` where this would work better

bindings/examples/common.py
```diff
@@ -249,3 +250,31 @@ class RobotModelConfig:
...
+        scene: The Scene object used to get joint information.
+
+    Returns:
+        The matplotlib figure object. Use plt.show() to display it.
+    """
+    plt.ion()
+    plt.plot(trajectory.times, trajectory.positions)
+    plt.xlabel("Time")
+    plt.ylabel("Joint positions")
+    plt.title("Time-parameterized trajectory")
```
sea-bass: Maybe this title should be more generic, like "Joint trajectory", with an ability to override the default via arg?

bindings/examples/common.py
sea-bass: Now that I recall... I think the `trajectory` input may actually already have a `.joint_names` field? Might be a chance to clean up my own inefficiencies :)

sea-bass (commented):
Thank you for tackling this! A few suggestions and should be good

roboplan_examples/python/example_rrt.py
sea-bass, line 1: looks like the file was changed to be executable, which tripped up the linter as this then needs a shebang. Was this needed? (I think yes, and in my standing PR this is changed I believe)

sea-bass:
Last comment: please make sure this includes the incidental fix I found in https://github.com/open-planning/roboplan/pull/95/commits/00d90644de5b1de4a05bd893a4cc33e8f6c7d7a6... unless https://github.com/open-planning/roboplan/pull/98#discussion_r2530607247 renders this comment useless

(You should be able to verify this by the finger joints for the Franka no longer being part of the legend)

bindings/src/roboplan/visualization.py
```diff
@@ -120,3 +121,30 @@ def visualizeTree(
             colors=goal_tree_color,
             line_width=1.0,
         )
+
+
+def visualizeJointTrajectory(trajectory, scene: Scene, plot_titel: str = "Trajectory"):
```
sea-bass: there is no type hint for `trajectory`, and also `plot_titel` -> `plot_title`

bindings/src/roboplan/visualization.py
```diff
@@ -120,3 +121,30 @@ def visualizeTree(
...
+    """
+    Visualize a joint trajectory as a plot of joint positions over time.
+
+    Args
+        trajectory: The trajectory object to be visualized.
+        scene: The Scene object used to get joint information.
```
sea-bass, line 135: plot_title needs to be documented here too

bindings/src/roboplan/visualization.py
sea-bass: technically there should also be a return type annotation

bindings/src/roboplan/visualization.py
```diff
@@ -120,3 +121,30 @@ def visualizeTree(
...
+    plt.ion()
+    plt.plot(trajectory.times, trajectory.positions)
+    plt.xlabel("Time")
+    plt.ylabel("Joint positions")
+    plt.title(plot_titel)
+
+    dof_names = []
+    for name in trajectory.joint_names:
+        for idx in range(scene.getJointInfo(name).num_position_dofs):
+            dof_names.append(f"{name}:{idx}")
```
sea-bass: this is up to you, but I was thinking maybe if the joint just has 1 position DOF, we can completely skip adding the `:0` suffix? Might make things look nicer.

I just didn't do it because all the extra logic didn't look good inline with the example, but it seems reasonable now that you've cleaned it up

bindings/src/roboplan/visualization.py
sjahr: Whoops, titel is actually the German word for title :sweat_smile:

sea-bass (approved):
thanks @sjahr!

PR #102, opened 2025-12-13
Add caching and timeouts for CI

eholum:
Added it in the ros module so adding here too. We only get 10 GB of free gha storage but I think it'll help with rebuilds of PRs.

### DIFF
--- a/docs/python_docs_requirements.txt
+++ b/docs/python_docs_requirements.txt
@@ -1,5 +1,5 @@
 breathe
-sphinx
+sphinx<9.0.0
 sphinx-autoapi
 sphinx-autodoc-typehints
 sphinx-copybutton


docs/python_docs_requirements.txt
```diff
@@ -1,5 +1,5 @@
 breathe
-sphinx
+sphinx<9.0.0
```
sea-bass, line 2: stuck a temporary pin here due to https://github.com/readthedocs/sphinx_rtd_theme/issues/1670

eholum: Cheers dawg I was panicking.

sea-bass: This does uncover that we are not running CI on a regular schedule beyond "pushed to main". I should add that too

EDIT: wait, CI is... just not the docs build

eholum: Is that an RTD config issue?

sea-bass: apparently scheduled builds are a thing, but not configurable in the "community" version of RTD 🚀

eholum: Clearly they don't appreciate the juggernaut that is open planning.

PR #104, opened 2025-12-26
Fix: Prevent Docker volume mounting from clearing build output (#99)

### LINKED ISSUE #99: In the ubuntu Docker container, the build output is cleared away on container startup
This has to do because of the volume mounting setup, so basically you need to enter the container and rebuild to run the examples.

This could be addressed most easily by putting all the source code in a `src/` subfolder and only mounting that volume, rather than... well.. the whole repo.

Could also be done by properly separating the layers for dev containers vs. fully built CI containers, for example.

weijieyong:
Fixes #99

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -35,6 +35,9 @@ services:
     volumes:
       # Mount the source code.
       - ./:/roboplan_ws/:rw
+      # Preserve build artifacts from the image with anonymous volumes.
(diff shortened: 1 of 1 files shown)


sea-bass (approved):
LGTM, thank you for this fix!

CI fails because you have slashes in your branch names, which will be fixed in https://github.com/open-planning/roboplan/pull/105

PR #106, opened 2025-12-30
Add SO-101 arm model

### LINKED ISSUE #54: Add SO-100/SO-101 arm models
For the LeRobot stans.

sea-bass:
what it says on the tin.

Seems like the `<material>` tags on the STL geometries don't make it into Viser, but meh

image

image

Closes #54


sea-bass:
> I wonder if it's worth using pymeshlab to simplify the meshes a bit to reduce the size?

This would be good to do for the Franka/Kinova models too. Franka folder especially is like 70 MB :|

Issue #100, opened 2025-11-21
Nanobind interfaces change to symlinks on pixi build


eholum:
Is this just me? After compiling the latest with `pixi run build_all` the bindings are changed to symlinks instead of files,

```
On branch pinocchio-3.8
Your branch is up to date with 'origin/pinocchio-3.8'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        typechange: bindings/src/roboplan/roboplan_ext/__init__.pyi
        typechange: bindings/src/roboplan/roboplan_ext/core.pyi
        typechange: bindings/src/roboplan/roboplan_ext/example_models.pyi
        typechange: bindings/src/roboplan/roboplan_ext/rrt.pyi
        typechange: bindings/src/roboplan/roboplan_ext/simple_ik.pyi
        typechange: bindings/src/roboplan/roboplan_ext/toppra.pyi
```

```
$  ls -la bindings/src/roboplan/roboplan_ext/*.pyi
lrwxrwxrwx 1 parallels parallels 62 Nov 21 15:28 bindings/src/roboplan/roboplan_ext/core.pyi -> ../../../build/cp311-cp311-linux_aarch64/roboplan_ext/core.pyi
lrwxrwxrwx 1 parallels parallels 72 Nov 21 15:28 bindings/src/roboplan/roboplan_ext/example_models.pyi -> ../../../build/cp311-cp311-linux_aarch64/roboplan_ext/example_models.pyi
lrwxrwxrwx 1 parallels parallels 66 Nov 21 15:28 bindings/src/roboplan/roboplan_ext/__init__.pyi -> ../../../build/cp311-cp311-linux_aarch64/roboplan_ext/__init__.pyi
lrwxrwxrwx 1 parallels parallels 61 Nov 21 15:28 bindings/src/roboplan/roboplan_ext/rrt.pyi -> ../../../build/cp311-cp311-linux_aarch64/roboplan_ext/rrt.pyi
lrwxrwxrwx 1 parallels parallels 67 Nov 21 15:28 bindings/src/roboplan/roboplan_ext/simple_ik.pyi -> ../../../build/cp311-cp311-linux_aarch64/roboplan_ext/simple_ik.pyi
[...]

PR #110, opened 2026-01-04
Optimal differential IK solver

sjahr:
## TODO
- [X] Skeleton implementation
- [X] Task support
- [X] Python bindings + simple example
- [x] Constraint support
- [X] ~~Barrier support~~ --> Follow-up
- [x] Tests

## Run example
```
python3 roboplan_examples/python/example_oink.py --model [franka, ur5. dual ...]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ The main folders found in this repo are as follows.
 
 - `roboplan` : The core C++ library.
 - `roboplan_simple_ik` : A simple inverse kinematics (IK) solver.
+- `roboplan_oink` : A task-based Optimal Inverse Kinematics (OInK) solver.
 - `roboplan_rrt` : A Rapidly-exploring Random Tree (RRT) based motion planner.
 - `roboplan_toppra` : A wrapper around the TOPP-RA algorithm for trajectory timing.
 - `roboplan_example_models` : Contains robot models used for testing and examples.


bindings/src/roboplan/roboplan_ext/optimal_ik.pyi
```diff
@@ -0,0 +1 @@
+../../../build/cp311-cp311-linux_x86_64/roboplan_ext/optimal_ik.pyi
```
sea-bass: this is annoying but the `pixi` setup creates symlinks whereas committing requires copies. see https://github.com/open-planning/roboplan/issues/100#issuecomment-3651741561

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -59,6 +59,10 @@ class Scene {
   /// @return The Pinocchio model.
   const pinocchio::Model& getModel() const { return model_; };
 
+  /// @brief Gets the scene's internal Pinocchio data.
+  /// @return The Pinocchio data.
+  pinocchio::Data& getData() const { return model_data_; };
```
sea-bass: can/should this return a `const`?

roboplan_examples/python/example_oink.py
```diff
@@ -0,0 +1,229 @@
...
+            if current_tasks:
+                # Solve IK for one step with constraints
+                success, delta_q = oink.solveIk(current_tasks, constraints, scene)
+
+                if not success:
+                    print("Warning: IK solver failed, using zero delta_q")
+                    delta_q = np.zeros(num_variables)
+
+                # Integrate: update current config by adding delta_q
+                q_current = q_current + delta_q
```
sea-bass: should be we be using `pin.integrate` here just to be safe... and should this be a method of the OINK solver itself for convenience?

roboplan_oink/include/roboplan_oink/tasks/frame.hpp
```diff
@@ -0,0 +1,51 @@
...
+struct FrameTask : public Task {
+  std::string frame_name;
+  CartesianConfiguration target_pose;
+  double position_cost = 1.0;
+  double orientation_cost = 1.0;
+
+  // Constructor
+  FrameTask(const std::string& name, const CartesianConfiguration& target_pose,
+            double task_weight = 1.0, double position_cost = 1.0, double orientation_cost = 1.0,
+            double task_gain = 1.0, double lm_damp = 0.0)
```
sea-bass: should we consider a struct to capture all these parameters?

roboplan_oink/src/constraints/position_limit.cpp
```diff
@@ -0,0 +1,54 @@
+#include <roboplan_oink/constraints/position_limit.hpp>
+
+namespace roboplan {
+
+bool PositionLimit::computeQpConstraints(const std::shared_ptr<Scene>& scene,
+                                         Eigen::MatrixXd& constraint_matrix,
+                                         Eigen::MatrixXd& lower_bounds,
+                                         Eigen::MatrixXd& upper_bounds) {
+  const auto& model = scene->getModel();
```
sea-bass: is there any risk in getting the model at every iteration this way?

roboplan_oink/src/constraints/position_limit.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  for (int i = 0; i < model.nv; ++i) {
+    if (std::isfinite(q_max[i])) {
+      delta_q_max[i] = q_max[i] - q[i];
+    } else {
+      delta_q_max[i] = std::numeric_limits<double>::infinity();
+    }
+
+    if (std::isfinite(q_min[i])) {
+      delta_q_min[i] = q[i] - q_min[i];
```
sea-bass: use `.at()` on the right hand side of all these indexing operations

roboplan_oink/src/constraints/position_limit.cpp
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  // Scale by gain parameter
+  Eigen::VectorXd p_max = config_limit_gain * delta_q_max;
+  Eigen::VectorXd p_min = config_limit_gain * delta_q_min;
+
+  // Construct constraint matrix: identity matrix
+  constraint_matrix.resize(model.nv, model.nv);
+  constraint_matrix = Eigen::MatrixXd::Identity(model.nv, model.nv);
```
sea-bass: do you need the resize if you're just setting its value in the following line?

roboplan_oink/src/constraints/position_limit.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  // For box constraints l <= G*dq <= u where G = I
+  lower_bounds.resize(model.nv, 1);
+  upper_bounds.resize(model.nv, 1);
```
sea-bass: likewise here like with the `getModel()` call, is it okay to set this every iteration or can this be pre-computed somehow?

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+Oink::Oink(int num_variables, int num_constraints)
+    : task_J(Eigen::MatrixXd::Zero(6, num_variables)), task_e(Eigen::VectorXd::Zero(6)),
+      task_c(Eigen::VectorXd::Zero(num_variables)), task_H(num_variables, num_variables),
+      H(num_variables, num_variables), c(Eigen::VectorXd::Zero(num_variables)) {
+
+  // Initialize sparse matrices with small regularization to ensure positive definiteness
+  task_H.setZero();
+  H.setIdentity();  // Start with identity instead of zero for numerical stability
+  H *= 1e-6;        // Small regularization term
```
sea-bass: use a constant instead of magic number

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -0,0 +1,159 @@
...
+    throw std::runtime_error("Solver initialization failed");
+  }
+}
+
+bool Oink::solveIk(const std::vector<std::shared_ptr<Task>>& tasks,
+                   const std::vector<std::shared_ptr<Constraints>>& constraints,
+                   const std::shared_ptr<Scene>& scene, Eigen::VectorXd& delta_q) {
+  H.setZero();
+  c.setZero();
+  for (auto const& task : tasks) {
```
sea-bass: I guess our linter doesn't care, but we use `const auto&` here

roboplan/include/roboplan/core/scene.hpp
sjahr: On second look, it doesn't work with [computeFrameJacobian](https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/namespacepinocchio.html#a7b1189aea93a252f4c06d9a7e3c9f5e7).
I need to take a look why the data needs to be mutable for this function

roboplan_oink/src/constraints/position_limit.cpp
sjahr: I updated the memory management a bit in general

roboplan_oink/src/constraints/position_limit.cpp
sjahr: Good question :sweat_smile: At the moment, the constraint instantiations doesn't contain any state which I like from a cleanness perspective, however it could be a small drawback when it comes to performance. For now, I'd propose to keep it like that and do a performance analysis with a flame graph later on.

roboplan_oink/src/constraints/position_limit.cpp
sjahr: In the solveIk function, we check every iteration if the number of constraints has changed and only then we do a resize to avoid unnecessary memory allocation in the critical path.

roboplan/include/roboplan/core/scene.hpp
sea-bass: Could also consider not adding this function, but rather creating a new `pinocchio::Data` inside the solver. This is probably better for thread safety in the long term.

.docker/ubuntu/Dockerfile
```diff
@@ -29,6 +29,24 @@ RUN PIP_PATH="$(pip3 show pin | grep "Location:" | cut -d' ' -f2)/cmeel.prefix"
     echo "export LD_LIBRARY_PATH=${PIP_PATH}/lib:\${LD_LIBRARY_PATH}" >> /roboplan_env.sh \
     echo "export CMAKE_PREFIX_PATH=${PIP_PATH}:\${CMAKE_PREFIX_PATH}" >> /roboplan_env.sh
 
+# Install OSQP from source
+RUN rm -rf roboplan_oink/external/osqp && \
```
sjahr: This turned out to be the most straight forward way to get these two libraries installed in the Docker. For plain cmake, the way it is installed for ROS did not work for me because, the libraries need to be installed system wide for proper detection. Let me know what you think.

roboplan_oink/CMakeLists.txt
```diff
@@ -0,0 +1,143 @@
...
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# Find dependencies.
+find_package(roboplan REQUIRED)
+
+# Try to find OsqpEigen
+find_package(OsqpEigen QUIET)
+
+if(NOT OsqpEigen_FOUND)
```
sjahr, line 17: This entire block only exists for the colon workflow. osqp-eigen is not available as an upstream ROS 2 package at all and osqp-vendor uses an outdated 6.X version of osqp that is not compatible with the latest stabe 1.X version I'd like to use.

.docker/ubuntu/Dockerfile
```diff
@@ -10,7 +10,7 @@ SHELL ["/bin/bash", "-c"]
 RUN apt update -y \
     && apt upgrade -y \
     && apt install -y \
-        cmake libeigen3-dev libgmock-dev liburdfdom-dev libyaml-cpp-dev python3-pip
+        cmake git libeigen3-dev libgmock-dev liburdfdom-dev libyaml-cpp-dev python3-pip
```
sjahr, line 13: Required by osqp cmake code

.docker/ubuntu/Dockerfile
sea-bass: Why don't the submodules work in this container?

roboplan/include/roboplan/core/scene.hpp
sea-bass: bumping this comment

roboplan_oink/CMakeLists.txt
```diff
@@ -0,0 +1,143 @@
+cmake_minimum_required(VERSION 3.22)
+project(roboplan_oink VERSION 0.0.0)
```
sea-bass: Should go 0.1.0 now (same in package.xml)

roboplan_oink/CMakeLists.txt
sea-bass: Again, shouldn't this stuff mean you can build the ubuntu container with just the submodules instead of having to delete and then re-clone the external folder?

sea-bass (commented):
Gave it a test and the example works with the UR, but not with the Franka, Kinova, or SO-101 models. Those seem to error with a Jacobian size mismatch, maybe due to gripper joints or something?

Also, strange things happen when you hit the "Randomize Pose" button. I think instead of continuing to solve from its previous pose and get stuck in a local minimum, we should just temporarily stop the solving timer and teleport the arm to the new pose. Try it for yourself?

---

Trying the ROS 2 / colcon approach, I was worried that osqp/osqp-eigen would take a while to build. It wasn't too bad (took 15 sec on my machine to build the whole package).

However, I initially found a problem because you have to remove `ros-${ROS_DISTRO}-osqp-vendor` first or you get ABI incompatibility errors. This is not ideal, due to `apt` when i remove this package it also took down a whole bunch of other MoveIt ones so this library won't play nice with that. I also don't think we should downgrade our stuff, so maybe there is a path to upgrade the OSQP vendor package instead?

roboplan_oink/CMakeLists.txt
```diff
@@ -0,0 +1,143 @@
...
+    # (try_compile doesn't work with ALIAS targets from add_subdirectory)
+    set(OSQP_IS_V1 TRUE CACHE BOOL "OSQP is version 1.0+" FORCE)
+    set(OSQP_IS_V1_FINAL TRUE CACHE BOOL "OSQP is version 1.0.0 final" FORCE)
+  endif()
+
+  set(OSQPEIGEN_COMPILE_tests OFF CACHE INTERNAL "Disable OsqpEigen tests")
+
+  # Add the external osqp-eigen project
+  # Note: Not using EXCLUDE_FROM_ALL so it can be installed for downstream packages
+  add_subdirectory(external/osqp-eigen)
```
sea-bass: the ubuntu build is failing because of this, I think doing it with `add_subdirectory` means that the package itself is not visible when you try to find it later. You may want to look into `FetchContent` instead, perhaps? https://cmake.org/cmake/help/latest/module/FetchContent.html

.docker/ubuntu/Dockerfile
sjahr: It doesn't work because of the Docker ignore line: https://github.com/open-planning/roboplan/blob/87970d4e47bc7b2a5435373bcc8a40ac6d9dba27/.dockerignore#L1. Ignoring the .git folder leads to the submodule information not being available. I guess the reason to ignore is to keep the container size smaller. Let me change that to verify it works as expected.

sjahr: Outdated, since we wont need that with FetchContent :tada:

roboplan/include/roboplan/core/scene.hpp
sjahr: I looked into this again and I'd like the OInK core struct to be as stateless as possible. Still trying to understand the scene & pinocchio design better but at the moment I see three potential solutions:

1. Data is mutable from outside the scene. Thread-safety needs to be handled by the scene user. I guess this would be a simple solution. The advantage I see here is that you can use pinocchio::computeFunctions like [computeFrameJacobian](https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/namespacepinocchio.html#a7b1189aea93a252f4c06d9a7e3c9f5e7) or  computeDistances() directly with the scene without the need to dublicate the Data structure
2. Data is immutable and scene exposes these functions through it's own members e.g. 
```
scene->computeDistances(...){
return pinocchio::computeDistances(... data_, ...);
}
```
This would be clean but we'd need to essentially duplicate the entire pinocchio API. It looks like this approach is already partially implemented. 
3. Maintain copies of data outside of the scene and sync them back if necessary. Do we even need data as a member of the scene in this case :thinking:

What are your thoughts here? I still might miss something in the bigger picture :wink:

sea-bass: I'd like 3. long term; however the fact that Pinocchio doesn't yet use nanobind would make it tricky to pass Pinocchio data structs through the Python bindings interface at the moment.

So I think for now it's easiest to live with 1. or 2., though it might be nice for the scene to have the potential for multiple `data_` instances stored; maybe in a map or something?

sjahr: Thanks for your quick response!
Yes, claude proposed a multiple data_ instances approach too and it makes sense to me. For this PR I simply added a computeFrameJacobian function to the scene to avoid too heavy changes to the scene class and to be more consistent with the existent API. getData is now const. Let's push that to future work :+1:

bindings/src/modules/core.cpp
```diff
@@ -182,8 +182,31 @@ void init_core_scene(nanobind::module_& m) {
...
+                                       "' not found: " + maybe_frame_id.error());
+            }
+            const auto reference_frame =
+                local ? pinocchio::ReferenceFrame::LOCAL : pinocchio::ReferenceFrame::WORLD;
+            Eigen::MatrixXd jacobian = Eigen::MatrixXd::Zero(6, self.getModel().nv);
+            self.computeFrameJacobian(q, maybe_frame_id.value(), reference_frame, jacobian);
+            return jacobian;
+          },
+          "Computes the frame Jacobian for a specific frame.", "q"_a, "frame_name"_a,
+          "local"_a = true)
```
sea-bass, line 209: why couldn't this just bind with the regular code, because there needs to be a matrix allocated before it's passed by reference?

bindings/src/modules/optimal_ik.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+  // Bind the abstract Task base class with shared_ptr holder
+  nanobind::class_<Task>(m, "Task")
+      .def_ro("gain", &Task::gain)
+      .def_ro("weight", &Task::weight)
+      .def_ro("lm_damping", &Task::lm_damping)
+      .def_ro("num_variables", &Task::num_variables)
+      .def_rw("jacobian_container", &Task::jacobian_container)
+      .def_rw("error_container", &Task::error_container)
+      .def_rw("H_dense", &Task::H_dense);
```
sea-bass: why are some read-only and some read-write?

bindings/src/modules/optimal_ik.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+  // Bind FrameTaskParams configuration struct
+  nanobind::class_<FrameTaskParams>(m, "FrameTaskParams")
+      .def(nanobind::init<>())
+      .def(
+          "__init__",
+          [](FrameTaskParams* self, double position_cost, double orientation_cost, double task_gain,
+             double lm_damping) {
+            new (self) FrameTaskParams{position_cost, orientation_cost, task_gain, lm_damping};
+          },
```
sea-bass, line 36: why is a lambda needed here?

bindings/src/modules/optimal_ik.cpp
```diff
@@ -0,0 +1,149 @@
...
+          "scene"_a)
+      .def(
+          "computeJacobian",
+          [](FrameTask& self, const Scene& scene) {
+            auto result = self.computeJacobian(scene);
+            if (!result.has_value()) {
+              throw std::runtime_error("computeJacobian failed: " + result.error());
+            }
+          },
```
sea-bass: no need to do this, @eholum already created an `unwrap_expected()` utility that will help avoid this boilerplate

bindings/src/modules/optimal_ik.cpp
```diff
@@ -0,0 +1,149 @@
...
+          "scene"_a);
+
+  // Bind ConfigurationTaskParams configuration struct
+  nanobind::class_<ConfigurationTaskParams>(m, "ConfigurationTaskParams")
+      .def(nanobind::init<>())
+      .def(
+          "__init__",
+          [](ConfigurationTaskParams* self, double task_gain, double lm_damping) {
+            new (self) ConfigurationTaskParams{task_gain, lm_damping};
+          },
```
sea-bass, line 61: again, why is the lambda needed here?

bindings/src/modules/optimal_ik.cpp
```diff
@@ -0,0 +1,149 @@
...
+          [](Oink& self, const std::vector<std::shared_ptr<Task>>& tasks,
+             const std::vector<std::shared_ptr<Constraints>>& constraints,
+             const std::shared_ptr<Scene>& scene) -> Eigen::VectorXd {
+            Eigen::VectorXd delta_q;
+            auto result = self.solveIk(tasks, constraints, *scene, delta_q);
+            if (!result.has_value()) {
+              throw std::runtime_error("IK solve failed: " + result.error());
+            }
+            return delta_q;
+          },
```
sea-bass, line 108: and here

bindings/src/roboplan/roboplan_ext/optimal_ik.pyi
sea-bass, line 1: I know it's tedious, but you can choose to modify the bindings code so the `.pyi` files have more docstrings and arg names, as done in https://github.com/open-planning/roboplan/pull/114

roboplan_examples/python/example_oink.py
```diff
@@ -0,0 +1,317 @@
...
+    FrameTask,
+    FrameTaskParams,
+    Oink,
+    PositionLimit,
+)
+from roboplan.viser_visualizer import ViserVisualizer
+
+# Canonical "home" or "ready" poses for each robot model
+# Maps model name -> joint configuration vector
+CANONICAL_POSES = {
```
sea-bass: consider setting these in `roboplan_examples/python/common.py` -- maybe `starting_joint_config` which is already defined takes care of this for you already? But otherwise feel free to add them there

roboplan_examples/python/example_oink.py
```diff
@@ -0,0 +1,317 @@
...
+                for goal in goals:
+                    scene.forwardKinematics(q_current, goal.tip_frame)
+
+                viz.display(q_current)
+
+            # Maintain 100Hz loop rate
+            elapsed = time.time() - loop_start
+            sleep_time = dt - elapsed
+            if sleep_time > 0:
+                time.sleep(sleep_time)
```
sea-bass: you can just do `time.sleep(max(0, dt - elapsed)` ⛳

sea-bass (commented):
Looks super close!

I did run into a few issues with the Python examples, though:

1. When I hit "randomize pose", the robots still freak out a bit instead of not moving until you move the interactive marker again. Any way you can somehow reset the state of the solver when that is pressed?
2. Also, any way to tune these loops to be a little more responsive? Right now there is a lot of lag in chasing the marker targets for all models... it's not entirely clear to me if the `* dt` in the call to `integrate()` is needed.
3. The `--model kinova` option may be misconfigured; the robot visualization starts  in a broken state. Please check this. Likely has to do with the continuous joints.

Image

```
Exception in thread Thread-2 (control_loop):
  File "/home/sebastian/workspace/roboplan_ws/src/roboplan/.pixi/envs/default/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
    self.run()
  File "/home/sebastian/workspace/roboplan_ws/src/roboplan/.pixi/envs/default/lib/python3.11/threading.py", line 982, in run
    self._target(*self._args, **self._kwargs)
  File "/home/sebastian/workspace/roboplan_ws/src/roboplan/roboplan_examples/python/example_oink.py", line 258, in control_loop
    q_current = scene.integrate(q_current, delta_q * dt)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: wrong argument size: expected 13, got 17
hint: The joint velocity vector is not of the right size
```

bindings/src/modules/core.cpp
sjahr: Yes 👍 Do you know a better way to do this? Looks like they use a similar approach in Pinocchio itself:
https://github.com/stack-of-tasks/pinocchio/blob/devel/bindings/python/algorithm/expose-jacobian.cpp

bindings/src/modules/optimal_ik.cpp
sjahr: Without it, this wasn't possible:

```
    task_params = FrameTaskParams(
        position_cost=2.0,
        orientation_cost=1.0,
        task_gain=task_gain,
        lm_damping=lm_damping,
    )
```
alternatively I could do this:

```
task_params = FrameTaskParams
task_params.position_cost=2.0,
task_params.orientation_cost=1.0,
task_params.task_gain=task_gain,
task_params.lm_damping=lm_damping,
```
But I like the first syntax more. Am I missing something here?

bindings/src/modules/optimal_ik.cpp
sjahr: Same as above

sjahr:
> Looks super close!
> 
> I did run into a few issues with the Python examples, though:
> 
>     1. When I hit "randomize pose", the robots still freak out a bit instead of not moving until you move the interactive marker again. Any way you can somehow reset the state of the solver when that is pressed?
> 
>     2. Also, any way to tune these loops to be a little more responsive? Right now there is a lot of lag in chasing the marker targets for all models... it's not entirely clear to me if the `* dt` in the call to `integrate()` is needed.
> 
>     3. The `--model kinova` option may be misconfigured; the robot visualization starts  in a broken state. Please check this. Likely has to do with the continuous joints.
> 
> 
> Image
> 
> ```
> Exception in thread Thread-2 (control_loop):
> Traceback (most recent call last):
>   File "/home/sebastian/workspace/roboplan_ws/src/roboplan/.pixi/envs/default/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
>     self.run()
>   File "/home/sebastian/workspace/roboplan_ws/src/roboplan/.pixi/envs/default/lib/python3.11/threading.py", line 982, in run
>     self._target(*self._args, **self._kwargs)
>   File "/home/sebastian/workspace/roboplan_ws/src/roboplan/roboplan_examples/python/example_oink.py", line 258, in control_loop
>     q_current = scene.integrate(q_current, delta_q * dt)
>                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ValueError: wrong argument size: expected 13, got 17
> hint: The joint velocity vector is not of the right size
> ```

Thanks for your review. I addressed the main points and updated the documentation for the bindings.
Regarding your 3 points:
1. Done, I made sure the doesn't start right away when you randomize the pose
2. Also done, instead of the velocity scaling, I apply the velocity limits now and that improves it.
3. I fixed it, however it is still easy to bring the robot in an unstable position. I need to dig a bit more into it, but if you don't mind, I'd say we don't let this block the PR as I think it is a configuration issue (e.g. by adjusting the lm damping for the kinova)

bindings/src/modules/optimal_ik.cpp
sea-bass: I agree, but also suspect there are better ways to do this with nanobind. ChatGPT quickly gave me something like this to try?

```
namespace nb = nanobind;
using namespace nb::literals;

nb::class_<FrameTaskParams>(m, "FrameTaskParams")
    .def(nb::init<
        double, double, double, double
    >(),
        "position_cost"_a = 0.0,
        "orientation_cost"_a = 0.0,
        "task_gain"_a = 0.0,
        "lm_damping"_a = 0.0
    )
    .def_rw("position_cost", &FrameTaskParams::position_cost)
    .def_rw("orientation_cost", &FrameTaskParams::orientation_cost)
    .def_rw("task_gain", &FrameTaskParams::task_gain)
    .def_rw("lm_damping", &FrameTaskParams::lm_damping);
```

bindings/src/modules/core.cpp
sea-bass: again, chatgpt tells me `Eigen::Ref` could be useful here... https://chatgpt.com/s/t_6980a19ea00c8191b0a066f1ba607cc3

sea-bass (approved):
LET'S GO!

(tested and all works as you said)

Issue #57, opened 2025-08-19
Enhancement: Iterative TOPP-RA option


sea-bass:
If you use TOPP-RA with cubic splines instead of cubic Hermite splines, you can get path parameterizations with nonzero velocity at intermediate waypoints. However this can cause overshoots and therefore collisions.

Consider adopting an interative technique that looks for collisions at certain segments and adds intermediate waypoints until a trajectory is valid.

PR #111, opened 2026-01-05
Collision-aware iterative TOPPRA for cubic-spline paths

weijieyong:
implementing #57 

- switch to cubic splines (vs cubic Hermite) to allow nonzero velocities at knots.
- add iterative collision-refinement: adaptive sampling + insert chord point near collision to mitigate overshoot/collisions.
- No formal tests added yet; I validated on existing examples and a constructed collision case.

ran test with 
```bash
pixi run python3 roboplan_examples/python/example_rrt.py --rng-seed 99 --include-obstacles
```

tested with obstacles position, in `roboplan_examples/python/common.py`
```
box: [-0.27, -0.27, 1.0]
sphere: [1.05, -0.2, 0.2]
```

Comparison video of cubic-spline TOPPRA with iterative collision refinement (enabled) vs baseline (disabled). 
Baseline(left) overshoots and collides; refinement(right) inserts a waypoint near the collision and produces a collision-free trajectory

https://github.com/user-attachments/assets/c4481686-5bbd-4df1-ad31-60ba74ae00ea

### DIFF
--- a/roboplan_toppra/src/toppra.cpp
+++ b/roboplan_toppra/src/toppra.cpp
@@ -101,12 +101,9 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
   acc_constraint->discretizationType(toppra::DiscretizationType::Interpolation);
   toppra::LinearConstraintPtrs constraints = {vel_constraint, acc_constraint};
 
-  // Create initial cubic spline with path and random times.
-  toppra::Vectors path_pos_vecs, path_vel_vecs;
+  // Create initial cubic spline with path points.
+  toppra::Vectors path_pos_vecs;
   path_pos_vecs.reserve(num_pts);
-  path_vel_vecs.reserve(num_pts);
-  std::vector<double> steps;
-  steps.reserve(num_pts);
   for (size_t idx = 0; idx < path.positions.size(); ++idx) {
     const auto& pos = path.positions.at(idx);
     auto maybe_collapsed_pos = collapseContinuousJointPositions(*scene_, group_name_, pos);
@@ -139,6 +136,19 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
   const int kMaxIterations = 10;
   bool found_valid_path = false;
 
+  // Collision checking helper.
+  auto check_collision = [&](const Eigen::VectorXd& q_collapsed)
+      -> tl::expected<bool, std::string> {
+    auto maybe_expanded = expandContinuousJointPositions(*scene_, group_name_, q_collapsed);
+    if (!maybe_expanded) {
+      return tl::make_unexpected("Failed to expand joint positions: " +
+                                 maybe_expanded.error());
+    }
+    const Eigen::VectorXd full_q =
+        scene_->toFullJointPositions(group_name_, maybe_expanded.value());
+    return scene_->hasCollisions(full_q);
+  };
+
   for (int iter = 0; iter < kMaxIterations; ++iter) {
     // 1. Re-calculate steps based on current path points (chordal parameterization)
     std::vector<double> steps;
@@ -151,9 +161,14 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
       steps.push_back(s);
     }
 
-    // 2. Build Cubic Spline with Natural BCs
-    std::array<toppra::BoundaryCond, 2> bc_type = {toppra::BoundaryCond("natural"),
-                                                   toppra::BoundaryCond("natural")};
+    // 2. Build Cubic Spline with clamped endpoint derivatives (smooth velocity transition)
+    const Eigen::VectorXd start_deriv =
+        (path_pos_vecs[1] - path_pos_vecs[0]) / (steps[1] - steps[0]);
+    const Eigen::VectorXd end_deriv =
+        (path_pos_vecs[path_pos_vecs.size() - 1] - path_pos_vecs[path_pos_vecs.size() - 2]) /
+        (steps[steps.size() - 1] - steps[steps.size() - 2]);
+    std::array<toppra::BoundaryCond, 2> bc_type = {
+        toppra::BoundaryCond(1, start_deriv), toppra::BoundaryCond(1, end_deriv)};
     Eigen::Map<Eigen::VectorXd> spline_times_vec(steps.data(), steps.size());
     const auto spline =
         toppra::PiecewisePolyPath::CubicSpline(path_pos_vecs, spline_times_vec, bc_type);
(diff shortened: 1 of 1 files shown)


bindings/src/roboplan/roboplan_ext/__init__.pyi
sea-bass, line 1: Don't modify these files as they will break non-Pixi CI

EDIT: this is a known issue explained here: https://github.com/open-planning/roboplan/issues/100#issuecomment-3651741561

roboplan_toppra/src/toppra.cpp
```diff
@@ -131,19 +127,134 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
...
     }
-
     path_pos_vecs.push_back(curr_collapsed);
-    path_vel_vecs.push_back(Eigen::VectorXd::Zero(curr_collapsed.size()));
+  }
+
+  // Iterative Spline Refinement for Collision Avoidance
+  // We loop to generate the spline, check for collisions, and insert waypoints if needed.
+  std::shared_ptr<toppra::PiecewisePolyPath> geom_path;
+  const int kMaxIterations = 10;
```
sea-bass, line 136: this should be a parameter

roboplan_toppra/src/toppra.cpp
```diff
@@ -131,19 +127,134 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
...
+        toppra::BoundaryCond(1, start_deriv), toppra::BoundaryCond(1, end_deriv)};
+    Eigen::Map<Eigen::VectorXd> spline_times_vec(steps.data(), steps.size());
+    const auto spline =
+        toppra::PiecewisePolyPath::CubicSpline(path_pos_vecs, spline_times_vec, bc_type);
+    geom_path = std::make_shared<toppra::PiecewisePolyPath>(spline);
+
+    // 3. Check for collisions
+    bool collision_detected = false;
+    size_t segment_to_split = 0;
+    double split_alpha = 0.5;
```
sea-bass, line 180: what is `split_alpha`? Is this the interpolation factor along the current segment at which to add a new waypoint? Might be worth a comment explaining this as it's a pretty important part of the algorithm

roboplan_toppra/src/toppra.cpp
```diff
@@ -131,19 +127,134 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
...
+
+    // Always verify that all waypoints are collision-free (including inserted ones).
+    for (size_t i = 0; i < path_pos_vecs.size(); ++i) {
+      const auto maybe_collision = check_collision(path_pos_vecs[i]);
```
sea-bass, line 184: use `path_pos_vecs.at(i)` instead

roboplan_toppra/src/toppra.cpp
```diff
@@ -131,19 +127,134 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
...
+    }
+    
+    // Adaptive check resolution in joint space (approx 0.05 rad / 5 cm).
+    const double max_joint_step = 0.05;
+    
+    for (size_t i = 0; i < steps.size() - 1; ++i) {
+      const double t_start = steps[i];
+      const double t_end = steps[i+1];
+      
+      const Eigen::VectorXd delta = path_pos_vecs[i + 1] - path_pos_vecs[i];
```
sea-bass, line 201: same here, use `.at()` if you're not modifying the value

roboplan_toppra/src/toppra.cpp
```diff
@@ -131,19 +127,134 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
...
+          return tl::make_unexpected(maybe_collision.error());
+        }
+        if (maybe_collision.value()) {
+          collision_detected = true;
+          segment_to_split = i;
+          split_alpha = alpha;
+          break;
+        }
+      }
+      if (collision_detected) break;
```
sea-bass, line 221: use braces/newlines anyway, single-line if-statements can be risky

roboplan_toppra/src/toppra.cpp
```diff
@@ -131,19 +127,134 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
...
+    }
+
+    if (!collision_detected) {
+      found_valid_path = true;
+      break;
+    }
```
sea-bass, line 227: I feel like we can simplify this to only use a single variable. If `collision_detected` is pre-initialized to `false` and it's only set to `true` in the loop, then that already tells us whether the path is value.

roboplan_toppra/src/toppra.cpp
```diff
@@ -131,19 +127,134 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
...
+
+    // 4. Resolve Collision: Insert a point on the chord at the detected collision fraction.
+    // This biases the path toward the straight-line segment which is more likely to be collision-free.
+    if (iter < kMaxIterations - 1) {
+        const Eigen::VectorXd chord_point =
+            (1.0 - split_alpha) * path_pos_vecs[segment_to_split] +
+            split_alpha * path_pos_vecs[segment_to_split + 1];
```
sea-bass, line 234: again use `.at()` here

roboplan_toppra/src/toppra.cpp
```diff
@@ -131,19 +127,134 @@ PathParameterizerTOPPRA::generate(const JointPath& path, const double dt,
...
+                                 std::to_string(kMaxIterations) + " iterations.");
   }
-  const auto spline =
-      toppra::PiecewisePolyPath::CubicHermiteSpline(path_pos_vecs, path_vel_vecs, steps);
-  const auto geom_path = std::make_shared<toppra::PiecewisePolyPath>(spline);
 
   // Solve TOPP-RA problem.
   toppra::PathParametrizationAlgorithmPtr algo =
       std::make_shared<toppra::algorithm::TOPPRA>(constraints, geom_path);
+  algo->setN(500);  // increase the number of grid points to prevent sharp acceleration changes
```
sea-bass, line 257: why is this needed?

weijieyong:
Hi @sea-bass,

Thank you for the kind message, and sorry for disappearing on this PR.
I appreciate you picking this up and finishing it, and thanks as well for explaining the difference in approach with proposeGridpoints(). I’m glad the initial implementation was useful as a starting point.

Also, thank you for your work on such an amazing project. I’ve really enjoyed learning from it.

Thanks again, and I hope to contribute again in the future.

PR #112, opened 2026-01-11
CMake and example model path fixes for conda-forge building

sea-bass:
When doing experiments building with `conda-forge`, I realized that the models install locations breaks when you build this way. A small fix seems to do the trick.

EDIT: The `dladdr` hacks in the `get_install_prefix()` function did not work as intended, so these were updated to handle the conda install case as well.

### DIFF
--- a/roboplan_example_models/src/resources.cpp
+++ b/roboplan_example_models/src/resources.cpp
@@ -1,6 +1,5 @@
 #include <cstdlib>
 #include <dlfcn.h>
-#include <iostream>
 
 #include <roboplan_example_models/resources.hpp>
 


roboplan_example_models/src/resources.cpp
```diff
@@ -1,10 +1,19 @@
...
 
 #include <roboplan_example_models/resources.hpp>
 
 namespace roboplan::example_models {
 
 std::filesystem::path get_install_prefix() {
+  // If roboplan was installed using conda, get it from the prefix.
+  // Note that this will break in the event that you've built from source and are
+  // using conda yourself, but it's a better default than not supporting conda at all.
+  const auto conda_prefix = std::getenv("CONDA_PREFIX");
```
eholum: Ah I should probably try this with conda... but is this _really_ necessary? If the `dladdr` doesn't cover conda installs then something might be amiss with it. Can you elaborate on exactly the problem you ran into?

(I think... reminder that I only vaguely understand how any of this works)

sea-bass: The key issue is that the conda workflow installs the bindings to `$CONDA_PREFIX/lib/python3.xx/site-packages` which is where `dladdr` points to...

This doesn't match where the C++ lib installed models, which after this PR is (corrected to) `$CONDA_PREFIX/share/roboplan_example_models`, to match the library location at `$CONDA_PREFIX/lib/libroboplan_example_models.so`

If you can find a better solution, we are in dire need of it lol

sea-bass: Basically, I would have expected the `libroboplan_example_models.so` location to be returned, and not the `lib/python3.xx/site_packages` location... so that would be a way to fix it

eholum: This branch might do it... 

https://github.com/open-planning/roboplan/tree/eho-conda-crapola

sea-bass: Oh cool, that looks promising. I will test this later, and if all is good will cherry pick the one commit to this branch, as conda recipes won't get committed to this repo.

Also why did you need to limit the Python versions to just 3.11 to 3.12? We definitely need to support 3.10 and above.

eholum: > Also why did you need to limit the Python versions to just 3.11 to 3.12?

Ah, I was running into conflict resolution issues (for some reason my environment defaulted to 3.13... And I'm assuming arm differences). I noticed that these are the limits we set in [the pixi file](https://github.com/open-planning/roboplan/blob/eho-conda-crapola/pixi.toml#L211) so I just tossed them in! That restriction might not be necessary, though... Lots of guessing here.

sea-bass: ooh issue, in `pixi` install you now get:

```
>>> get_install_prefix()
PosixPath('/home/sebastian/workspace/roboplan_ws/src/roboplan/build')
```

which breaks the `dladdr` stuff in a different way :(

For some reason, the `pixi` setup also has libs installed to `${CONDA_PREFIX}/lib` and models to `${CONDA_PREFIX}/share`, but it's the build-time ones that are showing up first. Any ideas?

roboplan_example_models/src/resources.cpp
```diff
@@ -1,10 +1,19 @@
+#include <cstdlib>
 #include <dlfcn.h>
 
 #include <roboplan_example_models/resources.hpp>
 
 namespace roboplan::example_models {
 
 std::filesystem::path get_install_prefix() {
+  // If roboplan was installed using conda, get it from the prefix.
+  // Note that this will break in the event that you've built from source and are
```
sjahr: Is there a way to print a warning in this case?

sea-bass: That's a good idea if we have to keep this, but I think we can find a way to avoid it. Just need to figure out what Pixi is doing differently so we can configure the project to pick up the installed libraries instead.

roboplan_example_models/src/resources.cpp
sea-bass: ahhh @eholum the solution was simple here -- use your commit, but just don't run `std::filesystem::canonical()` on it. That makes it all work on my end.

eholum: Ah man what a faff

roboplan_example_models/src/resources.cpp
sea-bass: not needed after all 🤞🏻

eholum (approved):
If you tell me it works I'll believe you!

roboplan_example_models/src/resources.cpp
```diff
@@ -1,15 +1,21 @@
+#include <cstdlib>
 #include <dlfcn.h>
+#include <iostream>
```
sea-bass:
```suggestion
```

sea-bass:
Another update... it seems the issue was that `std::filesystem::canonical()` was useful in the conda case to get rid of relative paths like `/../../` that were in the string, which was needed, but inadvertently followed symlinks.

`std::filesystem::path::lexically_normal()` actually does what we want, blessed be the fruit

PR #113, opened 2026-01-14
Improve error reporting in scene_utils map access

olaghattas:
Description:
Problem: std::map::at failures did not indicate which key was missing, making debugging difficult.

Change: Improved error reporting in scene_utils to expose missing keys clearly. Updated the map in scene_utils.hpp to include the missing key.


sea-bass (dismissed):
Thank you @olaghattas !

olaghattas:
Should be all set.

PR #115, opened 2026-01-18
0.1.0

sea-bass:
This is going to be the inaugural version before I blast it out to conda-forge.


bindings/test/test_import.py
```diff
@@ -11,16 +11,15 @@ def test_import_roboplan() -> None:
...
 
 
 def test_import_pinocchio() -> None:
     assert importlib.util.find_spec("pinocchio")
     importlib.import_module("pinocchio")
-
-
-def test_import_viser() -> None:
-    assert importlib.util.find_spec("viser")
-    importlib.import_module("viser")
```
sea-bass, line 26: this really should not be part of the core module, Viser is optional

sjahr (approved):

PR #117, opened 2026-02-01
[pixi] Rely on conda-forge versions of xacro and libtoppra

sea-bass:
@eholum I would be interested in having you test this on your `osx-64` platform since `libtoppra` is only available on `linux-64` for now. I had to add some conditionals for the tasks there...

I guess this may also be pointing to wanting to set up windows/macos CI jobs here, but... is that even available for free tier github?

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -210,14 +210,17 @@ yaml-cpp    = ">=0.8.0,<0.9"
 eigen     = ">=3.4.0,<4"
 pinocchio = ">=3.7.0,<4"
 
-# Python binding dependencies
+# Python bindings dependencies
 python            = ">=3.11,<3.12"
 nanobind          = ">=2.8.0,<3"
 scikit-build-core = ">=0.11.5,<0.12"
 numpy             = "*"
-viser             = ">=1.0.1,<2"
-tyro              = ">=0.9.26,<0.10"
 xacro             = ">=2.1.1,<3"
+
+# Dependencies for examples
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -213,6 +217,7 @@ scikit-build-core = ">=0.11.5,<0.12"
 numpy             = "*"
 viser             = ">=1.0.1,<2"
 tyro              = ">=0.9.26,<0.10"
+xacro             = ">=2.1.1,<3"
```
eholum, line 218: I had to add `zstandard = ">=0.25.0,<0.26"` otherwise I was getting import errors with Viser - but that might be a them thing.

Do you not have that issue?

sea-bass: I did not have that issue with linux-64, no. 😢 

Let me add that to the full deps and test on my end.

eholum: Macs. They #JustWork.

sea-bass: This is very deep, but I think the Viser feedstock does not have a dependency on zstandard when it maybe should?

https://github.com/conda-forge/viser-feedstock/blob/70316cce2bef14770a6b8000bfe6a302b8f4afa7/recipe/meta.yaml#L24

However, the `pyproject.toml` does

https://github.com/nerfstudio-project/viser/blob/6bd9d2260bb5622176079aa261c74ecbbd26ac84/pyproject.toml#L53

---

put up a PR: https://github.com/conda-forge/viser-feedstock/pull/24

PR #119, opened 2026-02-04
Remove lambdas from oink python bindings and use Eigen::Ref

sjahr:
Follow-up on:
- https://github.com/open-planning/roboplan/pull/110#discussion_r2754251356: Removes the lambdas for the parameters constructors while keeping the option to use aggregated initialization
- https://github.com/open-planning/roboplan/pull/110#discussion_r2754229976: Use Eigen::Ref to avoid allocations inside solveIk in both Python & C++ 
- Rename params to options to make the naming consistent with RRT and SimpleIK

@sea-bass your proposed solution was correct by the way. I asked in nanobind repository: https://github.com/wjakob/nanobind/discussions/1282#discussion-9424889

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -24,7 +24,7 @@ concurrency:
 jobs:
 
   ubuntu_build_and_test:
-    timeout-minutes: 15
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
@@ -64,7 +64,7 @@ jobs:
 
   # Build ROS docker images and run tests.
   ros_build_and_test:
-    timeout-minutes: 15
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
@@ -103,7 +103,7 @@ jobs:
             /bin/bash -c './src/roboplan/scripts/run_tests.bash'
 
   pixi_build_and_test:
-    timeout-minutes: 15
+    timeout-minutes: 30
     env:
       SCCACHE_DIR: ${{ github.workspace }}/.sccache
     strategy:
--- a/bindings/src/modules/rrt.cpp
+++ b/bindings/src/modules/rrt.cpp
@@ -22,7 +22,10 @@ void init_rrt(nanobind::module_& m) {
       .def_ro("parent_id", &Node::parent_id, "The parent node ID.");
 
   nanobind::class_<RRTOptions>(m, "RRTOptions", "Options struct for RRT planner.")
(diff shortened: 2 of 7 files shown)


roboplan_oink/include/roboplan_oink/optimal_ik.hpp
```diff
@@ -126,15 +126,31 @@ struct Oink {
...
+  /// Example usage:
+  /// @code
+  /// Eigen::VectorXd delta_q(oink.num_variables);
+  /// auto result = oink.solveIk(tasks, constraints, scene, delta_q);
+  /// @endcode
   tl::expected<void, std::string>
   solveIk(const std::vector<std::shared_ptr<Task>>& tasks,
           const std::vector<std::shared_ptr<Constraints>>& constraints, const Scene& scene,
-          Eigen::VectorXd& delta_q);
+          Eigen::Ref<Eigen::VectorXd, 0, Eigen::InnerStride<Eigen::Dynamic>> delta_q);
```
sea-bass, line 153: Do you really need all these template parameters?

sea-bass:
> Looks like the CI failure is related to the ROS 2 build farm 🙈

Just more cache fun...

roboplan_oink/include/roboplan_oink/optimal_ik.hpp
sjahr: ~~On second thought: No. Let me update that~~
Third thought (feat. claude):
```bash
Eigen::Ref has three template parameters:


Eigen::Ref<MatrixType, Options, StrideType>
Simple form: Eigen::Ref<Eigen::VectorXd> defaults to StrideType = InnerStride<1>, meaning it only accepts contiguous memory (elements are adjacent).

nanobind s DRef<Eigen::VectorXd> converts to Eigen::Ref<VectorXd, 0, Stride<Dynamic, Dynamic>> — it allows any stride because numpy arrays can have arbitrary memory layouts (e.g., arr[::2] has stride 2).

The original type Eigen::Ref<VectorXd, 0, InnerStride<Dynamic>> accepts dynamic inner stride, which is more permissive than the simple form but still less general than what nanobind provides.

The conversion fails because:

nanobind gives: Stride<Dynamic, Dynamic> (both strides can vary)
Simple Ref expects: InnerStride<1> (unit stride only)
These are incompatible at compile time.
```

See here for more context: https://nanobind.readthedocs.io/en/latest/api_extra.html#eigen-convenience-type-aliases. I'd propose to keep it for now. What do you think?

sea-bass (approved):

PR #121, opened 2026-02-04
Increase CI timeouts

sea-bass:
Have been hitting these limits after deleting caches


.github/workflows/build_and_test.yml
```diff
@@ -24,7 +24,7 @@ concurrency:
 jobs:
 
   ubuntu_build_and_test:
-    timeout-minutes: 15
+    timeout-minutes: 30
```
eholum, line 27: CI go brrrrrrrrrrrrrrrrrrrr

PR #122, opened 2026-02-05
[oink] Control Barrier Functions

sjahr:
This PR adds control barrier functions to the oink solver and a simple box barrier as first reference example.

The demo is now updated to use a box barrier around the end-effector and you can use it to test it. While in theory, the control barrier guarantees safety, I encountered some issues in practice :sweat_smile: . The way we implement the barrier `-J_h · δq ≤ dt · α(h(q))` is a linearisation and thus only valid for smaller δq. It was relatively easy for me to push the robot out of the barrier by commanding a large jump of the reference frame. As soon as the frame is outside the barrier, the solver becomes unstable. For an actual controller implementation besides the proof of concept demo, this needs to be taken into account, e.g. by a quick re-validation of the barrier for the output of the QP.

This PR introduces 3 countermeasures to make it generally more robust:
- A safety margin for the barrier to prevent the system from overshooting
- Error saturation for the frame task to make sure the error stays within reasonable bounds and doesn't get too large. Maybe this would make sense for all tasks :thinking: 
- Input command lowpass filtering in the example to avoid large jumps of the reference frames
- Added enforceBarriers function to make sure, the barrier is not violated at h(q + δq)

Happy to hear any feedback about this new addition!

### DIFF
--- a/roboplan_examples/python/example_oink.py
+++ b/roboplan_examples/python/example_oink.py
@@ -233,9 +233,7 @@ def main(
 
                     # Solve IK for one step with constraints
                     try:
-                        oink.solveIk(
-                            tasks, constraints, barriers, scene, delta_q, regularization
-                        )
+                        oink.solveIk(tasks, constraints, scene, delta_q, regularization)
                     except RuntimeError as e:
                         delta_q = np.zeros(num_variables)
                         print(f"Warning: IK solver failed: {e}, using zero delta_q")
@@ -303,9 +301,6 @@ def main(
             raw_targets[idx] = initial_pose.copy()
             if reference_filter_tau > 0:
                 reference_filters[idx].reset(initial_pose)
-
-    # No barriers in this example (see example_oink_with_barriers.py for a tutorial on barriers)
-    barriers = []
 
     viz.display(q_full)
     reset_position(None)


roboplan_examples/python/example_oink.py
```diff
@@ -30,11 +32,17 @@ def main(
...
+    safety_margin: float = 0.05,
+    max_position_error: float = 0.15,
+    max_rotation_error: float = 0.5,
+    reference_filter_tau: float = 0.1,
     host: str = "localhost",
     port: str = "8000",
 ):
     """
-    Run the optimal IK example with the provided parameters.
+    Run the optimal IK example with position barrier constraints.
```
sea-bass: I feel like the barrier should be optional, or at least contained in a totally separate example to not bloat the existing code.

roboplan_examples/python/example_oink.py
```diff
@@ -203,9 +246,23 @@ def control_loop():
...
+                    # Update reference filters if enabled (smooths target pose changes)
+                    # The filter gradually approaches the raw target to prevent sudden jumps
+                    if reference_filter_tau > 0:
+                        for idx, ref_filter in enumerate(reference_filters):
+                            filtered_target = ref_filter.update(raw_targets[idx], dt)
+                            frame_tasks[idx].setTargetFrameTransform(filtered_target)
+                    else:
+                        # No filtering - use raw targets directly
+                        for idx in range(len(frame_tasks)):
+                            frame_tasks[idx].setTargetFrameTransform(raw_targets[idx])
```
sea-bass, line 232: can you save yourself one indent/logic by just checking the size of `reference_filters`?

roboplan_examples/python/example_oink.py
```diff
@@ -240,9 +297,15 @@ def reset_position(_):
...
-                fk_tform = scene.forwardKinematics(q_current, tasks[idx].frame_name)
+                fk_tform = scene.forwardKinematics(
+                    q_current, frame_tasks[idx].frame_name
+                )
                 controls.position = fk_tform[:3, 3]
                 controls.wxyz = pin.Quaternion(fk_tform[:3, :3]).coeffs()[[3, 0, 1, 2]]
+                # Reset raw target and filter state to current pose
+                raw_targets[idx] = fk_tform.copy()
+                if reference_filter_tau > 0:
+                    reference_filters[idx].reset(fk_tform)
```
sea-bass, line 279: again here you can just loop through all the reference filters instead to reset, no need to check with an `if`

roboplan_examples/python/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    Smoothly interpolates between the current filtered pose and the target pose
+    using exponential filtering for position and SLERP for orientation.
+    """
+
+    def __init__(self, tau: float = 0.1):
+        """Initialize the filter.
+
+        Args:
+            tau: Time constant in seconds. Larger values = slower, smoother tracking.
+                 tau=0.1 means ~63% of the step is taken per 0.1 seconds.
```
sea-bass:
```suggestion
                 tau=0.1 means ~63% of the step is taken per 0.1 second.

File: roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+  /// @param gain Barrier gain (gamma), controls convergence to safe set. Default 1.0
+  /// @param dt Timestep matching your control loop period. Default 0.01 (100 Hz)
+  /// @param safe_displacement_gain Gain for safe displacement regularization. Default 1.0
+  /// @param safety_margin Conservative margin for hard constraint guarantee. Default 0.0
+  /// @note The dt parameter significantly affects barrier behavior - ensure it matches
+  ///       your actual control/integration timestep
+  PositionBarrier(const std::string& frame_name, const Eigen::Vector3d& p_min,
+                  const Eigen::Vector3d& p_max, int num_variables, double gain = 1.0,
+                  double dt = 0.01, double safe_displacement_gain = 1.0,
+                  double safety_margin = 0.0);
```
sea-bass: feels like `dt` shouldn't have a default value, should be required

roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp
```diff
@@ -0,0 +1,81 @@
...
+  /// @param frame_name Name of the frame to constrain
+  /// @param indices Vector of axis indices to constrain (0=x, 1=y, 2=z)
+  /// @param p_min Minimum bounds for selected axes in world frame (size must match indices)
+  /// @param p_max Maximum bounds for selected axes in world frame (size must match indices)
+  /// @param num_variables Number of optimization variables (model.nv)
+  /// @param gain Barrier gain (gamma), controls convergence to safe set. Default 1.0
+  /// @param dt Timestep matching your control loop period. Default 0.01 (100 Hz)
+  /// @param safe_displacement_gain Gain for safe displacement regularization. Default 1.0
+  /// @param safety_margin Conservative margin for hard constraint guarantee. Default 0.0
+  PositionBarrier(const std::string& frame_name, const std::vector<int>& indices,
```
sea-bass: rather than an overload, do you want to just create a method to set the axes indices?

roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp
```diff
@@ -0,0 +1,81 @@
...
+                  const Eigen::VectorXd& p_min, const Eigen::VectorXd& p_max, int num_variables,
+                  double gain = 1.0, double dt = 0.01, double safe_displacement_gain = 1.0,
+                  double safety_margin = 0.0);
+
+  int getNumBarriers(const Scene& scene) const override;
+
+  tl::expected<void, std::string> computeBarrier(const Scene& scene) override;
+
+  tl::expected<void, std::string> computeJacobian(const Scene& scene) override;
```
sea-bass, line 84: missing docstrings?

roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp
```diff
@@ -0,0 +1,81 @@
...
+  Eigen::Vector3d getFramePosition(const Scene& scene) const;
+
+  const std::string frame_name;
+  const std::vector<int> indices;  ///< Axes to constrain (0, 1, 2 for x, y, z)
+  const Eigen::VectorXd p_min;     ///< Min bounds for each constrained axis
+  const Eigen::VectorXd p_max;     ///< Max bounds for each constrained axis
+
+private:
+  mutable pinocchio::FrameIndex frame_id = 0;
+  mutable bool frame_id_cached = false;
```
sea-bass, line 122: missing docstrings

roboplan_oink/include/roboplan_oink/tasks/frame.hpp
```diff
@@ -28,6 +29,16 @@ struct FrameTaskOptions {
...
+
+  /// @brief Maximum position error magnitude in meters (default: unlimited).
+  /// Limits the position error norm to prevent large jumps that can invalidate
+  /// CBF linearization. Recommended: 0.1-0.2m for systems with barriers.
+  double max_position_error = std::numeric_limits<double>::infinity();
+
+  /// @brief Maximum rotation error magnitude in radians (default: unlimited).
+  /// Limits the rotation error norm to prevent large jumps.
+  /// Recommended: 0.5-1.0 rad for systems with barriers.
+  double max_rotation_error = std::numeric_limits<double>::infinity();
```
sea-bass, line 41: do these need to exist for the task, or can they be clipped at the barrier level?

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,133 @@
...
+PositionBarrier::PositionBarrier(const std::string& frame_name_, const std::vector<int>& indices_,
+                                 const Eigen::VectorXd& p_min_, const Eigen::VectorXd& p_max_,
+                                 int num_variables_, double gain, double dt,
+                                 double safe_displacement_gain, double safety_margin)
+    : Barrier(gain, dt, safe_displacement_gain, safety_margin), frame_name(frame_name_),
+      indices(indices_), p_min(p_min_), p_max(p_max_) {
+  // Validate indices
+  for (int idx : indices_) {
+    if (idx < 0 || idx > 2) {
+      throw std::invalid_argument("Axis index must be 0, 1, or 2 (x, y, z)");
```
sea-bass: Might want to also add what the failing number / index was

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  // Validate sizes match
+  if (static_cast<int>(indices_.size()) != p_min_.size() ||
+      static_cast<int>(indices_.size()) != p_max_.size()) {
+    throw std::invalid_argument("p_min and p_max size must match indices size");
+  }
+
+  // Count active constraints (finite bounds)
+  int num_barriers = 0;
+  for (int i = 0; i < static_cast<int>(indices_.size()); ++i) {
```
sea-bass: instead of `static_cast` you could make the loop variable `size_t`

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    // Lower bound barrier: h = p - p_min >= 0 when p >= p_min (safe)
+    if (std::isfinite(p_min[i])) {
+      barrier_values[idx++] = p_i - p_min[i];
+    }
+
+    // Upper bound barrier: h = p_max - p >= 0 when p <= p_max (safe)
+    if (std::isfinite(p_max[i])) {
+      barrier_values[idx++] = p_max[i] - p_i;
+    }
```
sea-bass: personal preference for readability to do the indexing and incrementing in 2 separate lines

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    // Lower bound: h = p - p_min, so J_h = dp/dq = J_p[axis, :]
+    if (std::isfinite(p_min[i])) {
+      jacobian_container.row(idx++) = frame_jacobian.row(axis);
+    }
+
+    // Upper bound: h = p_max - p, so J_h = -dp/dq = -J_p[axis, :]
+    if (std::isfinite(p_max[i])) {
+      jacobian_container.row(idx++) = -frame_jacobian.row(axis);
+    }
```
sea-bass: same comment for readability

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -76,9 +180,31 @@ Oink::solveIk(const std::vector<std::shared_ptr<Task>>& tasks,
...
+  for (const auto& barrier : barriers) {
+    auto obj_result =
+        barrier->computeQpObjective(scene, barrier_H_contribution, barrier_c_contribution);
+    if (obj_result.has_value()) {
+      // Add dense contribution to sparse H
+      for (int i = 0; i < num_variables; ++i) {
+        for (int j = 0; j < num_variables; ++j) {
+          H.coeffRef(i, j) += barrier_H_contribution(i, j);
+        }
+      }
```
sea-bass: is this any more/less efficient than just incrementing `H` by `barrier_H_contribution` in one operation?

roboplan_oink/include/roboplan_oink/optimal_ik.hpp
```diff
@@ -149,13 +257,17 @@ struct Oink {
...
+  /// // With barriers and custom regularization:
+  /// auto result = oink.solveIk(tasks, constraints, barriers, scene, delta_q, 1e-6);
   /// @endcode
   tl::expected<void, std::string>
   solveIk(const std::vector<std::shared_ptr<Task>>& tasks,
-          const std::vector<std::shared_ptr<Constraints>>& constraints, const Scene& scene,
+          const std::vector<std::shared_ptr<Constraints>>& constraints,
+          const std::vector<std::shared_ptr<Barrier>>& barriers, const Scene& scene,
           Eigen::Ref<Eigen::VectorXd, 0, Eigen::InnerStride<Eigen::Dynamic>> delta_q,
           double regularization = 1e-12);
```
sea-bass, line 327: it might be worthwhile to add overloads to this method so you can call `solveIk` with no constraints/barriers and not have to specify a bunch of empty vectors.

roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp
sjahr: Unfortunately no, because the memory allocation happens in the constructor but I think this should be a single constructor with default argument.

roboplan_oink/include/roboplan_oink/tasks/frame.hpp
sjahr: I think we actually need them for every task. My understanding right now is that if we allow for unbounded errors, the weighting can be invalidated by e.g. by commanding a very very large jump in the task frame. That also messes with the barriers, because the linearization only works in a bounded local environment around the current position. Let's discuss this in our next meeting in more detail!

roboplan_oink/src/barriers/position_barrier.cpp
sjahr: This got removed

roboplan_oink/src/barriers/position_barrier.cpp
sjahr: Same here

roboplan_examples/python/example_oink_with_barriers.py
```diff
@@ -0,0 +1,397 @@
...
+):
+    """
+    Tutorial on optimal IK with Control Barrier Functions (CBF) safety constraints.
+
+    This example demonstrates how to use PositionBarrier to enforce safety constraints
+    on end-effector motion. Barriers prevent the robot from leaving a safe region while
+    tracking target poses.
+
+    Parameters:
+        model: The name of the model to use (ur5, franka, or dual).
```
sea-bass: kinova and so101?

roboplan_examples/python/filters.py
sea-bass, line 1: Just `utils` is way too vague.

Maybe call this `filters.py`?

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+Eigen::Vector3d PositionBarrier::getFramePosition(const Scene& scene) const {
+  return scene.getData().oMf[frame_id].translation();
+}
+
+double PositionBarrier::evaluateAtConfiguration(const pinocchio::Model& model,
+                                                pinocchio::Data& data,
+                                                const Eigen::VectorXd& q) const {
+  // Compute FK for the candidate configuration
+  pinocchio::forwardKinematics(model, data, q);
```
sea-bass: should we like do the forward kinematics in the main `oink` loop before evaluating all the costs/constraints/barriers so we're not doing them multiple times?

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+  // Get frame ID (use cached value if available, otherwise look up)
+  pinocchio::FrameIndex fid = frame_id;
+  if (!frame_id_cached) {
+    if (!model.existFrame(frame_name)) {
+      return std::numeric_limits<double>::infinity();  // Frame not found
+    }
+    fid = model.getFrameId(frame_name);
+  }
```
sea-bass, line 184: the `Scene` class has methods that already put these into dictionaries, consider using those

roboplan_oink/test/tasks/test_frame_task.cpp
```diff
@@ -344,7 +344,7 @@ TEST_F(FrameTaskTest, GradientDirectionTowardTarget) {
 
   // Solve IK
   Eigen::VectorXd delta_q(num_variables_);
-  auto result = oink.solveIk(tasks, constraints, *scene_, delta_q);
+  auto result = oink.solveIk(tasks, constraints, {}, *scene_, delta_q);
```
sea-bass: now that we have the overloads, we can omit this extra `{}` right? Do check this in all other places

roboplan_oink/README.md
sea-bass, line 1: this whole page is great, but should go in the actual docs!

roboplan_examples/python/example_oink_with_barriers.py
sjahr: Removed the list. This way it will never get outdated

roboplan_oink/src/barriers/position_barrier.cpp
sjahr: I like it, but maybe for now I'd keep it like this for the sake of readability. I need to think about a good way to ensure that when you call evaluateAtConfiguration returns a meaningful error if forward kinematics is not called before evaluateAtConfiguration. Let's do this in a follow-up PR

sjahr: I create an issue to track this: #168

external/toppra
sea-bass, line 1: make sure you aren't updating `toppra` submodule! Bring it up to date with `main`

roboplan_examples/python/example_oink.py
```diff
@@ -30,11 +31,14 @@ def main(
     lm_damping: float = 0.01,
     regularization: float = 1e-6,
     control_freq: float = 100.0,
+    max_position_error: float = 0.15,
+    max_rotation_error: float = 0.5,
```
sea-bass: for this non-barrier example since we split it out, can we keep these as defaults of infinity without affecting performance? (thereby simplifying this example and only adding them in the barrier example)

roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp
sea-bass: still missing!

roboplan_oink/include/roboplan_oink/optimal_ik.hpp
```diff
@@ -111,27 +111,201 @@ struct Constraints {
...
+  explicit Oink(int num_variables);
 
   /// @brief Constructor with custom settings
   ///
   /// @param num_variables Number of optimization variables (typically number of DOFs)
   /// @param custom_settings Custom OSQP solver settings
   Oink(int num_variables, const OsqpEigen::Settings& custom_settings);
 
-  /// @brief Solve inverse kinematics for given tasks and constraints
+  /// @brief Solve inverse kinematics for tasks only-
```
sea-bass: ultra nit -- remove the trailing dash

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+PositionBarrier::PositionBarrier(const std::string& frame_name, const Eigen::Vector3d& p_min,
+                                 const Eigen::Vector3d& p_max, int num_variables, double dt,
+                                 const ConstraintAxisSelection& axis_selection, double gain,
+                                 double safe_displacement_gain, double safety_margin)
+    : Barrier(gain, dt, safe_displacement_gain, safety_margin), frame_name(frame_name),
+      axis_selection(axis_selection), p_min(p_min), p_max(p_max) {
+  // Validate that p_min < p_max for enabled axes with finite bounds
+  const std::array<bool, 3> axes_enabled = {axis_selection.x, axis_selection.y, axis_selection.z};
+  const std::array<char, 3> axis_names = {'x', 'y', 'z'};
```
sea-bass: wondering if this instead would work better with an enum instead of char, but not major

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,224 @@
...
+  Eigen::Vector3d p = getFramePosition(scene);
+
+  // Compute barrier values for each active constraint
+  int idx = 0;
+
+  // X axis
+  if (axis_selection.x) {
+    if (std::isfinite(p_min[0])) {
+      barrier_values[idx] = p[0] - p_min[0];
+      idx++;
```
sea-bass: for all of these, `++idx` works and is slightly faster

roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -0,0 +1,224 @@
...
+                                    "] (got " + std::to_string(p_min[i]) +
+                                    " >= " + std::to_string(p_max[i]) + ")");
+      }
+    }
+  }
+  // Count active constraints (finite bounds and enabled axes)
+  int num_barriers = 0;
+  if (axis_selection.x) {
+    if (std::isfinite(p_min[0]))
+      num_barriers++;
```
sea-bass: for all of these `++num_barriers` works and is slightly faster

roboplan_oink/src/tasks/frame.cpp
```diff
@@ -49,6 +53,28 @@ tl::expected<void, std::string> FrameTask::computeError(const Scene& scene) {
   error_container.head<3>() = e_pos;
   error_container.tail<3>() = e_rot;
 
+  // Soft saturation of position error using tanh for smooth gradients
+  // This prevents large jumps that can invalidate CBF linearization while maintaining
+  // smooth error dynamics. Uses saturate(e) = e_max * tanh(||e|| / e_max) * (e / ||e||)
+  if (std::isfinite(max_position_error)) {
+    Eigen::Vector3d pos_error = error_container.head<kPositionDimension>();
+    const double pos_norm = pos_error.norm();
+    if (pos_norm > 1e-9) {  // Avoid division by zero
```
sea-bass: make this into a named constant, this is a magic number atm

roboplan_oink/src/tasks/frame.cpp
```diff
@@ -49,6 +53,28 @@ tl::expected<void, std::string> FrameTask::computeError(const Scene& scene) {
...
+      const double scale = max_position_error * std::tanh(pos_norm / max_position_error) / pos_norm;
+      error_container.head<kPositionDimension>() = pos_error * scale;
+    }
+  }
+
+  // Soft saturation of rotation error using tanh for smooth gradients
+  if (std::isfinite(max_rotation_error)) {
+    Eigen::Vector3d rot_error = error_container.tail<kOrientationDimension>();
+    const double rot_norm = rot_error.norm();
+    if (rot_norm > 1e-9) {  // Avoid division by zero
```
sea-bass: and reuse the constant here

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -1,8 +1,123 @@
...
+  auto jacobian_result = computeJacobian(scene);
+  if (!jacobian_result) {
+    return jacobian_result;
+  }
+
+  // Compute squared Frobenius norm of Jacobian: ‖J_h‖²
+  const double jacobian_norm_sq = jacobian_container.squaredNorm();
+
+  // Avoid division by zero - if Jacobian is near zero, no regularization needed
+  constexpr double kMinNormSq = 1e-12;
```
sea-bass: put this constant up in an anonymous namespace instead of here

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -1,8 +1,123 @@
...
+  // QP objective contribution: weight · ‖δq - δq_safe‖²
+  // = weight · (δq^T δq - 2 δq^T δq_safe + δq_safe^T δq_safe)
+  // = weight · δq^T I δq - 2·weight · δq_safe^T δq + const
+  //
+  // H_contribution = 2·weight · I  (factor of 2 because QP is 1/2 x^T H x)
+  // c_contribution = -2·weight · δq_safe
+
+  H.setIdentity();
+  H *= 2.0 * weight;
+  c = -2.0 * weight * dq_safe;
```
sea-bass: kinda nitpick, but if you exclude the `2.0` from the definition of `weight`, would that cancel out with the `2.0` in these `H`/`c` calculations and you can just remove all 3 instances?

sea-bass (commented):
Home stretch! I read the changes in more detail (except for the 800+(!!!) lines of tests) and only have minor comments now.

Fine with your proposal, but can you please make an issue for converting the `SE3` low-pass filter to C++ w/ Python bindings so we can keep track of this?

roboplan_oink/src/barriers/position_barrier.cpp
sjahr: I think it would make things a bit more complicated. But I made a local constant and moved it out of the function

roboplan_oink/src/optimal_ik.cpp
sjahr: Good catch!

PR #124, opened 2026-02-07
Fix usage of tinyxml2 and tl_expected dependencies

sea-bass:
This PR does two things:

1. Stops depending on `tinyxml2_vendor` package in newer ROS distros (Rolling / Lyrical and later) by using conditionals in the `roboplan/package.xml`. This is to avoid deprecation warnings.
2. If using ROS, relies on binary version of `tl_expected` instead of building it from source. This adds a little LLM-assisted CMake magic to handle these cases.

### DIFF
--- a/bindings/CMakeLists.txt
+++ b/bindings/CMakeLists.txt
@@ -135,14 +135,25 @@ target_include_directories(roboplan_ext PUBLIC
   $<INSTALL_INTERFACE:include>
 )
 set_property(TARGET roboplan_ext PROPERTY CXX_STANDARD 20)
+
+# Ensures that the library is always called tl_expected::tl_expected regardless
+# of whether it was built from source in roboplan or comes from ROS binaries.
+find_package(tl_expected QUIET)
+if(NOT TARGET tl_expected::tl_expected AND TARGET roboplan::tl_expected)
+    add_library(tl_expected::tl_expected INTERFACE IMPORTED)
+    set_target_properties(tl_expected::tl_expected PROPERTIES
+        INTERFACE_LINK_LIBRARIES roboplan::tl_expected
+    )
+endif()
+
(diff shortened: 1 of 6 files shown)


bindings/CMakeLists.txt
```diff
@@ -135,14 +135,25 @@ target_include_directories(roboplan_ext PUBLIC
...
 set_property(TARGET roboplan_ext PROPERTY CXX_STANDARD 20)
+
+# Ensures that the library is always called tl_expected::tl_expected regardless
+# of whether it was built from source in roboplan or comes from ROS binaries.
+find_package(tl_expected QUIET)
+if(NOT TARGET tl_expected::tl_expected AND TARGET roboplan::tl_expected)
+    add_library(tl_expected::tl_expected INTERFACE IMPORTED)
+    set_target_properties(tl_expected::tl_expected PROPERTIES
+        INTERFACE_LINK_LIBRARIES roboplan::tl_expected
+    )
```
eholum, line 146: Ah, so basically making an empty interface library that wraps up the roboplan source lib if the package isn't already installed? Neat.

roboplan/package.xml
```diff
@@ -11,8 +11,10 @@
 
   <depend>eigen</depend>
   <depend>pinocchio</depend>
-  <depend>tinyxml2</depend>
-  <depend>tinyxml2_vendor</depend>
+  <!-- NOTE: Ensure to update this as humble, jazzy, and kilted go end of life. -->
+  <depend condition="$ROS_DISTRO != humble and $ROS_DISTRO != jazzy and $ROS_DISTRO != kilted">tinyxml2</depend>
```
eholum, line 15: This is hilarious.

sea-bass: This holds up the entire fabric of robotics software wdym

PR #125, opened 2026-02-08
Support multiple tip frames in simple IK

### LINKED ISSUE #68: Support multiple tip frames in IK solver
This should be pretty straightforward to do, and now that we have the dual Franka we can test this.

Unclear whether we want to introduce scalar vs. vector target pose overloads of the `solveIk` function, as there is a tradeoff between ease of use and ease of Python bindings development. But probably worth it.

This would also include modifying the demo to have 2 interactive markers; one for each end effector frame.

... but first, let's support joint groups in #41 as this will have some sweeping API changes.

sea-bass:
Lots of Jacobian and error stacking fun!

[multi_tip_ik_roboplan.webm](https://github.com/user-attachments/assets/3419eb3a-3fe9-4916-8c7c-b5493dcaf708)

Closes #68

### DIFF
--- a/bindings/CMakeLists.txt
+++ b/bindings/CMakeLists.txt
@@ -135,14 +135,25 @@ target_include_directories(roboplan_ext PUBLIC
   $<INSTALL_INTERFACE:include>
 )
 set_property(TARGET roboplan_ext PROPERTY CXX_STANDARD 20)
+
+# Ensures that the library is always called tl_expected::tl_expected regardless
+# of whether it was built from source in roboplan or comes from ROS binaries.
+find_package(tl_expected QUIET)
+if(NOT TARGET tl_expected::tl_expected AND TARGET roboplan::tl_expected)
+    add_library(tl_expected::tl_expected INTERFACE IMPORTED)
+    set_target_properties(tl_expected::tl_expected PROPERTIES
+        INTERFACE_LINK_LIBRARIES roboplan::tl_expected
+    )
+endif()
+
 target_link_libraries(roboplan_ext PRIVATE
-  roboplan::expected
   roboplan::roboplan
   roboplan_example_models::roboplan_example_models
   roboplan_oink::roboplan_oink
   roboplan_rrt::roboplan_rrt
   roboplan_simple_ik::roboplan_simple_ik
   roboplan_toppra::roboplan_toppra
+  tl_expected::tl_expected
 )
 
 if(AMENT_BUILD)
--- a/bindings/package.xml
+++ b/bindings/package.xml
@@ -17,6 +17,7 @@
   <depend>roboplan_rrt</depend>
   <depend>roboplan_simple_ik</depend>
   <depend>roboplan_toppra</depend>
+  <depend>tl_expected</depend>
 
   <test_depend>roboplan_example_models</test_depend>
 
--- a/roboplan/CMakeLists.txt
+++ b/roboplan/CMakeLists.txt
@@ -16,14 +16,22 @@ find_package(yaml-cpp REQUIRED)
 # issues with symlink installs.
 find_package(ament_cmake QUIET)
 
-# Add libraries
-add_library(expected INTERFACE)
-target_include_directories(expected
-  INTERFACE
-    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
-    $<INSTALL_INTERFACE:include>)
+set(TARGETS_LIST roboplan)
 
-# ROS already has a TinyXML2 vendor, so add this only in non-ROS contexts.
+# Add tl_expected from source only if not found.
+find_package(tl_expected QUIET)
+if (NOT tl_expected_FOUND)
+    add_library(tl_expected INTERFACE)
+    target_include_directories(
+        tl_expected
+        INTERFACE
+            $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+            $<INSTALL_INTERFACE:include>
+    )
+    list(APPEND TARGETS_LIST tl_expected)
+endif()
+
+# Add tinyxml2 from source only if not found.
 find_package(tinyxml2_vendor QUIET)
 find_package(TinyXML2 QUIET)
 if ( NOT tinyxml2_vendor_FOUND AND NOT TinyXML2_FOUND )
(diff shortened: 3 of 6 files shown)


bindings/src/roboplan/roboplan_ext/optimal_ik.pyi
sea-bass, line 1: note @sjahr all of these `.pyi` changes get auto-generated on build, but I assume you never committed them because our Pixi setup (by default) also turns these files into symlinks. So adding it all here.

roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -18,31 +18,28 @@ SimpleIk::SimpleIk(const std::shared_ptr<Scene> scene, const SimpleIkOptions& op
...
   const auto& v_indices = joint_group_info_.v_indices;
 
-  const auto goal_tform = pinocchio::SE3(goal.tform);
   solution = start;
   auto q = scene_->toFullJointPositions(options_.group_name, start.positions);
 
+  const auto n_dims = 6 * goals.size();
+  error_.resize(n_dims, 1);
+  jacobian_.resize(n_dims, v_indices.size());
+  jjt_.resize(n_dims, n_dims);
```
sea-bass: it's mildly annoying that we can't resize these vectors as architected now, because we don't know the number of frames at initialization time. Any ideas?

One idea that balances simplicity and performance is for the `SimpleIkOptions` to specify a number of frames a priori, and then the solve loop simply returns unexpected if the number of frames you pass in at solve time doesn't match?

bindings/src/roboplan/roboplan_ext/optimal_ik.pyi
sea-bass: actually just to keep changes clean I'm merging them in https://github.com/open-planning/roboplan/pull/127

bindings/src/modules/simple_ik.cpp
```diff
@@ -35,8 +36,14 @@ void init_simple_ik(nanobind::module_& m) {
...
-      .def("solveIk", &SimpleIk::solveIk, "Solves inverse kinematics.", "goal"_a, "start"_a,
-           "solution"_a);
+      .def("solveIk",
+           (bool (SimpleIk::*)(const CartesianConfiguration&, const JointConfiguration&,
+                               JointConfiguration&))&SimpleIk::solveIk,
+           "Solves inverse kinematics (single goal).", "goal"_a, "start"_a, "solution"_a)
+      .def("solveIk",
+           (bool (SimpleIk::*)(const std::vector<CartesianConfiguration>&,
+                               const JointConfiguration&, JointConfiguration&))&SimpleIk::solveIk,
+           "Solves inverse kinematics (multiple goal).", "goals"_a, "start"_a, "solution"_a);
```
sea-bass, line 47: Somehow `nb::overload_cast` usage didn't seem to compile for me...

roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -56,25 +53,46 @@ bool SimpleIk::solveIk(const CartesianConfiguration& goal, const JointConfigurat
...
+        jacobian_.block(idx * 6, 0, 6, v_indices.size()) =
+            full_jacobian_(Eigen::placeholders::all, v_indices);
+      }
+
+      // Finalize solution if the error is below threshold.
+      // Note the error norm is accumulated per target frame (batch of 6).
+      double error_norm = 0.0;
+      for (size_t idx = 0; idx < goals.size(); ++idx) {
+        error_norm += error_.segment(idx * 6, 6).norm();
+      }
```
sea-bass, line 91: is this truly necessary? I guess anyways leaving this here will help for #126

roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -56,25 +53,46 @@ bool SimpleIk::solveIk(const CartesianConfiguration& goal, const JointConfigurat
...
+      // Loop through all the frames and accumulate errors and Jacobians.
+      for (size_t idx = 0; idx < goals.size(); ++idx) {
+        const auto& goal = goals.at(idx);
+        const auto goal_tform = pinocchio::SE3(goal.tform);
+
+        const auto frame_id_result = scene_->getFrameId(goal.tip_frame);
+        if (!frame_id_result) {
+          throw std::runtime_error("Failed to get frame ID: " + frame_id_result.error());
+        }
+        auto frame_id = frame_id_result.value();
```
eholum: Might be worth pulling this out of the loops, right? Saves a string lookup every time. Could pre-compute and store in a pre-allocated vector or something.

roboplan_simple_ik/src/simple_ik.cpp
eholum: >  for the SimpleIkOptions to specify a number of frames a priori

I don't hate that idea? We already require a joint group, so in some ways it would make sense to specify frames in that joint group for the particular instance of the solver.

Though is the annoyance just that we're resizing them every call? Because I bet the Eigen does nothing or at least returns very fast if the sizes haven't actually changed.

sea-bass: It is just that... maybe I leave it, then. The flexibility of being able to switch frames for bigger groups seems nice and almost free

roboplan_simple_ik/src/simple_ik.cpp
eholum: Necessary as in putting it here? Theoretically you could compute it in the for loop above after getting the error?

As you say though, will help with separating the error types...

sea-bass: Sorry, I meant is it necessary to separately compute the norms of each frame error (the 6-element block) and add them together, vs. just take the whole norm of the 6*N length vector

eholum: Oh I get it. I'm reasonably sure this is what you want? At least for now. We're more interested in the error on each individual frame.  We definitely want the sum of the norms, rather than the norm of the sums. Though since it's just a check on completeness... it probably doesn't _really_ matter since users can just turn the acceptable error knob up and down.

I think the way you have it is basically just [giving equal weight to each error](https://ntrs.nasa.gov/api/citations/19950005142/downloads/19950005142.pdf), but down the road we could add the weightings that prioritize different things (assuming I understand this).

sea-bass: Wonderful! In that case let's ship it

PR #128, opened 2026-02-11
Separates 6D IK error tolerance into linear (meters) and angular (radians) components

kumar-sanjeeev:
This PR splits the IK error tolerance into linear and angular components, with updates to python bindings and examples

**Logic:** 
Current implementation, checking error tolerances for each linear and angular error per target frame. Convergence is only achieved if all frames satisfy both thresholds.

### DIFF
--- a/bindings/src/modules/simple_ik.cpp
+++ b/bindings/src/modules/simple_ik.cpp
@@ -29,9 +29,9 @@ void init_simple_ik(nanobind::module_& m) {
       .def_rw("step_size", &SimpleIkOptions::step_size, "The integration step for the solver.")
       .def_rw("damping", &SimpleIkOptions::damping, "Damping value for the Jacobian pseudoinverse.")
       .def_rw("max_linear_error_norm", &SimpleIkOptions::max_linear_error_norm,
-              "The maximum linear error norm.")
+              "The maximum linear error norm, in meters.")
       .def_rw("max_angular_error_norm", &SimpleIkOptions::max_angular_error_norm,
-              "The maximum angular error norm.")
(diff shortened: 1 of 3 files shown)


roboplan_examples/python/example_ik.py
```diff
@@ -19,7 +19,8 @@ def main(
     model: str = "ur5",
     max_iters: int = 100,
     step_size: float = 0.25,
-    max_error_norm: float = 0.001,
+    max_linear_error: float = 0.001,
```
sea-bass: These seem to be missing the `_norm`

roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -83,13 +83,22 @@ bool SimpleIk::solveIk(const std::vector<CartesianConfiguration>& goals,
...
-      // Note the error norm is accumulated per target frame (batch of 6).
-      double error_norm = 0.0;
+      // Check for convergence, ensure every target frame meets both
+      // its linear and angular error tolerances
+      bool converged = true;
       for (size_t idx = 0; idx < n_frames; ++idx) {
-        error_norm += error_.segment(idx * 6, 6).norm();
+        const auto target_frame_error = error_.segment(idx * 6, 6);
+        double linear_error = target_frame_error.head<3>().norm();
+        double angular_error = target_frame_error.tail<3>().norm();
```
sea-bass: nit: instead of `double` you can just make these `const auto`

sea-bass (approved):
nice! I just made a really nitpicky style change, but tested and LGTM

EDIT: I also committed the changes to the `.pyi` stub file

PR #131, opened 2026-02-14
Simplify FrameTask interface and allow modifying target transforms at runtime

### LINKED ISSUE #130: Allow oink frame task to be modified rather than recreated
It's very inefficient to create a new `FrameTask` every time in a loop and then rebuild a list of tasks.

We could just add a method to update the transform and keep reusing the same task.

sea-bass:
This PR does a few things related to `FrameTask` in oink:

* Allows modifying the target transform of the `FrameTask` on the fly, so you don't have to create new tasks every time on a loop (as that is inefficient)
* Simplifies the constructor since the `CartesianConfiguration` class already includes the tip frame name
* Pre-computes the frame ID once so we don't have to keep doing string lookups in a loop
* Makes the convention of transform variables less "backwards"
* Simplifies the Python example for oink, and sets some default values to more sensible numbers

Closes #130

### DIFF
--- a/roboplan_oink/src/tasks/frame.cpp
+++ b/roboplan_oink/src/tasks/frame.cpp
@@ -26,7 +26,7 @@ tl::expected<void, std::string> FrameTask::computeError(const Scene& scene) {
   if (!frame_id.has_value()) {
     const auto maybe_frame_id = scene.getFrameId(frame_name);
     if (!maybe_frame_id) {
-      throw std::runtime_error("Frame '" + frame_name + "' not found: " + maybe_frame_id.error());
+      return tl::make_unexpected("Frame '" + frame_name + "' not found: " + maybe_frame_id.error());
     }
     frame_id = maybe_frame_id.value();
   }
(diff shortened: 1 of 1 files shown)


roboplan_oink/include/roboplan_oink/tasks/frame.hpp
```diff
@@ -96,7 +84,20 @@ struct FrameTask : public Task {
   /// @return A 6×6 diagonal weight matrix.
   static Eigen::MatrixXd createWeightMatrix(double position_cost, double orientation_cost);
 
-  // Pre-allocated logarithmic Jacobian (mutable for use in const computeJacobian)
+  /// @brief Sets the target transform for this frame task.
+  /// @param tform The target transform.
+  void setTargetFrameTransform(const Eigen::Matrix4d& tform) { target_pose.tform = tform; }
+
+  /// @brief Name of the frame to track (e.g., end-effector link name).
+  std::string frame_name;
```
sjahr, line 92: Nit: Should we get rid of this variable and just use target_pose.tip_frame everywhere instead?

sea-bass: I thought about it too, but it's nice to have the variable more easily accessible in the public API

PR #133, opened 2026-02-14
Add cmake option to generate Python stubs

sea-bass:
Needed for cross-compilation support in conda

### DIFF
--- a/bindings/CMakeLists.txt
+++ b/bindings/CMakeLists.txt
@@ -147,7 +147,6 @@ target_link_libraries(roboplan_ext PRIVATE
   roboplan_toppra::roboplan_toppra
   tl_expected::tl_expected
 )
-set_target_properties(roboplan_ext PROPERTIES PREFIX "")
 
 if(AMENT_BUILD)
   ament_get_python_install_dir(PYTHON_INSTALL_DIR)


bindings/CMakeLists.txt
eholum: Lots of conditionals but I think this needs to be in a check? We don't need these if we don't have the stubs, and it creates an error otherwise:

E.g.

```
$ colcon build --cmake-args -DGENERATE_PYTHON_STUBS=OFF --packages-select roboplan_bindings 
...
--- stderr: roboplan_bindings                                
CMake Error at cmake_install.cmake:83 (file):
  file INSTALL cannot find
  "/workspace/roboplan_ws/build/roboplan_bindings/roboplan_ext": No such file
  or directory.

```

sea-bass: I thought I did put that in the conditional?

sea-bass: Oh oops I did not test the ROS case. Good catch

eholum: It's behind the ament conditional but I think that directory used to be created by the stubs generation? So cmake is unhappy.

sea-bass: I am curious now if this even needs conditionals... testing...

sea-bass: the answer is yes, it does. `${PYTHON_INSTALL_DIR}` is not defined without `colcon`

PR #137, opened 2026-02-15
Fix oink bindings

sjahr:
... shame :bell:


sea-bass (approved):
😆  -- we can add tests on the Python side for this reason too. I've been a bit lazy on that front myself, though.

PR #138, opened 2026-02-16
Use nanobind from pip install for ament workflows

eholum:
:/

Because of: https://github.com/open-planning/roboplan-ros/pull/26.

Apparently building these bindings with 2.9 and the ros bindings with 2.10 causes issues (bad casts).

In any case, one less network requirement at build time? I'm a little worried about maintaining consistency if/when we don't submodule this in the ros wrappers but I think that's a bridge will cross when we get there.

### DIFF
--- a/bindings/package.xml
+++ b/bindings/package.xml
@@ -11,8 +11,9 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
   <buildtool_depend>ament_cmake_python</buildtool_depend>
 
-  <!-- While nanobind-dev is available from rosdistro the apt package installs it in -->
-  <!-- inconsistent places. Users should ensure nanobind is just installed with pip. -->
+  <!-- While nanobind-dev is available from rosdistro the apt package is very out of date. -->
+  <!-- Nanobind should only be installed manually from pip until this is released in a sync: -->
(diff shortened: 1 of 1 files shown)


sea-bass:
For whatever reason, the humble job seems to be `pip install`ing `nanobind` as part of `rosdep install`, with version 2.11.0.

For jazzy onwards it's instead `apt install`ing `nanobind-dev`, with version 1.9.2.

bindings/package.xml
```diff
@@ -11,6 +11,9 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
   <buildtool_depend>ament_cmake_python</buildtool_depend>
 
+  <!-- While nanobind-dev is available from rosdistro the apt package installs it in -->
+  <!-- inconsistent places. Users should ensure nanobind is just installed with pip. -->
+  <!-- <depend>nanobind-dev</depend> -->
```
eholum, line 17: In case I ever get this idea again.

sea-bass:
Go for the green button!

I can get a 0.2.0 release out and on conda after this?

eholum:
https://github.com/ros/rosdistro/pull/49894

🫡🫡🫡

PR #139, opened 2026-02-18
Add octree support

### LINKED ISSUE #123: Add Octomap support
`coal` (the collision checking library attached to Pinocchio) supports Octrees as geometries. We should add support and an example for this!

CihatAltiparmak:
Hello, I've created this basic PR to support OcTree geometry as wanted at https://github.com/open-planning/roboplan/issues/123 .
I am yet to add test cases. However I'll add all these asap.

On the contrary, I should say that I've only work on ubuntu docker, not ros. I've tried to work on ros platform, but I neither could neither find any integrated support in precompiled hppfcl as to octomap methods (especially) nor install coal library. That's because it wants you to install upgraded numpy.

Secondly, I saw some warnings which indicates you should include coal instead of hppfcl on ubuntu docker.

Anyway, following picture shows basic visualization as to this PR.

Improving example config is on my plan as well.

How to run octree demo

```
python3 roboplan_examples/python/example_rrt.py --include-octrees
```

Screenshot from 2026-04-16 00-29-22


sea-bass:
Thanks for working on this!

I think the best way to check would be to create a unit test that uses the octree feature. CI should catch whether this works in all the 3 types of environments (ubuntu, ROS, and pixi)

bindings/src/roboplan/viser_visualizer.py
sea-bass, line 1: FYI once Pinocchio 3.9 is available on all ros distros, I will be deleting this file and using the ViserVisualizer implementation in Pinocchio itself.

bindings/src/roboplan/viser_visualizer.py
```diff
@@ -224,10 +224,82 @@ def loadViewerGeometryObject(self, geometry_object, prefix="", color=None):
...
                     opacity=color_override[3],
                 )
+        elif isinstance(geom, hppfcl.OcTree):
+            frame = self.createMeshFromOcTree(name, geom, color_override)
         else:
             raise RuntimeError(f"Unsupported geometry type for {name}: {type(geom)}")
 
         self.frames[name] = frame
+    
+    def createMeshFromOcTree(self, name, geom, color_override):
```
sea-bass: Can you check whether Viser already has a point cloud and/or octree visualization built in?

CihatAltiparmak: Yes, there is builtin pointcloud visualization [here](https://viser.studio/main/examples/scene/point_clouds/) If you wish, I can do it with pointcloud support. I am aware that this code block seems ugly.

sea-bass: It's less about how the code looks and more about if we could have just reused something.

I guess the long-term "right thing" would be to add an octree visualization function directly to Viser, but we can start with one here instead.

sea-bass:
The tests seem to pass in all CI jobs... while that's good, I am curious what may be happening on your local (non-Docker setup) to make the octree collision support fail.

sea-bass:
> Q1/ Should I wait until pinocchio 3.9 is released? (I can wait no problem.)

Yes, please! Right now all that is missing is for ROS 2 Kilted and Jazzy to release new syncs. You may feel free to rebase on my PR branch for now if you want to test using Pixi or ROS 2 Humble/Rolling.

> Q2/ What kind of demonstrations should I add under example_octree.py? For now, I've added only how to add octree into your scene etc. I need your suggestions here. Btw for a newbie, python implementation seems quite complex IMO.

I actually really liked the example that was contributed to PyRoboPlan (https://github.com/sea-bass/pyroboplan/pull/101#discussion_r2020160658), where a "realistic" scene was added with a voxel size that was large enough to make sense of it. Otherwise, I don't think the Python example looks that bad!

> Q3/ I thought so much about how I can use roboplan's `OcTree` class as `OcTree(octree_geom_of_hppfcl)`. (Shortly `OcTree(const std::shared_ptr<hppfcl::OcTree> & other_geom_ptr)`). Is it okay for you?
> 
> Unfortunately, I couldn't find a proper way to implement it due to the fact that the hppfcl released by pinocchio team prefers boost library in order to bind cpp functions to python. On the contrary roboplan relies on nanobind.

Right. Actually Pinocchio is planning a new major version (Pinocchio 4) that switches to nanobind, which is why I chose nanobind. Presumably hppfcl/coal will also follow suit for the bindings to work properly. This is why I made "wrapper" classes around the hppfcl Box/Sphere primitives in roboplan for now, and you should feel free to copy the same patterns as needed for now.

sea-bass:
@CihatAltiparmak ROS 2 jazzy finally had a sync, so I merged https://github.com/open-planning/roboplan/pull/97 and this should be good to rebase now!

You'll need your octree visualization to just be a new helper function in this repo now.

CihatAltiparmak:
Cool news, I've opened this feature request on pinocchio (https://github.com/stack-of-tasks/pinocchio/issues/2868) . Until this feature request is implemented. I need to inherit ViserVisualizer temporaryly. I will sync my branch with newest updates tonight and make it ready to merge.

sea-bass:
Great! Rather than inherit, I'd recommend just making a function that accepts a ViserVisualizer instance, but otherwise SGTM

Reference: https://github.com/open-planning/roboplan/blob/main/bindings/src/roboplan/visualization.py

bindings/src/roboplan/visualization.py
```diff
@@ -1,12 +1,104 @@
 from matplotlib.figure import Figure
 import matplotlib.pyplot as plt
 import numpy as np
+from pinocchio import pinocchio_pywrap_default as pin
```
sea-bass: Can this just be `import pinocchio as pin`?

bindings/test/test_scene.py
```diff
@@ -17,7 +17,6 @@
 )
 from roboplan.example_models import get_install_prefix
 
-
```
sea-bass, line 20: Restore?

roboplan/include/roboplan/core/geometry_wrappers.hpp
```diff
@@ -29,4 +30,27 @@ struct Sphere {
   std::shared_ptr<hpp::fcl::Sphere> geom_ptr;
 };
 
+struct OcTree {
+  OcTree(const std::vector<Eigen::Matrix<double, 6, 1>>& boxes, const double resolution) {
+    auto octree = std::make_shared<octomap::OcTree>(resolution);
+
+    if (!boxes.empty()) {
+      const double thresh = boxes[0][5];
```
sea-bass, line 38: Can you add comments explaining these indices? So 0 to 3 are position, 4 is log odds, 5 is ???

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -215,6 +215,10 @@ class Scene {
   addSphereGeometry(const std::string& name, const std::string& parent_frame, const Sphere& sphere,
                     const Eigen::Matrix4d& tform, const Eigen::Vector4d& color);
 
+  tl::expected<void, std::string>
```
sea-bass, line 225: Needs a docstring. Please add!

roboplan/test/test_scene.cpp
```diff
@@ -182,6 +182,96 @@ TEST_F(RoboPlanSceneTest, TestCollisionGeometry) {
...
+TEST_F(RoboPlanSceneTest, TestSetCollisionsForOcTree) {
+  // Nominally, this configuration is collision free.
+  Eigen::VectorXd q(6);
+  q << 0.0, 0.0, 0.0, 0.0, 0.0, 0.0;
+  ASSERT_FALSE(scene_->hasCollisions(q));
+
+  const auto color = Eigen::Vector4d(0.5, 0.5, 0.5, 0.5);
+  // Create octree with resolution 0.1 which generates Box(1.0, 1.0, 0.5)
+  const double resolution = 0.1;
+  Eigen::Matrix<double, 500, 3> point_cloud;
```
sea-bass: Rather than repeating this code, maybe consider making a helper function to create the test octree geometry object?

roboplan_examples/python/example_rrt.py
```diff
@@ -39,6 +40,7 @@ def main(
     port: str = "8000",
     rng_seed: int | None = None,
     include_obstacles: bool = False,
+    include_octrees: bool = False,
```
sea-bass, line 43: Document this param

roboplan_examples/python/example_scene.py
```diff
@@ -6,7 +6,6 @@
 from roboplan.core import JointConfiguration, Scene
 from roboplan.example_models import get_package_share_dir
 
-
```
sea-bass, line 9: Restore?

.pre-commit-config.yaml
```diff
@@ -39,7 +39,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/psf/black
-    rev: 25.12.0
+    rev: 26.1.0
```
sea-bass: Don't upgrade pre-commit stuff in this PR, maybe save for a follow up?

I should add the auto upgrade bot to this repo actually...

pixi.toml
```diff
@@ -226,6 +226,7 @@ matplotlib        = ">=3.10.5,<4"
 tyro              = ">=0.9.26,<0.10"
 viser             = ">=1.0.21,<2"
 zstandard         = ">=0.25.0,<0.26"
+plyfile           = "1.1"
```
sea-bass: Can this take in a range of versions instead?

Also got a warning:

```
 WARN Encountered ambiguous version specifier `1.1`, could be `1.1.*` but assuming you meant `==1.1`. In the future this will result in an error.

File: roboplan_example_models/CMakeLists.txt [hunk 11]
```diff
@@ -79,6 +79,7 @@ install(TARGETS roboplan_example_models
 install(
   DIRECTORY
   models
+  pointclouds
```
sea-bass: I think this would work better in the examples package.

Unless you're putting it here because of unit tests?

roboplan_examples/python/common.py
```diff
@@ -91,10 +129,14 @@ class RobotModelConfig:
     base_link: str
     starting_joint_config: List[float]
     obstacles: List[ObstacleConfig]
+    octrees: List[ObstacleConfig]
 
 
 # Base directory for all robot models
 ROBOPLAN_MODELS_DIR = get_package_models_dir()
+ROBOPLAN_POINCLOUDS_DIR = os.path.join(
```
sea-bass: Typo, should be `ROBOPLAN_POINTCLOUDS_DIR`

sea-bass: But also, should you instead use the models dir function?

roboplan_examples/python/common.py
```diff
@@ -105,6 +147,18 @@ class RobotModelConfig:
         ee_names=["tool0"],
         base_link="base",
         starting_joint_config=[0.0, -np.pi / 2, np.pi / 2, -np.pi / 2, -np.pi / 2, 0.0],
+        octrees=[
+            ObstacleConfig(
+                name="octree_cloud",
+                geom=load_point_cloud(
+                    os.path.join(ROBOPLAN_POINCLOUDS_DIR, "example_point_cloud.ply"),
+                    0.04,
```
sea-bass: Is this the resolution? Might be worth at least commenting that instead of using a literal

roboplan_examples/python/common.py
```diff
@@ -206,6 +272,18 @@ class RobotModelConfig:
             0.04,
             0.04,
         ],
+        octrees=[
+            ObstacleConfig(
+                name="octree_cloud",
+                geom=load_point_cloud(
+                    os.path.join(ROBOPLAN_POINCLOUDS_DIR, "example_point_cloud.ply"),
```
sea-bass: If the point cloud is the same for all robots, maybe just load it elsewhere instead of maki g so many copies?

Also why is one of them 0.04 and the others 0.4?

roboplan_examples/python/example_dual_arm.py
```diff
@@ -6,7 +6,6 @@
 from roboplan.core import Scene
 from roboplan.example_models import get_package_share_dir
 
-
```
sea-bass, line 9: Restore?

sea-bass (commented):
Thanks for rebasing this! Gave it a closer read and have a bunch of relatively minor comments.

pixi.toml
sea-bass: oh and update the lock file too!

roboplan_examples/python/example_rrt.py
```diff
@@ -105,6 +107,14 @@ def main(
     viz = ViserVisualizer(model, collision_model, visual_model)
     viz.initViewer(open=True, loadModel=True, host=host, port=port)
 
+    if include_octrees:
+        for obstacle in model_data.octrees:
+            obstacle.addToScene(scene)
+
+            geom_obj = obstacle.createGeometryObject(model)
+            visualizeOcTree(viz, geom_obj, viz.collisionRootNodeName, obstacle.color)
+            visualizeOcTree(viz, geom_obj, viz.visualRootNodeName, obstacle.color)
```
sea-bass: when I ran, I got an error here about number of arguments, I think because `obstacle.color` shouldn't be an argument anymore -- it worked for me when I removed it!

BTW, any reason why we're visualizing it both as visual and collision geometry?

bindings/src/roboplan/visualization.py
CihatAltiparmak: Solved in https://github.com/open-planning/roboplan/pull/139/commits/455adec7c73abf092eddaa7be0c2bd7306388222

roboplan/include/roboplan/core/scene.hpp
CihatAltiparmak: Solved in https://github.com/open-planning/roboplan/pull/139/commits/455adec7c73abf092eddaa7be0c2bd7306388222

roboplan_examples/python/example_rrt.py
CihatAltiparmak: Solved in https://github.com/open-planning/roboplan/pull/139/commits/455adec7c73abf092eddaa7be0c2bd7306388222

.pre-commit-config.yaml
CihatAltiparmak: I run `pre-commit run -a` command. It causes this unnecessary upgrade. You are right. Solved in https://github.com/open-planning/roboplan/pull/139/commits/455adec7c73abf092eddaa7be0c2bd7306388222

roboplan_examples/python/example_rrt.py
CihatAltiparmak: I've tried to follow the way pinocchio do it. pinocchio visualizes both collision and visual models. Thank you for giving a try. Solved in https://github.com/open-planning/roboplan/pull/139/commits/455adec7c73abf092eddaa7be0c2bd7306388222

bindings/test/test_scene.py
CihatAltiparmak: Solved in https://github.com/open-planning/roboplan/pull/139/commits/455adec7c73abf092eddaa7be0c2bd7306388222

roboplan/test/test_scene.cpp
CihatAltiparmak: Solved in https://github.com/open-planning/roboplan/pull/139/commits/455adec7c73abf092eddaa7be0c2bd7306388222. If you don't like it, please reopen it so I can follow.

roboplan_examples/python/example_dual_arm.py
CihatAltiparmak: Solved in https://github.com/open-planning/roboplan/commit/455adec7c73abf092eddaa7be0c2bd7306388222

roboplan/include/roboplan/core/geometry_wrappers.hpp
CihatAltiparmak: https://octomap.github.io/octomap/doc/classoctomap_1_1OccupancyOcTreeBase.html#a7c46e7320b83fa8a2272069d464c4e43 This is the definition of the method. I you don't like, please reopen

roboplan_examples/python/common.py
CihatAltiparmak: Yes, In that commit (https://github.com/open-planning/roboplan/pull/139/commits/d135d2f79235e8b44286a4397781eadf08769b3c), I've tried to apply suggestion and removed the redundant copies there.

roboplan_examples/python/common.py
CihatAltiparmak: My mistake, solved in https://github.com/open-planning/roboplan/pull/139/commits/d135d2f79235e8b44286a4397781eadf08769b3c

roboplan_examples/python/common.py
CihatAltiparmak: Solved there : https://github.com/open-planning/roboplan/pull/139/commits/d135d2f79235e8b44286a4397781eadf08769b3c

pixi.toml
CihatAltiparmak: Updated in https://github.com/open-planning/roboplan/pull/139/commits/a7509a583c87a3006cda82bda85a0dd9a0e539d2 . Tested with pixi and I could run example_rrt without addional step.

roboplan_example_models/CMakeLists.txt
CihatAltiparmak: I've thought that putting pointclouds on robot_example_models could be better at using it in common.py easier. are'nt pointclouds in the obstacle category as well?

CihatAltiparmak:
Hello, thank you for review and sorry for my spamming meesages. I've sorted out other reviews. The only one left. Would you take a look at my question at that review when you are available?

roboplan_example_models/CMakeLists.txt
sea-bass: Not exactly. The URDFs and meshes are used by unit tests in most other packages, so that's why they are in the a common example models package.

The pointcloud .ply files are used only in the `roboplan_examples` package, so there is an argument for only including them there. Although it will not necessarily be robust this way to different current folders and dev workflows

Indeed, it's true that we (mostly @eholum) worked very hard on creating those example models functions that work across ROS, regular cmake, and conda/pixi workflows.

Let's keep it as is, because it will anyway be useful to have one real point cloud file handy for other unit testing.

sea-bass (approved):
LGTM, thank you for working on this!

One last favor: can you update the screenshot in the PR description to show the new octree that looks way better? Just for others to see the final version of the example.

CihatAltiparmak:
I updated the description of PR.

PR #144, opened 2026-02-21
Fix stub install folder

sea-bass:
was off by one folder... ugh cmake and its slashes


bindings/src/roboplan/roboplan_ext/optimal_ik.pyi
```diff
@@ -170,16 +170,14 @@ class Oink:
...
+                            task tracking accuracy. Default: 1e-12.
 
         Raises:
             RuntimeError: If the QP solver fails to find a solution.
 
         Example:
             delta_q = np.zeros(oink.num_variables)
             oink.solveIk(tasks, constraints, scene, delta_q)
-            # Or with custom regularization:
-            oink.solveIk(tasks, constraints, scene, delta_q, regularization=1e-6)
```
sea-bass, line 184: incidental fix because once again these were not updated. Should really have a CI check

PR #145, opened 2026-02-21
Ensure oink tests run in all workflows

sea-bass:
These were only running in the ROS CI jobs.

Pixi one was easy to fix; we just needed to actually add the package to the relevant tasks in the `toml` file.

For the Ubuntu container, there was some weird subtlety with the case of fetching OSQP libraries from github. Something was unsetting the `BUILD_TESTING` option in there, so I had to make a side option to ensure oink tests built in this case. Open to alternatives if anyone has them!

### DIFF
--- a/roboplan_oink/CMakeLists.txt
+++ b/roboplan_oink/CMakeLists.txt
@@ -15,12 +15,31 @@ include(FetchContent)
 # Find dependencies.
 find_package(roboplan REQUIRED)
 
+# Ament CMake is not immediately required, but we include it at the top so that
+# if it is found we can use it in build testing before requiring a
+# find_package(ament_cmake_test REQUIRED). Otherwise this can cause infinite recursion
+# issues with symlink installs.
+find_package(ament_cmake QUIET)
+
 # Try to find OsqpEigen
 find_package(OsqpEigen QUIET)
 
 if(NOT OsqpEigen_FOUND)
-  # When building OsqpEigen from source, we need OSQP >= 1.0.0
+  # When building OsqpEigen from source, we need OSQP >= 1.0.0.
   find_package(osqp QUIET)
+
+  # ROS osqp vendor package can set osqp_FOUND without osqp_VERSION.
+  # Disable follow-up osqp package lookup so OsqpEigen uses the vendored OSQP.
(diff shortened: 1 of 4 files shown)


scripts/build_cmake.bash
```diff
@@ -8,13 +8,6 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 REPO_ROOT_DIR="${SCRIPT_DIR}/.."
 pushd ${REPO_ROOT_DIR} || exit
 
-# TODO: First install Pinocchio via https://stack-of-tasks.github.io/pinocchio/download.html
-# This whole block should probably just be in a non-ROS Dockerfile.
-export PATH=/opt/openrobots/bin:$PATH
-export PKG_CONFIG_PATH=/opt/openrobots/lib/pkgconfig:$PKG_CONFIG_PATH
-export LD_LIBRARY_PATH=/opt/openrobots/lib:$LD_LIBRARY_PATH
-export CMAKE_PREFIX_PATH=/opt/openrobots:$CMAKE_PREFIX_PATH
-
```
sea-bass, line 17: no longer needed

roboplan_oink/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+# Set a new oink specific build testing option since osqp can mess with the original.
+option(BUILD_TESTING "" ON)
```
sjahr, line 9: What's the value in keeping `BUILD_TESTING`?

sjahr: & would it be worth raising an issue in osqp about this?

sea-bass: BUILD_TESTING is still how you enable/disable testing in all packages (including ROS package, I believe colcon uses this flag as well), so it's pretty standard. Hence why it's good to keep it as the main entrypoint.

sea-bass: Re: raising issue in osqp, it may help, but I think in general the original issue is simply that we (or nobody, really) understand CMake well enough to have it do our bidding lol

PR #147, opened 2026-02-21
Fix position limits oink constraint for models with continuous joints

sea-bass:
Oink was going unstable with the Kinova model because using the Pinocchio model's joint limits fields actually considers continuous joints as `[cos(theta), sin(theta)]`, meaning there are 2 elements both with limits +/- 1.0.

This properly uses the RoboPlan `Scene`'s joint limits info, as well as the `collapseContinuousJointPositions()` utility, to handle continuous joints. Kinova example is now numerically stable by using correct position limits.

[Screencast from 2026-02-21 15-25-17.webm](https://github.com/user-attachments/assets/70a69e18-3940-4508-8544-97df17d43067)


roboplan_oink/src/constraints/position_limit.cpp
```diff
@@ -16,6 +17,50 @@ tl::expected<void, std::string> PositionLimit::computeQpConstraints(
...
   const auto& q = scene.getCurrentJointPositions();
 
+  auto maybe_q_collapsed = collapseContinuousJointPositions(scene, "", q);
+  if (!maybe_q_collapsed) {
+    return tl::make_unexpected("Failed to compute position constraint: " +
+                               maybe_q_collapsed.error());
+  }
+  const auto& q_collapsed = maybe_q_collapsed.value();
+
+  // Get joint limits from the model (only do this once)
```
sea-bass, line 27: This highlights a need to possibly have tasks accept the scene in their constructor so that information can be pre-allocated at construction time.

The same thing happens with `FrameTask` re: getting the `frame_id` value for the joint.

Although, I do like the solver doing its own thing so that the scene usage is consistent. Maybe then the best approach is to have a `setup()` method in these tasks/constraints that is automatically called by the solver? Or at least we can just call `solver.setup()` and that invokes it for all the tasks/constraints/barriers?

bindings/CMakeLists.txt
```diff
@@ -193,7 +193,7 @@ if(GENERATE_PYTHON_STUBS)
       set(PYTHON_STUB_INSTALL_DIR "roboplan")
   endif()
   install(
-    DIRECTORY "${CMAKE_BINARY_DIR}/roboplan_ext/"
+    DIRECTORY "${CMAKE_BINARY_DIR}/roboplan_ext"
```
sea-bass, line 196: fixed in https://github.com/open-planning/roboplan/pull/144 but not yet merged

roboplan_oink/src/constraints/position_limit.cpp
sjahr: Good point! I opened an issue to have to track this as additional work. https://github.com/open-planning/roboplan/issues/152

roboplan_oink/src/constraints/position_limit.cpp
```diff
@@ -16,6 +17,50 @@ tl::expected<void, std::string> PositionLimit::computeQpConstraints(
...
+  if (q_min.size() == 0u) {
+    q_min.resize(num_variables);
```
sjahr, line 29: How about adding this as function to the scene class or scene_utils.cpp? I can see this being useful outside of this constraint.

roboplan_oink/src/constraints/position_limit.cpp
```diff
@@ -16,6 +17,50 @@ tl::expected<void, std::string> PositionLimit::computeQpConstraints(
...
+    q_max.resize(num_variables);
+    const auto joint_names = scene.getJointNames();
+    for (int idx = 0; idx < num_variables; ++idx) {
+      const auto& joint_name = joint_names.at(idx);
+      const auto maybe_joint_info = scene.getJointInfo(joint_name);
+      if (!maybe_joint_info) {
+        throw std::runtime_error("Failed to get joint limits for position constraint: " +
```
sjahr: Should we throw or just return tl::make_unexpected to let the caller handle this?

roboplan_oink/src/constraints/position_limit.cpp
```diff
@@ -16,6 +17,50 @@ tl::expected<void, std::string> PositionLimit::computeQpConstraints(
...
+                               maybe_q_collapsed.error());
+  }
+  const auto& q_collapsed = maybe_q_collapsed.value();
+
+  // Get joint limits from the model (only do this once)
+  if (q_min.size() == 0u) {
+    q_min.resize(num_variables);
+    q_max.resize(num_variables);
+    const auto joint_names = scene.getJointNames();
+    for (int idx = 0; idx < num_variables; ++idx) {
```
sjahr, line 32: Nitpick: Would this work: `for joint_name : joint_names`

roboplan_oink/src/constraints/position_limit.cpp
sea-bass: oh yeah, didn't realize this returned an expected!

roboplan_oink/src/constraints/position_limit.cpp
sea-bass: no because we use `idx` to assign to `q_min` and `q_max` too

roboplan_oink/src/constraints/position_limit.cpp
sea-bass: I will make a separate ticket for this.

roboplan_oink/src/constraints/position_limit.cpp
sea-bass: I already captured that in #148 I think!

sjahr:
@sea-bass Looks like the last commit broke something
```
[OsqpEigen::Settings::setTimeLimit] OSPQ has been set without PROFILING, hence this setting is disabled.
ERROR in validate_data: Lower bound at index 21 is greater than upper bound: 5.0000e-03 > -5.0000e-03
ERROR in osqp_setup: Problem data validation.
[OsqpEigen::Solver::initSolver] Unable to setup the workspace.
Warning: IK solver failed: IK solve failed: Failed to initialize solver, using zero delta_q
```

I just quickly tested it and got this error. Does it happen to you as well? The commit before seems to work. Do you mind taking a look into this? Not sure why the test isn't catching it.

sea-bass:
> Ah I think I know what this is. The max velocity/accel/jerk has to be multiplied by the _absolute value_ of the scaling, so in the negative case it gets messed up.
> 
> Will fix and test when home.

This was indeed it. Fixed, thank you!

PR #150, opened 2026-02-22
Robot Learning Application Examples: Adding teleop functionality via keyboard

Theonewhomadethings:
## Test 

`python roboplan_examples/python/example_teleop.py --model ur5 --teleop keyboard`

`python roboplan_examples/python/example_teleop.py --model franka --teleop keyboard`

## Keyboard controls
Click/focus the running terminal window first.

Translation movement:
- `w` / `s` ->  x-direction movement
- `a` / `d` ->  y-direction movement
- `q` / `e` ->  z-direction movement

Rotational movement:
- `i` / `k` ->  roll direction
- `j` / `l` ->  pitch direction
- `u` / `o` ->  yaw direction

Other keys:
- `space` -> pause/resume teleop
- `r` -> reset robot to home position
- `t` -> reset target to current end-effector pose
- `x` -> quit teleop

## GUI controls
- `Linear Sensitivity` slider
- `Angular Sensitivity` slider
- `Control Frame` dropdown (`ee`/`world`)
- `Reset to Home` button

## quick demo

https://github.com/user-attachments/assets/9b91c7ef-c49d-430c-a3c4-8053ea5b1c85


roboplan_examples/python/teleop/command_filter.py
```diff
@@ -0,0 +1,52 @@
+from __future__ import annotations
```
sea-bass: Is this import necessary?

roboplan_examples/python/teleop/config.py
```diff
@@ -0,0 +1,46 @@
...
+    rz_neg: str = "o"
+    gripper_open: str = "g"
+    gripper_close: str = "h"
+    pause_toggle: str = "space"
+    reset_home: str = "r"
+    reset_target_to_current: str = "t"
+
+
+@dataclass
+class TeleopConfig:
```
sea-bass, line 31: Make sure units are in the comments

sea-bass (commented):
Thanks for taking this one! Just gave it an initial cursory look for now and providing high level comments.

- As personal preference, I would favor shorter, less LLMy text for the issues, PR descriptions, and commit messages. Most of the explanations can and should go in docstrings in the code itself so they can be modified later and accompany changes to the code.
- What is the difference between the regular and the `m0` teleop script? I would just keep one of those.
- It's really cool seeing all those viser hooks, but we shouldn't go too far with those yet. I do like the sliders for controlling certain tunable parameters though!
- Recall that this is a C++ library with Python bindings. While I think overall you chose correctly here, in that this teleop stuff should be mostly Python as it's mostly example code and not core library. But, as you flesh things out do consider whether any of this should also be made available to people developing purely in C++.

Also sorry for the generated stubs noise. This is fixed in https://github.com/open-planning/roboplan/pull/144, hasn't been reviewed but if this works for you and you can confirm, let me know!

roboplan_examples/python/teleop/command_filter.py
Theonewhomadethings: addressed in commit https://github.com/open-planning/roboplan/pull/150/changes/a1688d98bc65c6fe3fbc547e7810d8e177836325

roboplan_examples/python/teleop/config.py
Theonewhomadethings: addressed in commit https://github.com/open-planning/roboplan/pull/150/changes/afc462e4cebab299d050b5debc900b6e18f61354

bindings/src/roboplan/roboplan_ext/README.md
```diff
@@ -1,5 +0,0 @@
-This folder is generated automatically by building the bindings.
```
sjahr, line 1: Is this removed intentionally? Maybe I missed that in the conversation

bindings/CMakeLists.txt
```diff
@@ -193,7 +193,7 @@ if(GENERATE_PYTHON_STUBS)
       set(PYTHON_STUB_INSTALL_DIR "roboplan")
   endif()
   install(
-    DIRECTORY "${CMAKE_BINARY_DIR}/roboplan_ext/"
+    DIRECTORY "${CMAKE_BINARY_DIR}/roboplan_ext"
```
sjahr, line 196: Same here

roboplan_examples/python/teleop/command_filter.py
```diff
@@ -0,0 +1,50 @@
+import numpy as np
+
+from .config import TeleopConfig
+from .types import TwistCommand
+
+
+class CommandFilter:
```
sjahr, line 7: I am confused, isn't that a re-implementation of the command filter you created in the library?

roboplan_oink/CMakeLists.txt
```diff
@@ -76,6 +76,8 @@ add_library(roboplan_oink SHARED
     src/tasks/configuration.cpp
     src/constraints/position_limit.cpp
     src/constraints/velocity_limit.cpp
+    src/teleop/command_filter.cpp
```
sjahr, line 79: I wouldn't add this to the oink package, since it is not necessarily related to oink. For example, you could also use Simple IK for the steering. Either, you move this to it's own package, or don't have any C++ code at all since this is only meant to be used as a python example.

sjahr (commented):
Thank you for you contribution! A couple of thoughts:
- As mentioned below, I think this should be unrelated of the oink package
- Would it be possible to condense the example to fewer files to simplify the it a bit.

sea-bass:
Closing since it was superseded by #213

PR #177, opened 2026-04-12
Incorporate scene and joint groups into OInK

### LINKED ISSUE #146: Oink solver should accept joint groups
Right now the oink solver uses the entire model, meaning all its actuated joints. For robots with lots of gripper joints, like the Kinova, this is not ideal.

A solution would be for the Oink solver itself to accept a group name and be able to down-select its number of variables based on the joints in that group.

### LINKED ISSUE #148: Oink tasks, constraints, and barriers should have a "setup" step that uses the scene
See https://github.com/open-planning/roboplan/pull/147#discussion_r2836579742

As a bonus, this could also serve as a way to automatically determine `num_variables` instead of having to pass them in.

sea-bass:
This PR changes the interfaces of OInK to improve various aspects of developer quality of life:

1. Scene information is passed in at construction time so we can precompute things at constructio time
2. OInK solver can operate on joint groups, so you can remove redundant degrees of freedom from solves and greatly improve OSQP matrix sizes and thereby solve time
3. Number of variables are all managed internally as a consequence.

This is a sweeping and breaking change that will very much affect older versions, so we should get this in for 0.3.0.

The key design change is that:

* The `Oink` solver accepts a scene and a group name -- this defines the dimensionality of the tangent space (i.e., which subset of `model.nv` to use)
* Each task/constraint/barrier now accepts an `Oink` instance in its constructor (and in some cases, also a scene when needed), which allows passing of all the necessary information from the solver itself.

It's not perfect, but it works! I was able to run all the oink examples with all our example robots. Looking forward to some feedback.

Closes #146 
Closes #148 
Closes #152


bindings/src/modules/core.cpp
```diff
@@ -183,12 +183,9 @@ void init_core_scene(nanobind::module_& m) {
            "Converts partial joint positions to full joint positions.", "group_name"_a, "q"_a)
       .def("interpolate", &Scene::interpolate, "Interpolates between two joint configurations.",
            "q_start"_a, "q_end"_a, "fraction"_a)
-      .def(
-          "integrate",
-          [](const Scene& self, const Eigen::VectorXd& q, const Eigen::VectorXd& v)
-              -> Eigen::VectorXd { return pinocchio::integrate(self.getModel(), q, v); },
-          "Integrates a velocity vector from a configuration using Lie group operations.", "q"_a,
-          "v"_a)
```
sea-bass, line 194: Turned this into a proper method of the C++ scene instead of a Python bindings exclusive

roboplan_examples/python/example_oink.py
```diff
@@ -100,30 +97,26 @@ def main(
     viz = ViserVisualizer(model_pin, collision_model, visual_model)
     viz.initViewer(open=True, loadModel=True, host=host, port=port)
 
-    # Determine the velocity space dimension (nv) by computing a Jacobian
-    # For models with quaternion joints, nv < nq
-    jac = scene.computeFrameJacobian(q_full, model_data.ee_names[0])
-    num_variables = jac.shape[1]  # nv (velocity space dimension)
```
sea-bass, line 106: i am so happy to get rid of this block of code, since this was a very awkward way of getting the dimensionality of our problem

roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp
```diff
@@ -114,15 +116,15 @@ struct PositionBarrier : public Barrier {
...
 
-private:
-  /// @brief Cached frame index for fast lookups.
-  mutable pinocchio::FrameIndex frame_id = 0;
+  /// @brief Velocity indices of the joint group (for Jacobian column selection).
+  Eigen::VectorXi v_indices;
 
-  /// @brief Whether frame_id has been cached.
-  mutable bool frame_id_cached = false;
+private:
```
sjahr: It's a struct, I think `private` can be removed or we need to make it a class

roboplan_oink/src/constraints/velocity_limit.cpp
```diff
@@ -23,14 +25,7 @@ int VelocityLimit::getNumConstraints(const Scene& /*scene*/) const { return num_
...
     Eigen::Ref<Eigen::VectorXd> lower_bounds, Eigen::Ref<Eigen::VectorXd> upper_bounds) const {
-  const auto& model = scene.getModel();
-
-  // Validate that model dimensions match constructor
-  if (model.nv != num_variables) {
-    return tl::make_unexpected("VelocityLimit: model.nv (" + std::to_string(model.nv) +
-                               ") does not match num_variables (" + std::to_string(num_variables) +
-                               ") from constructor");
-  }
+  (void)scene;
```
sjahr: What does this do?

roboplan/src/core/scene.cpp
```diff
@@ -371,6 +371,10 @@ Eigen::VectorXd Scene::interpolate(const Eigen::VectorXd& q_start, const Eigen::
   return pinocchio::interpolate(model_, q_start, q_end, fraction);
 }
 
+Eigen::VectorXd Scene::integrate(const Eigen::VectorXd& q, const Eigen::VectorXd& v) const {
+  return pinocchio::integrate(model_, q, v);
```
sjahr, line 375: Does pinocchio handles dimension mismatches here gracefully or do we need guardrails for the inputs or need to catch something / return tl::expected?

roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp
```diff
@@ -114,15 +116,15 @@ struct PositionBarrier : public Barrier {
...
+  /// @brief Frame index (resolved eagerly at construction).
+  pinocchio::FrameIndex frame_id;
 
-  /// @brief Pre-allocated workspace for frame Jacobian (6 x nv).
-  mutable Eigen::MatrixXd frame_jacobian;
+  /// @brief Pre-allocated full-robot Jacobian workspace (6 x model.nv).
+  mutable Eigen::MatrixXd full_jacobian_;
```
sjahr: If we remove the private, the _ needs to be removed. Also, I liked frame jacobian. Any benefits in renaming it?

roboplan_oink/src/tasks/frame.cpp
```diff
@@ -81,25 +83,17 @@ tl::expected<void, std::string> FrameTask::computeError(const Scene& scene) {
...
-                             jacobian_container);
+  // Compute full-robot frame Jacobian, then select the group's velocity columns
+  full_jacobian_.setZero();
+  scene.computeFrameJacobian(q, frame_id, pinocchio::ReferenceFrame::LOCAL_WORLD_ALIGNED,
+                             full_jacobian_);
 
   // The negative sign ensures that with the QP formulation (min ||J*dq + gain*e||^2),
   // the solution dq = -gain * J^{-1} * e moves toward the target.
-  jacobian_container *= -1.0;
+  jacobian_container.noalias() = -full_jacobian_(Eigen::placeholders::all, v_indices);
```
sjahr: Is this more performant?

sjahr (commented):
Thanks for these changes! I read through them and still need to test the PR but here is a first set of comments/ minor questions.

One design decision I need to think more about is whether the scene should be a member of the oink struct or not. What I like about Pinocchio is the clean separation of state and model and my idea was to somewhat adapt this for OInK and not store robot state information in the oink struct. I like that a user needs to consciously pass the scene with the current state to the solveIK call. With the current change we loose this and when you call scene.integrate later in the code this also changes the internal state of oink itself which might not be obvious. However I can also see the benefit in having everything in one place aka the OInK struct.
What are your thoughts on this?

sea-bass:
> One design decision I need to think more about is whether the scene should be a member of the oink struct or not. What I like about Pinocchio is the clean separation of state and model

Agreed. There are probably other ways of doing this. In a previous version of the branch, I was instead passing in the scene and the group name into the constructor, but at some point I was like "meh, oink already has all this info". Let me think how it could look better

roboplan_oink/src/tasks/frame.cpp
sea-bass: The `noalias()` stuff, yes. It prevents creating a temporary matrix.

The indexing definitely is more performant. A big point of this PR is to specify joint groups so you don't allocate matrices the size of the full robot state if you're not using them.

roboplan_oink/include/roboplan_oink/barriers/position_barrier.hpp
sea-bass: I think it was auto-renamed because our frame jacobian is now a "slice" of the full Jacobian matrix. But I can prob clean this up

roboplan/src/core/scene.cpp
sea-bass: Would be a good idea to validate. Will do.

roboplan_oink/src/constraints/velocity_limit.cpp
sea-bass: I... don't know. I blame Claude for this one and will fix it.

sea-bass: oh this was an unused variable that instead of being removed was cast to void... yeet!

roboplan/src/core/scene.cpp
sea-bass: actually on second thought I'm gonna leave this for now.

sjahr (approved):
Nice, I like this better 👍 Thank you!

PR #180, opened 2026-04-18
Add SE3 low-pass filter C++ library and Python bindings

crisiumnih:
Refers #173 
- Adds a new `roboplan_filters` C++ package.
- Implements `SE3LowPassFilter` for first-order filtering of SE3 poses.
- Adds Python bindings under `roboplan.filters`.
- Adds C++ and Python tests for the filter.


roboplan/include/roboplan/filters/se3_low_pass_filter.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+  /// @brief Checks whether the filter has an active filtered state.
+  /// @return True if the filter has been initialized by reset() or update().
+  bool isInitialized() const;
+
+private:
+  double tau_;
+  bool initialized_;
+  Eigen::Vector3d filtered_position_;
+  Eigen::Quaterniond filtered_quaternion_;
```
sea-bass, line 48: could you add docstrings to these too?

sea-bass (commented):
Overall looks good! Two high level suggestions:

1. Rather than making a new `roboplan_filters` package, maybe just include these as a subfolder in the existing `roboplan` package as this seems like it belongs in the core. Just make sure it's a separate cmake target so you can have e.g. `roboplan::roboplan` and `roboplan::filters` both exported.
2. Ensure the IK examples are using these new bindings, and delete the old ones that exist on the pure Python side too!

crisiumnih:
> 1. Rather than making a new `roboplan_filters` package, maybe just include these as a subfolder in the existing `roboplan` package as this seems like it belongs in the core. Just make sure it's a separate cmake target so you can have e.g. `roboplan::roboplan` and `roboplan::filters` both exported.

Ok got it! That makes more sense

roboplan/test/test_se3_low_pass_filter.cpp
```diff
@@ -0,0 +1,87 @@
...
+}  // namespace
+
+TEST(SE3LowPassFilterTest, ConstructorDefaults) {
+  const SE3LowPassFilter filter;
+
+  EXPECT_DOUBLE_EQ(filter.tau(), 0.1);
+  EXPECT_FALSE(filter.isInitialized());
+}
+
+TEST(SE3LowPassFilterTest, SetTau) {
```
sjahr: I think we dont need setter and getter tests

roboplan/test/test_se3_low_pass_filter.cpp
```diff
@@ -0,0 +1,87 @@
...
+pinocchio::SE3 makePose(const Eigen::Vector3d& translation, const Eigen::Quaterniond& quaternion) {
+  return pinocchio::SE3(quaternion.normalized().toRotationMatrix(), translation);
+}
+
+}  // namespace
+
+TEST(SE3LowPassFilterTest, ConstructorDefaults) {
+  const SE3LowPassFilter filter;
+
+  EXPECT_DOUBLE_EQ(filter.tau(), 0.1);
```
sjahr: This check can be removed so we dont need to update this test when the default value changes

sjahr (approved):
Thanks for addressing this!
Just a minor non-blocking comment, but otherwise this looks good 😄

roboplan/test/test_se3_low_pass_filter.cpp
crisiumnih: okays

roboplan/src/filters/se3_low_pass_filter.cpp
```diff
@@ -0,0 +1,43 @@
...
+  if (!initialized_) {
+    reset(target_pose);
+    return target_pose;
+  }
+
+  const double alpha = 1.0 - std::exp(-dt / tau_);
+
+  filtered_position_ = filtered_position_ + alpha * (target_position - filtered_position_);
+  filtered_quaternion_ = filtered_quaternion_.slerp(alpha, target_quaternion);
+  filtered_quaternion_.normalize();
```
sea-bass: given that the initialized and target quaternions are already normalized, do we need to re-normalize after the `slerp`?

sea-bass (approved):
Just tested it and it works great -- thanks!

One optional comment/question remaining

roboplan/src/filters/se3_low_pass_filter.cpp
crisiumnih: tbh im confused too, i probably missed this my bad

PR #182, opened 2026-04-19
Add clang-tidy

sjahr:
Adds clang-tidy for better code quality.
~~@sea-bass @eholum do you have any thoughts on the following items:~~
- ~~Do you like the checks/ are any missing or should we remove some?~~
- ~~Should we include it via pre-commit or a dedicated CI job?~~
~~Personally I like having everything running in pre-commit but the clang-tidy job increases the execution time of `pre-commit run -a` so I can see why this is annoying. To me it would be acceptable though.~~
- ~~Are you happy with the naming conventions or should change any?~~

~~There are still some things I need to address and iron out but I wanted to hear your thoughts on these questions above while wrapping the last remaining things up.~~

Clang tidy can now be run as a pixi task in to flavors: with and without the fix flag, I also added the clang-tidy check as a CI stage to the build and test job. Thanks for your feedback!


.pre-commit-config.yaml
```diff
@@ -56,6 +56,13 @@ repos:
       - id: clang-format
         files: \.(c|cpp|h|hpp)$
 
+  - repo: https://github.com/pocc/pre-commit-hooks
+    rev: v1.3.5
+    hooks:
+      - id: clang-tidy
+        args: ["-p=build/roboplan"]
```
sea-bass: What is this path for? And does it only work in the pixi case?

.pre-commit-config.yaml
```diff
@@ -56,6 +56,13 @@ repos:
...
+        files: \.(c|cpp)$
```
sea-bass: Do you also want header files included?

pixi.toml
```diff
@@ -204,14 +203,14 @@ gcc         = ">=11,<12"
 lld         = ">=19.1.2,<20"
 libstdcxx   = ">=15.1.0,<16"
 llvm-openmp = ">=19.1.7,<20"
+clang-tools = ">=19.1.7,<20"
```
sea-bass: giga nit: can this be alphabetized

sea-bass (commented):
Overall looks great! Seems it caught a few things already. There are some conflicts but otherwise we should go for it and can always tweak later.

My one major question is whether this pre-commit check works with and without pixi?

.pre-commit-config.yaml
sjahr: Oh, I need to update that thanks! I played a bit around to check what influences the check time and removed it here.

.pre-commit-config.yaml
sjahr: Yes, this is what I still try to figure out. Right now it only works with the pixi case but I am on it

sea-bass: It's fine if it works only with the pixi case (our nanobind generated stub check is the same), as long as it doesn't make pre-commit fail otherwise, but rather acts as a "no-op"

eholum: I would probably vote to just have this as a pixi env + build target rather than in pre-commit? Might make more sense so that you don't have to build anything when running pre-commit and the deps stay lightweight and fast.

pixi.toml
eholum: I think we could add a linting target here or something, maybe that depends on the build?

```
# there's def going to be more magic needed...
clang-tidy = { cmd = "run-clang-tidy -p build/roboplan" ...}
```

Then we could add the check to the pixi build in CI?

eholum (commented):
yay for clang-tidy! is it super common to have it in pre-commit vs a build step (I don't know the answer I just know I like pre-commit to be mostly raw format checks...).

.github/workflows/build_and_test.yml
```diff
@@ -137,6 +137,9 @@ jobs:
           pixi run -e ci install_all
           pixi run -e ci build_tests
           pixi run -e ci test_all
+      - name: Run clang-tidy
+        run: |
+          pixi run clang-tidy
```
sjahr: @eholum Do I need the -e ci flag here too?

sea-bass: Actually I added this. The only thing this does is use copy install vs symlink, to check the nanobind generated stub file diffs

I would add it just because otherwise it might break caches from only having built in the ci environment.

roboplan_examples/cpp/example_oink.cpp
```diff
@@ -49,7 +49,7 @@ int main(int /*argc*/, char* /*argv*/[]) {
   const auto num_variables = oink.v_indices.size();
 
   // Create a FrameTask (high priority)
-  FrameTaskOptions frame_options{
+  FrameTaskOptions const frame_options{
```
sea-bass: Weird that the linter chose east const here. Can we move it left?

roboplan_examples/cpp/example_oink.cpp
```diff
@@ -59,20 +59,20 @@ int main(int /*argc*/, char* /*argv*/[]) {
 
   // Create a ConfigurationTask to regularize toward start configuration (low priority)
   // Using lower joint_weights (0.1) makes this task less important than the frame task
-  Eigen::VectorXd joint_weights = Eigen::VectorXd::Constant(num_variables, 0.1);
-  ConfigurationTaskOptions config_options{
+  Eigen::VectorXd const joint_weights = Eigen::VectorXd::Constant(num_variables, 0.1);
```
sea-bass: Same here and going forward

roboplan_examples/cpp/example_rrt.cpp
```diff
@@ -34,7 +34,7 @@ int main(int /*argc*/, char* /*argv*/[]) {
   // Pick random initial and goal configurations and plan motion
   const auto q_full_opt = scene->randomCollisionFreePositions();
   if (!q_full_opt) {
-    std::cout << "Failed to generate random collision free positions" << std::endl;
+    std::cout << "Failed to generate random collision free positions" << '\n';
```
sea-bass: If this is the case then you can omit the << and just put the newline in the previous string

roboplan_examples/cpp/example_rrt.cpp
```diff
@@ -48,34 +48,34 @@ int main(int /*argc*/, char* /*argv*/[]) {
   JointConfiguration goal;
   goal.positions = scene->randomCollisionFreePositions().value()(q_indices);
 
-  std::cout << "Planning..." << std::endl;
+  std::cout << "Planning..." << '\n';
```
sea-bass: Same here and going forward

roboplan_oink/include/roboplan_oink/optimal_ik.hpp
```diff
@@ -367,7 +367,7 @@ struct Oink {
   ///
   /// @note Only barriers that implement evaluateAtConfiguration() are checked.
   ///       Barriers returning infinity are assumed safe.
-  tl::expected<void, std::string>
+  static tl::expected<void, std::string>
```
sea-bass: Why is this static?

roboplan_oink/src/tasks/configuration.cpp
```diff
@@ -1,14 +1,15 @@
 #include <roboplan_oink/tasks/configuration.hpp>
 
 #include <pinocchio/algorithm/joint-configuration.hpp>
+#include <utility>
```
sea-bass: Nit: can we put system headers further up with the other ones? I believe I saw this in another instance further up in the PR but didnt comment

roboplan_oink/src/tasks/configuration.cpp
```diff
@@ -66,13 +67,13 @@ tl::expected<void, std::string> ConfigurationTask::computeJacobian(const Scene&
 
 Eigen::MatrixXd ConfigurationTask::createWeightMatrix(const Eigen::VectorXd& joint_weights) {
   const int nv = static_cast<int>(joint_weights.size());
-  Eigen::MatrixXd W = Eigen::MatrixXd::Zero(nv, nv);
+  Eigen::MatrixXd w = Eigen::MatrixXd::Zero(nv, nv);
```
sea-bass: You may not want clang-tidy to turn matrices variables into lowercase.


This one may be worth looking at the rules and disabling all throughout.

roboplan_oink/src/tasks/frame.cpp
```diff
@@ -49,19 +49,19 @@ tl::expected<void, std::string> FrameTask::computeError(const Scene& scene) {
   const pinocchio::SE3 transform_world_to_target(target_pose.tform);
 
   // Compute linear and angular errors from target to frame, in the world frame.
-  Eigen::Vector3d e_pos =
+  Eigen::Vector3d const e_pos =
```
sea-bass: It did east const here again

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -243,7 +243,7 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
   constraint_sizes.reserve(constraints.size());
   int total_constraint_rows = 0;
   for (const auto& constraint : constraints) {
-    int num_rows = constraint->getNumConstraints(scene);
+    int const num_rows = constraint->getNumConstraints(scene);
```
sea-bass: More east const

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -340,21 +340,21 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
 
     const OSQPSettings* stored_settings = settings.getSettings();
 #ifdef OSQP_EIGEN_OSQP_IS_V1
-    solver.settings()->setWarmStart(stored_settings->warm_starting);
-    solver.settings()->setPolish(stored_settings->polishing);
+    solver.settings()->setWarmStart(stored_settings->warm_starting != 0);
```
sea-bass: For these, is there a way to change the types or use static cast to bool instead? Looks fishy

roboplan_oink/test/barriers/test_position_barrier.cpp
```diff
@@ -49,14 +49,14 @@ class PositionBarrierTest : public ::testing::Test {
...
-  double dt_;
+  int num_variables_{};
+  double dt_{};
 };
 
 // Test basic construction with full box
 TEST_F(PositionBarrierTest, ConstructionFullBox) {
-  Eigen::Vector3d p_min(-1.0, -1.0, 0.0);
-  Eigen::Vector3d p_max(1.0, 1.0, 2.0);
+  Eigen::Vector3d const p_min(-1.0, -1.0, 0.0);
```
sea-bass: Return of the east const. Gonna stop saying this now

roboplan_oink/test/barriers/test_position_barrier.cpp
```diff
@@ -214,18 +214,18 @@ TEST_F(PositionBarrierTest, BarrierAllowsSafeMotion) {
...
-  std::vector<std::shared_ptr<Barrier>> barriers = {barrier};
+  std::vector<std::shared_ptr<Task>> const tasks = {frame_task};
+  std::vector<std::shared_ptr<Constraints>> const constraints;
+  std::vector<std::shared_ptr<Barrier>> const barriers = {barrier};
 
   // Run IK loop
   Eigen::VectorXd q_current = q;
-  constexpr int kMaxIterations = 100;
-  constexpr double kPositionTolerance = 0.02;  // 2cm
+  constexpr int k_max_iterations = 100;
```
sea-bass, line 225: These probably should be constants, just that they're in the wrong place. Anonymous namespace should fix it.

roboplan_rrt/src/rrt.cpp
```diff
@@ -39,10 +42,10 @@ RRT::RRT(const std::shared_ptr<Scene> scene, const RRTOptions& options)
       throw std::runtime_error("Multi-DOF joints not yet supported by RRT.");
     case JointType::CONTINUOUS:
       // The solution squashes the continuous position vectors to be used as an SO(2).
-      state_space_names.push_back("SO2");
+      state_space_names.emplace_back("SO2");
```
sea-bass: Does this matter?

roboplan_toppra/include/roboplan_toppra/toppra.hpp
```diff
@@ -26,7 +26,7 @@ class PathParameterizerTOPPRA {
   /// @brief Constructor.
   /// @param scene A pointer to the scene to use for path parameterization.
   /// @param group_name The name of the joint group to use.
-  PathParameterizerTOPPRA(const std::shared_ptr<Scene> scene, const std::string& group_name = "");
+  PathParameterizerTOPPRA(const std::shared_ptr<Scene> scene, std::string group_name = "");
```
sea-bass: Why did it change this from co st ref to value? Suspicious

roboplan_toppra/include/roboplan_toppra/toppra.hpp
```diff
@@ -70,15 +70,15 @@ class PathParameterizerTOPPRA {
   /// and therefore lead to collisions.
   /// @param path_pos_vecs The joint path position vectors.
   /// @return The resulting cubic spline.
-  std::shared_ptr<toppra::PiecewisePolyPath>
+  static std::shared_ptr<toppra::PiecewisePolyPath>
```
sea-bass: Why static?

roboplan_toppra/include/roboplan_toppra/toppra.hpp
```diff
@@ -70,15 +70,15 @@ class PathParameterizerTOPPRA {
...
   generateCubicSpline(const toppra::Vectors& path_pos_vecs);
 
   /// @brief Helper function to extract a cubic Hermite spline from a joint path.
   /// @details This enforces zero velocity and acceleration at all waypoints, meaning the desired
   /// path is exactly adhered to. If the path was checked for collisions, this spline is also safe.
   /// @param path_pos_vecs The joint path position vectors.
   /// @return The resulting cubic Hermite spline.
-  std::shared_ptr<toppra::PiecewisePolyPath>
+  static std::shared_ptr<toppra::PiecewisePolyPath>
```
sea-bass: Why static?

roboplan_toppra/src/toppra.cpp
```diff
@@ -8,12 +8,13 @@
 #include <roboplan/core/path_utils.hpp>
 #include <roboplan/core/scene_utils.hpp>
 #include <roboplan_toppra/toppra.hpp>
+#include <utility>
```
sea-bass: Another instance of system header being put in wrong place.

Though there are linters that can fix this

roboplan_toppra/test/test_toppra.cpp
```diff
@@ -86,7 +87,7 @@ JointPath createTestPathLong() {
   }
   {
     Eigen::VectorXd point(6);
-    point << -0.286816, 1.85311, 0.155756, -2.4643, -0.126047, -2.71853;
+    point << -0.286816, 1.85311, 0.155756, -2.4643, -0.126047, -std::numbers::e;
```
sea-bass: This is false. The joint angle is not meant to be e

scripts/run_clang_tidy.bash
```diff
@@ -0,0 +1,128 @@
...
+      echo "Usage: $0 [<package>] [--fix]"
+      echo ""
+      echo "Arguments:"
+      echo "  <package>  Run clang-tidy on specific package (optional)"
+      echo ""
+      echo "Options:"
+      echo "  --fix      Apply clang-tidy fixes automatically"
+      echo "  --help     Show this help message"
+      echo ""
+      echo "Valid package names: roboplan, roboplan_simple_ik, roboplan_rrt,"
```
sea-bass: Don't enumerate the packages in this hard coded fashion. This is prone to get outdated.

scripts/run_clang_tidy.bash
```diff
@@ -0,0 +1,128 @@
...
+      ;;
+  esac
+done
+
+# Define packages to check (in dependency order)
+if [[ -n "$PACKAGE_NAME" ]]; then
+  PACKAGES=("$PACKAGE_NAME")
+else
+  # All packages with C++ source code
+  PACKAGES=(
```
sea-bass: You have a variable here you could use for the error message

sea-bass (changes requested):
Requesting changes on 3 main things:

1. The linter seems to be using east const and const west haphazardly. I assume there are ways to enforce just one style, and it should be const west.
2. We may not want matrix variables to get shrunk down to lower case. If there are ways to make exceptions here, do consider it. Although I see some potentially problematic edge cases like `G_no`. Maybe just don't check this.
3. There are some suspicious things happening with making class member functions static that I don't understand. And a couple other freak instances like turning a string input arg from const ref to value, and a joint angle to the mathematical constant `e` just because it was close to a few decimal points.

Other less important comments inline, but this requires a little careful checking.

roboplan_oink/include/roboplan_oink/optimal_ik.hpp
sjahr: It's a bug, I'll fix it

roboplan_rrt/src/rrt.cpp
sjahr: No really :shrug:  but apparently emplace_back is more idiomatic for modern c++. I guess that's why clang flagged it. Here's claude's answer:
```
8:28 PM
Claude responded: For string literals, emplaceback has a slight theoretical edge — it constructs the std::string in-place from the const char* without creating a temporary.

For string literals, emplace_back has a slight theoretical edge — it constructs the std::string in-place from the const char* without creating a temporary. push_back constructs a temporary std::string first, then moves it in.

In practice, modern compilers optimize this away entirely, so there's no measurable difference. Both are fine.

The idiomatic preference these days:

    Use emplace_back when constructing from raw arguments (like a string literal, or multi-arg constructors) — it expresses intent clearly.
    Use push_back when you already have a std::string object — it's equally efficient and slightly more readable.

So emplace_back("SO2") is arguably the more modern/idiomatic choice here, but it won't matter in practice.
```
I am Ok with keeping it like this.

.github/workflows/build_and_test.yml
```diff
@@ -138,6 +138,9 @@ jobs:
           pixi run -e ci install_all
           pixi run -e ci build_tests
           pixi run -e ci test_all
+      - name: Run clang-tidy
+        run: |
+          pixi run -e clang-tidy
```
sea-bass:
```suggestion
          pixi run -e ci clang-tidy

sea-bass (commented):
2 things

1. The variable naming rules are still in place, so matrices are being turned lower case. I think this is a non-starter for me and we should disable it.
2. Can we make sure only 1 CI job runs clang-tidy? Like maybe just the Pixi 26.04 one? No reason for all 3 of them to.

sea-bass (approved):

PR #185, opened 2026-04-20
Use parallel builds in vanilla CMake script

sea-bass:
Should add this because why not?

We're not (and hope to never be) in a space where using all cores OOMs the system. Especially since this anyway builds 1 cmake project at a time.


eholum (approved):
before:

Image

after:

Image

... to the moon.

PR #193, opened 2026-05-05
Fix bugs in collision object scene operations

### LINKED ISSUE #192: Scene geometry bookkeeping bugs after remove/reparent/add operations
Hi!
I had some trouble handling repeated remove/add of objects in the scene with roboplan, and I verified there's an unexpected behavior while handling objects removal in roboplan. I currently implemented a workaround in my application, but I wonder if this is intended, consequential to some constraint or a bug to be fixed.
The following is a AI-generated report, but I verified the behavior in the app. 
I left out the suggested fixes but I can add them back if they could be useful.

## Summary

`roboplan::Scene` has three geometry bookkeeping issues that compound across valid scene mutation sequences:

1. **`Scene::removeGeometry()` leaves `collision_geometry_map_` partially stale.** `pinocchio::GeometryModel::removeGeometryObject(name)` compacts `geometryObjects` and shifts every later index down by one, but `removeGeometry` only erases the removed name from the cache. All other names retain their pre-removal indices and now resolve to the wrong geometry.

2. **`Scene::updateGeometryPlacement()` updates `parentFrame` and `placement` but not `parentJoint`.** Reparenting from world to a robot link leaves Pinocchio computing `oMg` under the old kinematic parent, decoupling the geometry from the new parent link.

3. **`Scene::addBoxGeometry()` / `addSphereGeometry()` swap `FrameIndex` and `JointIndex` in the `pinocchio::GeometryObject` constructor call.** The non-deprecated pinocchio constructor at `multibody/geometry-object.hpp:151` is `(name, JointIndex parent_joint, FrameIndex parent_frame, SE3, geom)`; the deprecated overload at line 191 has the older `(name, FrameIndex, JointIndex, geom, SE3)` order. Roboplan calls it as `(name, parent_frame_id, parent_joint_id, SE3, geom)` - matching neither, but resolving to the non-deprecated overload with arguments transposed. Both indices are typedef'd to `size_t`, so the compiler accepts the swap silently.

   On the UR-style models we tested, the only frame where the swap survives pinocchio's `frames[parentFrame].parentJoint == parentJoint` validation is `universe` (both indices are `0`). Other models could pass it by index coincidence, but the call order is still wrong.

Together, these silently corrupt the planner's collision world after any sequence containing: a non-LIFO removal, a reparent, or an add at a non-`universe` frame.

## Observed behavior

The following sequence reproduces bugs (1) and (2) end-to-end:

1. Add `ground`, `graspable`, `box_1` (in that order).
2. Remove and re-add each object in turn. Each non-LIFO remove shifts later indices.
3. Attach `graspable` to `tool0`.

After step 2, the cache state is:

```text
collision_geometry_map_:  ground -> 13   graspable -> 13   box_1 -> 13
geometryObjects[]:        ... ground@11   graspable@12   box_1@13
```

In step 3, `Scene::updateGeometryPlacement("graspable", "tool0", t)` follows `collision_geometry_map_["graspable"] = 13` and mutates `box_1` instead. Bug (2) then sets `box_1.parentFrame = tool0` while leaving `box_1.parentJoint = 0`, so pinocchio evaluates `oMg = oMi[0] * placement = identity * t_tool0_graspable â‰ˆ (0, 0, 0.05)` - `box_1` is teleported onto the robot base.

A name-keyed query for `graspable` reports it correctly attached to `tool0`, and a name-keyed query for `box_1` reports it at its requested pose. But `pinocchio::computeCollisions` over `geometryObjects[]` evaluates the corrupt entry and emits phantom collisions:

```text
Collision detected between base_link_0 and box_1
Collision detected between shoulder_link_0 and box_1
Collision detected between ground and box_1
```

Bug (3) reproduces independently with two lines:

```cpp
roboplan::Scene scene(...);                       // any URDF with a non-universe frame
auto r = scene.addBoxGeometry("foo", "tool0",
                              roboplan::Box(0.05, 0.05, 0.05),
                              Eigen::Matrix4d::Identity(), color);
// r.error() == "The object joint parent and its frame joint parent do not match."
```

## Diagnostic divergence

This bug class is hard to spot because `Scene` resolves geometries by **name** for some operations and by **index** for others:

- By name (via `collision_geometry_map_`): `removeGeometry`, `setCollisions`, `updateGeometryPlacement`, `getCollisionGeometryIds`.
- By index (over `geometryObjects[]` directly): `pinocchio::computeCollisions`, and any caller that iterates `collision_model_.geometryObjects` or `collisionPairs`.

When `collision_geometry_map_` goes stale, the nameâ†’index mapping silently disagrees with the underlying vector. The same name can resolve to one `GeometryObject` during a name-keyed mutation and a different one during collision evaluation. Callers that mix name-keyed scene queries with collision queries see the disagreement as "the scene I set up and the scene that's being collided against don't match," but the disagreement lives entirely inside `Scene`.

## Suspected root cause

```cpp
// 1. Scene::removeGeometry() - shifts indices but only erases removed name
collision_model_.removeGeometryObject(name);
collision_geometry_map_.erase(name);
collision_model_data_ = pinocchio::GeometryData(collision_model_);

// 2. Scene::updateGeometryPlacement() - parentJoint not updated
collision_geom.parentFrame = parent_frame_id.value();
collision_geom.placement   = pinocchio::SE3(tform);
// missing: collision_geom.parentJoint = model_.frames[parent_frame_id.value()].parentJoint;

// 3. Scene::addBoxGeometry() - swapped (FrameIndex, JointIndex) for current pinocchio API
pinocchio::GeometryObject geom_obj{
  name, parent_frame_id, parent_joint_id, pinocchio::SE3(tform), box.geom_ptr};
//      ^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^
// pinocchio expects:    JointIndex parent_joint,  FrameIndex parent_frame
```

`addSphereGeometry` and (apparently) `addOcTreeGeometry` follow the same pattern.

## Expected behavior

- Removing a geometry leaves all name-to-geometry lookups valid.
- Updating/reparenting keeps `parentFrame`, `parentJoint`, and `placement` consistent.
- Adding at any valid frame succeeds, with `parentFrame` and `parentJoint` consistent.
- `Scene` never silently mutates a different geometry than the requested name.

sea-bass:
Closes #192

### DIFF
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -601,7 +601,7 @@ tl::expected<void, std::string> Scene::addBoxGeometry(const std::string& name,
   const auto& parent_frame_id = maybe_parent_frame_id.value();
   const auto parent_joint_id = model_.frames.at(parent_frame_id).parentJoint;
 
-  pinocchio::GeometryObject geom_obj{name, parent_frame_id, parent_joint_id, pinocchio::SE3(tform),
+  pinocchio::GeometryObject geom_obj{name, parent_joint_id, parent_frame_id, pinocchio::SE3(tform),
                                      box.geom_ptr};
   geom_obj.meshColor = color;
   return addGeometry(geom_obj);
@@ -619,7 +619,7 @@ tl::expected<void, std::string> Scene::addSphereGeometry(const std::string& name
   const auto& parent_frame_id = maybe_parent_frame_id.value();
   const auto parent_joint_id = model_.frames.at(parent_frame_id).parentJoint;
 
-  pinocchio::GeometryObject geom_obj{name, parent_frame_id, parent_joint_id, pinocchio::SE3(tform),
+  pinocchio::GeometryObject geom_obj{name, parent_joint_id, parent_frame_id, pinocchio::SE3(tform),
                                      sphere.geom_ptr};
   geom_obj.meshColor = color;
   return addGeometry(geom_obj);
@@ -637,7 +637,7 @@ tl::expected<void, std::string> Scene::addOcTreeGeometry(const std::string& name
   const auto& parent_frame_id = maybe_parent_frame_id.value();
   const auto parent_joint_id = model_.frames.at(parent_frame_id).parentJoint;
 
-  pinocchio::GeometryObject geom_obj{name, parent_frame_id, parent_joint_id, pinocchio::SE3(tform),
+  pinocchio::GeometryObject geom_obj{name, parent_joint_id, parent_frame_id, pinocchio::SE3(tform),
                                      octree.geom_ptr};
   geom_obj.meshColor = color;
   return addGeometry(geom_obj);
@@ -675,13 +675,15 @@ tl::expected<void, std::string> Scene::updateGeometryPlacement(const std::string
   }
   const auto& collision_geom_idx = it->second;
 
-  const auto parent_frame_id = getFrameId(parent_frame);
-  if (!parent_frame_id) {
-    return tl::make_unexpected(parent_frame_id.error());
+  const auto maybe_parent_frame_id = getFrameId(parent_frame);
+  if (!maybe_parent_frame_id) {
+    return tl::make_unexpected(maybe_parent_frame_id.error());
   }
+  const auto parent_frame_id = maybe_parent_frame_id.value();
 
   auto& collision_geom = collision_model_.geometryObjects[collision_geom_idx];
-  collision_geom.parentFrame = parent_frame_id.value();
+  collision_geom.parentFrame = parent_frame_id;
+  collision_geom.parentJoint = model_.frames[parent_frame_id].parentJoint;
   collision_geom.placement = pinocchio::SE3(tform);
   return {};
 }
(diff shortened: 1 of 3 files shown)


pastoriomarco:
I tested this PR and it solves the original #192 !

roboplan/src/core/scene.cpp
```diff
@@ -601,7 +601,7 @@ tl::expected<void, std::string> Scene::addBoxGeometry(const std::string& name,
   const auto& parent_frame_id = maybe_parent_frame_id.value();
   const auto parent_joint_id = model_.frames.at(parent_frame_id).parentJoint;
 
-  pinocchio::GeometryObject geom_obj{name, parent_frame_id, parent_joint_id, pinocchio::SE3(tform),
+  pinocchio::GeometryObject geom_obj{name, parent_joint_id, parent_frame_id, pinocchio::SE3(tform),
```
eholum, line 604: lol, derp.

roboplan/src/core/scene.cpp
```diff
@@ -691,6 +693,16 @@ tl::expected<void, std::string> Scene::removeGeometry(const std::string& name) {
   if (it == collision_geometry_map_.end()) {
     return tl::make_unexpected("Could not find object '" + name + "' to remove.");
   }
+
+  // Update all the indices that came after since Pinocchio shifts them down when an object is
+  // removed.
+  const auto old_geom_idx = it->second;
+  for (auto& [name, index] : collision_geometry_map_) {
```
eholum: mega nit: reusing `name` since I did a double take...


```suggestion
  for (auto& [key, index] : collision_geometry_map_) {
```

eholum (approved):
Looks good! But didn't explicitly test #193, it just seems reasonable.

sea-bass:
Why doesn't this test explicitly test the issues? It checks whether the collision map's index matches what is in the Pinocchio model, and also that the object's `parentFrame` and `parentJoint` are as expected?

eholum:
Lol sorry, I didn't explicitly test it! But I believe the tests and others did!

PR #194, opened 2026-05-05
Lazy import plyfile to not require it in all examples

sea-bass:
Trying to use the ROS workflows, I noticed that running any example was requiring that `plyfile` was imported due to the new octree functionality. Changed things so that `plyfile` is only imported when octrees are selected, and generally cleaned things up a bit.

@CihatAltiparmak FYI


sea-bass:
I guess there's the other issue that in pixi we're installing `hppfcl` 3.0.2, which has the `toBoxes()` method, but in ROS we have version 2.5.4 which doesn't.

So octrees won't work until `hppfcl` / `coal` is updated on the ROS side.

roboplan_examples/python/common.py
```diff
@@ -1,52 +1,36 @@
...
-from typing import List
 
 import hppfcl
 import numpy as np
 from numpy.typing import NDArray
 import pinocchio as pin
 
 from roboplan.core import Box, Cylinder, Mesh, Scene, Sphere, OcTree
-from roboplan.example_models import get_package_models_dir, get_package_share_dir
-from plyfile import PlyData
```
eholum, line 16: Not sure how the heck this used to work in the ros containers but I think we need a rosdep on `python-plyfile-pip` for the examples package to install it? Still fails on the `toBoxes` call as you noted but hey... ros gonna ros.

eholum (commented):
Much cleaner! Am I missing something with the plyfile dep?

roboplan_examples/python/common.py
eholum: Or just add it to the pip install in the Dockerfile since it's a pip package...

sea-bass: I don't think it ever did work... because of old hppfcl versions. BUT I have seen the pinocchio 4.0 / coal 3.0.2 packages go up in rosdistro, so it's a matter of time

As for the plyfile import, yeah it's just pip installed in the container

eholum: Oh yeah it is for the [non-ROS workflow](https://github.com/open-planning/roboplan/blob/main/.docker/ubuntu/Dockerfile#L25) but not for the [ROS workflow](https://github.com/open-planning/roboplan/blob/main/.docker/ros/Dockerfile#L18). I get import errors running the example RRT on it so it's probably worth adding there too? Even if `toBoxes` is broken.

eholum: Update: well there it is.

sea-bass: 😬

.docker/ubuntu/Dockerfile
```diff
@@ -17,12 +17,12 @@ RUN pip3 install nanobind pytest scikit-build-core
 # Also install additional dependencies for examples.
 RUN pip3 install \
     matplotlib \
-    pin==${PINOCCHIO_VERSION} \
     opencv-contrib-python-headless \
+    pin==${PINOCCHIO_VERSION} \
+    plyfile \
     tyro \
     "viser>=1.0.1" \
     xacro \
```
eholum:
```suggestion
    xacro
```

eholum (approved):
Assuming the pipeline goes green now!

PR #196, opened 2026-05-08
Add OInK action chunk tracking example

### LINKED ISSUE #191: Proposal: Add example for constrained tracking of learned policy action chunks
Hi RoboPlanners, 

I am new to contributing to RoboPlan and would like to check whether this would be a useful contribution before I start implementing it. 

I am interested in adding a small Python example showing how RoboPlan can be used as a constraint-aware tracking layer for learned robot policies. 

Proposed example: 
- Use an existing example robot model, preferably Kinova Gen3 if suitable. 
- Define a mock learned-policy output as a short-horizon Cartesian action chunk. 
- Convert each Cartesian delta into a desired end-effector target pose. 
- Use RoboPlan IK or OInK to compute feasible robot configurations. 
- Visualize or print the resulting constrained trajectory. 
- Keep the example independent of any specific learning framework. 

My goal would be to start with a small, simple example first, then adjust based on maintainer feedback. Would this fit the direction of the project? If yes, I would be happy to work on a first version.

muslim-adedamola:
#191 
## Summary

This PR adds a RoboPlan example demonstrating how short-horizon action chunks from a learned policy can be tracked through OInK.

The example supports two mock policy action spaces:

- Cartesian/end-effector deltas: `[dx, dy, dz, droll, dpitch, dyaw]`
- Joint-space deltas: `[dq_1, dq_2, ..., dq_n]`

The sparse action chunks are converted into sparse target poses/configurations, interpolated at a smaller control timestep, and tracked using OInK with position and velocity limits.

## Visualization

The example visualizes:

- Orange markers: sparse policy waypoints
- Gray trace: dense interpolated targets
- Blue trace: executed OInK-constrained trajectory
- Red marker: current end-effector position during animation/stepping

The Viser GUI includes buttons to animate the trajectory, step through it, or reset.

## Tested

I tested the example with:

- Kinova Gen3
- Franka
- UR5

Example commands:

```bash
python roboplan_examples/python/example_action_chunk_tracking.py --model kinova --action-space cartesian
python roboplan_examples/python/example_action_chunk_tracking.py --model kinova --action-space joint
python roboplan_examples/python/example_action_chunk_tracking.py --model franka --action-space joint
python roboplan_examples/python/example_action_chunk_tracking.py --model ur5 --action-space cartesian
```

***Notes

While testing, I noticed that Kinova appears to initialize from the zero configuration by default, unlike the UR5 and Franka examples. I added an optional `--use-kinova-home` flag to demonstrate a more recognizable Kinova ready/home pose.

```bash
python roboplan_examples/python/example_action_chunk_tracking.py --model kinova --action-space cartesian --use-kinova-home
```


roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,792 @@
...
+    model_data,
+    v_indices: np.ndarray,
+    num_velocity_variables: int,
+    use_kinova_home: bool = False,
+) -> np.ndarray:
+    """Return a good starting configuration for the selected model.
+
+    For most models, it uses the starting configuration from common.py. For Kinova, it integrates
+    a recognizable home/ready arm pose from the neutral scene configuration.
+    """
```
sea-bass, line 192: honestly, I would just change the starting joint positions of the Kinova robot to be something like this. This is much better than the zero position.

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,792 @@
...
+        q_end = pin.Quaternion(end.rotation)
+
+        for k in range(substeps):
+            alpha = k / float(substeps)
+            translation = (1.0 - alpha) * start.translation + alpha * end.translation
+            rotation = q_start.slerp(alpha, q_end).matrix()
+            dense.append(pin.SE3(rotation, translation).homogeneous)
+
+    dense.append(np.asarray(waypoints[-1], dtype=float))
+    return dense
```
sea-bass: I feel like this stuff should be part of the core library in C++ and python binded.

But maybe okay for now just to make a utilities file in Python

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,792 @@
...
+    return np.asarray([np.asarray(tform)[:3, 3].copy() for tform in target_transforms])
+
+
+# Compute EE positions induced by full joint-space target configurations.
+def joint_target_positions(
+    scene: Scene,
+    q_full_targets: list[np.ndarray],
+    ee_frame_name: str,
+) -> np.ndarray:
+    return compute_end_effector_positions(scene, q_full_targets, ee_frame_name)
```
sea-bass: if this function just calls an existing function, seems unnecessary

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,792 @@
...
+        try:
+            viz.viewer.scene.add_icosphere(
+                f"{waypoint_root}/step_{idx}",
+                radius=waypoint_radius,
+                position=position,
+                color=waypoint_color,
+            )
+        except Exception as exc:
+            print(f"Warning: could not draw waypoint marker {idx}: {exc}")
+            break
```
sea-bass: maybe move this stuff to the `visualization.py` file?

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,792 @@
...
+    # Show the fixed start configuration before creating targets or solving any
+    # trajectory. This prevents the example from implicitly using the robot's
+    # default model pose as the beginning of the action chunk.
+    scene.setJointPositions(q_start)
+    viz.display(q_start)
+
+    # Set up OInK only after the scene has been placed at the selected start
+    # configuration.
+    scene.setJointPositions(q_start)
+    oink = Oink(scene, joint_group)
```
sea-bass: might be good in this case to set up the oink stuff in a helper function

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,792 @@
...
+            viz.viewer.scene.add_line_segments(
+                name,
+                points=line_segments,
+                color=color,
+                line_width=line_width,
+            )
+        except Exception as exc:
+            print(f"Warning: could not draw line-segment trace {name}: {exc}")
+    except Exception as exc:
+        print(f"Warning: could not draw line-segment trace {name}: {exc}")
```
sea-bass: too much try/catch, maybe consider just making this work on the latest version of viser

sea-bass (commented):
Thank you!

I didn't run/test it yet, but there is definitely a lot of code at ~700 lines of code for an example. There might be some redundant/unnecessary stuff (I suggested a few places), and others that can live elsewhere.

muslim-adedamola:
Thank you for the review! I have pushed a pass addressing all comments.

bindings/src/roboplan/interpolation.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def computeStepsPerSegment(policy_dt: float, control_dt: float) -> int:
+    """Compute the number of interpolation steps per policy segment."""
+    if policy_dt <= 0.0:
+        raise ValueError("policy_dt must be positive.")
+    if control_dt <= 0.0:
+        raise ValueError("control_dt must be positive.")
+
+    return max(1, int(math.ceil(policy_dt / control_dt)))
```
sea-bass: Two comments:

1. I'm not entirely sure what `policy_dt` means in this setting, but presumably the duration of an action chunk from a policy? Could this more generally be renamed to `segment_time` since it could be used in other applications too?
2. Shouldn't this be `ceil(policy_dt / control_dt) + 1`? For example, if a segment is 1 second and has a sample time of 0.1, there should be 11 points, not 10.

bindings/src/roboplan/interpolation.py
```diff
@@ -0,0 +1,69 @@
+import math
+
+import numpy as np
+import pinocchio as pin
+
+from roboplan.core import Scene
+
+
+def computeStepsPerSegment(policy_dt: float, control_dt: float) -> int:
+    """Compute the number of interpolation steps per policy segment."""
```
sea-bass: Make sure you also add the arguments/return values to all these docstrings!

bindings/src/roboplan/interpolation.py
```diff
@@ -0,0 +1,69 @@
...
+        raise ValueError("policy_dt must be positive.")
+    if control_dt <= 0.0:
+        raise ValueError("control_dt must be positive.")
+
+    return max(1, int(math.ceil(policy_dt / control_dt)))
+
+
+def interpolateConfigurationWaypoints(
+    scene: Scene,
+    waypoints: list[np.ndarray] | np.ndarray,
```
sea-bass: do we need to support `np.ndarray`, or should this just always be a list? Particularly because `len(waypoints)` means different things whether it's a list or not.

bindings/src/roboplan/interpolation.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    if len(waypoints) < 2:
+        return list(waypoints)
+
+    steps_per_segment = computeStepsPerSegment(policy_dt, control_dt)
+    dense_waypoints = []
+
+    for idx in range(len(waypoints) - 1):
+        start = np.asarray(waypoints[idx])
+        end = np.asarray(waypoints[idx + 1])
```
sea-bass: This is redundantly calling `np.asarray` on the same data for all points except for the first and last... but also, isn't this already a list of `np.ndarray`s? So do we even need the conversion in this case? I think not.

bindings/src/roboplan/interpolation.py
```diff
@@ -0,0 +1,69 @@
...
+
+        for step in range(steps_per_segment):
+            alpha = step / steps_per_segment
+            dense_waypoints.append(scene.interpolate(start, end, alpha))
```
sea-bass: I think I see a major flaw in this, in that it assumes that all waypoints are evenly spaced in time (which is that `policy_dt`). You could actually make this more generic if you use the `JointTrajectory` type in `roboplan/include/roboplan/core/types.hpp`. And then you can actually use the `times` field to determine actual spacing.

There would need to be a new `CartesianTrajectory` added here too, which I think is worthwhile!

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+
+    current_target = scene.forwardKinematics(q_start, ee_frame_name)
+    targets = [current_target.copy()]
+
+    target_se3 = pin.SE3(current_target)
+    for delta in action_chunk:
+        target_se3 = target_se3 * se3_from_delta(delta)
+        targets.append(target_se3.homogeneous.copy())
+
+    return targets
```
sea-bass: Cartesian action chunks don't have to be relative... they can just be absolute, and in fact I would recommend having them be absolute so you don't need the `se3_from_delta` function.

I mean, overall I suppose it would be nice to support both, but for simplicity maybe absolute is the way to go.

---

Secondarily, I find it a little inconvenient that we're converting numpy to Pinocchio to do transformations, and then back to numpy, and then back to Pinocchio to do the interpolation down the line. Another reason for doing all the relative stuff before this step, and having the homogeneous matrices ready to go.

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+    targets = [np.asarray(q_full_start, dtype=float).copy()]
+    q_target = np.asarray(q_full_start, dtype=float).copy()
+
+    for delta_group in action_chunk:
+        delta_full = np.zeros(num_velocity_variables)
+        delta_full[v_indices] = np.asarray(delta_group, dtype=float)
+        q_target = scene.integrate(q_target, delta_full)
+        targets.append(q_target.copy())
+
+    return targets
```
sea-bass: same comment here re: action chunks being absolute instead of relative

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+
+    The joint-space action chunk lives in the robot tangent/velocity space, not
+    directly in configuration space. Therefore, each delta is lifted into the full
+    velocity vector and applied using scene.integrate().
+    """
+    targets = [np.asarray(q_full_start, dtype=float).copy()]
+    q_target = np.asarray(q_full_start, dtype=float).copy()
+
+    for delta_group in action_chunk:
+        delta_full = np.zeros(num_velocity_variables)
```
sea-bass: you can just initialize this once outside the for loop

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+def compute_end_effector_positions(
+    scene: Scene,
+    configurations: list[np.ndarray],
+    ee_frame_name: str,
+) -> np.ndarray:
+
+    positions = []
+
+    for q in configurations:
+        scene.setJointPositions(q)
```
sea-bass: do you need to do this for the forward kinematics to work? I don't think so

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+        ee_tform = scene.forwardKinematics(q, ee_frame_name)
+        positions.append(ee_tform[:3, 3].copy())
+
+    return np.asarray(positions)
+
+
+# Extract xyz positions from Cartesian SE(3) target transforms
+def cartesian_target_positions(target_transforms: list[np.ndarray]) -> np.ndarray:
+    return np.asarray([np.asarray(tform)[:3, 3].copy() for tform in target_transforms])
```
sea-bass: why are we extracting only the position part? Is there any harm in having the trajectories be full 4x4 homogeneous matrices corresponding to position and orientation?

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+
+# Create an OInK solver from a fixed starting configuration
+def create_oink_solver(
+    scene: Scene,
+    joint_group: str,
+    q_start: np.ndarray,
+) -> Oink:
+
+    scene.setJointPositions(q_start)
+    return Oink(scene, joint_group)
```
sea-bass: this helper function is just 2 lines, I think you can remove it... my original comment was to also have all the other tasks and constraints in here, in which case it would justify needing this.

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+        pin.GeometryType.VISUAL,
+        package_dirs=package_paths,
+    )
+
+    viz = ViserVisualizer(model_pin, collision_model, visual_model)
+    viz.initViewer(open=True, loadModel=True, host=host, port=port)
+    viz.display(q_start)
+
+    # Set up OInK after fixing the scene at the selected start configuration.
+    oink = create_oink_solver(scene, joint_group, q_start)
```
sea-bass: if the action space is joint, we don't even need to create the oink solver, right?

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+        step_button.disabled = True
+        reset_button.disabled = True
+
+        if state["step_idx"] >= len(trajectory) - 1:
+            state["step_idx"] = 0
+
+        try:
+            for idx in range(state["step_idx"], len(trajectory)):
+                display_step(idx)
+                time.sleep(animation_dt)
```
sea-bass: I don't think this is correct, since the default parameters are saying you're going to sleep 0.03 seconds between every waypoint no matter what the frequency is? So it makes the animation look really slow. Could you please check it?

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,567 @@
...
+    print("  blue:   executed OInK-constrained trajectory")
+    print("  red:    current end-effector position")
+    print(
+        "Use the Viser GUI buttons to animate, step through, or reset the trajectory."
+    )
+
+    state = {
+        "step_idx": 0,
+        "animating": False,
+    }
```
sea-bass: Making this be a dictionary is not great. Either use 2 separate variables (my preference because it's simpler), or make this a dataclass.

sea-bass (commented):
Just tested and gave the code a more thorough look -- thanks for your patience!

First off, the default Cartesian action chunk is very short here and hard to visualize, especially since the Robotiq covers a lot of the information. Could we make this follow a longer path? It also isn't possible to view all the layers here, the orange and red dots dominate everything. Another reason to space out the action chunks some more!

Image

There might actually be a way to also set the "layer" or priority in Viser so these lines always display on top of the robot? But I'm not sure.

When I extended the action chunk size, I noticed that the oink trajectory does not track very well for the Kinova or Franka.

Image

Image

For the UR5 it's okay. So maybe this should be the default model? Or the trajectory should be better tuned for the Kinova kinematics?

Image

Even worse for the SO-101:

Image

Finally, when I choose `dual` as my model, the example works, but it only does the action chunk for one of the arms. Is that expected, or should it work for both end effectors?

Image

---

As a final comment, but not necessarily for this PR, it would be nice if there was an example showing switching between two action chunks online to see how the IK solver reacts, and then plotting the joint trajectories over time. That would really solidify the value of this.

sea-bass:
Another interesting idea (not required) is that the "Step Once" button is actually too dense to be useful, since the number of steps is pretty large. I think Viser has the option of setting up sliders, so users can manually drag the robot arm along the action chunk. That would be really cool!

bindings/src/roboplan/interpolation.py
sea-bass: ... although if this is done per segment, then the final point of one segment would be the first point of the next segment. So have to also consider that downstream to not duplicate the same point at the waypoints.

muslim-adedamola:
Thanks again for the detailed review!!

I pushed another update. Main changes:

* Renamed `policy_dt` to `segment_time` and updated the docstrings.
* Fixed interpolation sampling so each segment includes both endpoints while avoiding duplicate shared waypoints.
* Added `interpolateJointTrajectory(...)` so joint trajectories can use `JointTrajectory.times` instead of assuming equal segment timing.
* Changed the mock Cartesian and joint chunks to use sparse absolute targets.
* Kept Cartesian targets as full 4x4 homogeneous transforms, and only extract xyz positions for trace visualization.
* Removed the small `create_oink_solver(...)` helper.
* Only create/use OInK for Cartesian tracking; joint-space targets are directly interpolated as configurations.
* Made UR5 the default model and made the default Cartesian/joint chunks more visually clear.
* Replaced the dense “Step once” button with a slider.
* Updated animation timing to use the control timestep with a playback speed multiplier.
* Clarified that multi-arm models like `dual` currently track only the default joint group / first end-effector.

A few notes from my side:

* I initially tested mainly with Franka, UR5, and Kinova, so I did not catch the `dual` and SO-101 behavior until you pointed it out.
* I did not implement coordinated dual-arm tracking yet because it seems like it could make this example much larger. My current thinking is that dual-arm action chunk tracking may be better as a separate example, but I’m happy to add it here if you’d prefer.
* I didn’t add a `CartesianTrajectory` equivalent yet because, unlike `JointTrajectory`, it does not currently exist in the core C++ types / Python bindings. For now, Cartesian targets are still handled as lists of full 4x4 homogeneous transforms and interpolated with `interpolateSE3Waypoints(...)`. Would you prefer that I add a proper `CartesianTrajectory` type and expose it through the Python bindings in this PR, or should we keep that as a follow-up?

* I noticed that joint-space target chunks track well across the robots. The larger divergence seems to happen mainly with Cartesian tracking, likely because the OInK position/velocity limits add constraints to the motion (my thinking, i may be wrong). I attached screenshots showing this (first 4, cartesian tracking, last 4, joint-space tracking). I can debug that further too if you think it should be improved in this PR.
j4-imageonline co-merged(1)



Let me know what you think!

roboplan_examples/python/example_action_chunk_tracking.py
sea-bass: bumping this comment

roboplan_examples/python/example_action_chunk_tracking.py
sea-bass: bumping this comment

sea-bass (commented):
Looks much better! I think there are just 2 comments left from previous review and should be good.

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -0,0 +1,549 @@
...
+        animating = True
+        animate_button.disabled = True
+        step_slider.disabled = True
+        reset_button.disabled = True
+
+        start_step_idx = int(step_slider.value)
+        if start_step_idx >= len(trajectory) - 1:
+            start_step_idx = 0
+
+        try:
```
sea-bass: does this need to be in a try block?

muslim-adedamola: It doesn’t actually. I added it so the GUI controls would be re-enabled just in case the animation loop exited unexpectedly.

I’ll simplify it and remove it then.

muslim-adedamola:
> Thanks again for the detailed review!!
> 
> I pushed another update. Main changes:
> 
>     * Renamed `policy_dt` to `segment_time` and updated the docstrings.
> 
>     * Fixed interpolation sampling so each segment includes both endpoints while avoiding duplicate shared waypoints.
> 
>     * Added `interpolateJointTrajectory(...)` so joint trajectories can use `JointTrajectory.times` instead of assuming equal segment timing.
> 
>     * Changed the mock Cartesian and joint chunks to use sparse absolute targets.
> 
>     * Kept Cartesian targets as full 4x4 homogeneous transforms, and only extract xyz positions for trace visualization.
> 
>     * Removed the small `create_oink_solver(...)` helper.
> 
>     * Only create/use OInK for Cartesian tracking; joint-space targets are directly interpolated as configurations.
> 
>     * Made UR5 the default model and made the default Cartesian/joint chunks more visually clear.
> 
>     * Replaced the dense “Step once” button with a slider.
> 
>     * Updated animation timing to use the control timestep with a playback speed multiplier.
> 
>     * Clarified that multi-arm models like `dual` currently track only the default joint group / first end-effector.
> 
> 
> A few notes from my side:
> 
>     * I initially tested mainly with Franka, UR5, and Kinova, so I did not catch the `dual` and SO-101 behavior until you pointed it out.
> 
>     * I did not implement coordinated dual-arm tracking yet because it seems like it could make this example much larger. My current thinking is that dual-arm action chunk tracking may be better as a separate example, but I’m happy to add it here if you’d prefer.
> 
>     * I didn’t add a `CartesianTrajectory` equivalent yet because, unlike `JointTrajectory`, it does not currently exist in the core C++ types / Python bindings. For now, Cartesian targets are still handled as lists of full 4x4 homogeneous transforms and interpolated with `interpolateSE3Waypoints(...)`. Would you prefer that I add a proper `CartesianTrajectory` type and expose it through the Python bindings in this PR, or should we keep that as a follow-up?
> 
>     * I noticed that joint-space target chunks track well across the robots. The larger divergence seems to happen mainly with Cartesian tracking, likely because the OInK position/velocity limits add constraints to the motion (my thinking, i may be wrong). I attached screenshots showing this (first 4, cartesian tracking, last 4, joint-space tracking). I can debug that further too if you think it should be improved in this PR.
> 
> 
> j4-imageonline co-merged(1)
> 
> Let me know what you think!

Seems like you missed it. Could you please let me know what you think regarding the dual arm, CartesianTrajectory, and the divergence?

sea-bass:
Oh, sorry -- I didn't miss it, I was just OK with how it was now and think after the small fixes we should merge this!

So to clarify:
* No need to support dual end effector yet
* No need to create a new `CartesianTrajectory` in C++ yet
* The divergence worries me a little bit. Just for curiosity, what happens if you exclude the velocity limit and/or position limits in the OInK solver? Do the trajectories match better? It would be good to know if it's because of this, or if there's maybe bugs with OInK.

Thank you!

muslim-adedamola:
> Oh, sorry -- I didn't miss it, I was just OK with how it was now and think after the small fixes we should merge this!
> 
> So to clarify:
> 
>     * No need to support dual end effector yet
> 
>     * No need to create a new `CartesianTrajectory` in C++ yet
> 
>     * The divergence worries me a little bit. Just for curiosity, what happens if you exclude the velocity limit and/or position limits in the OInK solver? Do the trajectories match better? It would be good to know if it's because of this, or if there's maybe bugs with OInK.
> 
> 
> Thank you!

Great.
So regarding the divergence, I checked it more.

Removing just the PositionLimit and/or VelocityLimit did not significantly change the divergence. 

The bigger factor seems to be the ConfigurationTask regularization. When I removed it and used only the Cartesian FrameTask, UR5, Kinova, and Franka tracked the target much more closely, even with PositionLimit + VelocityLimit restored. (Likely Explanation, configuration task was pulling the robot toward the default configuration while the frame task was trying to track the Cartesian target. This bias created the divergence).

With only the FrameTask, I measured about:

mean error: 0.0046 m
max error: 0.0082 m
final error: 0.0082 m

For SO-101, it was a bit different. in addition to removing the FrameTask, it only tracked well when I removed either both limits or the PositionLimit. Once the limits were restored, the divergence came back. So my guess is that the target path may be outside what the constrained SO-101 model can comfortably follow, while for UR5/Kinova/Franka the earlier divergence was mostly from the configuration regularization task. I attached screenshots too.

c6-imageonline co-merged


--------------------------------------------------------------------------------------------------------------------------------------------------------------------------

What do you think? should I remove the ConfigurationTask from this example and keep the OInK position/velocity limits? That seems to give clearer default tracking while still demonstrating constrained Cartesian tracking.

sea-bass:
Oh yeah! Let's actually remove the configurationtask (or set the gain/regularization very low) and keep the limits, and that should be great. Thank you for checking.

muslim-adedamola:
All done
Thanks for the helpful reviews!

sea-bass:
Just tested it one more time and made small cleanups -- I really like how the slider turned out. Thanks for the iteration @muslim-adedamola !

muslim-adedamola:
Very happy to be of help :blush:

sea-bass (approved):

PR #197, opened 2026-05-08
Support cylinder and mesh scene collision objects

### LINKED ISSUE #195: Add mesh (and cylinder) collision objects
Hi, 
I'm working on a scene replication function from Isaac Sim to RoboPlan. The main blocker is that right now I can't add mesh collision objects in RoboPlan (nor cylinder objects, which would be handy too).

Are there plans to add the mesh objects soon?

sea-bass:
What the title says. Also expanded the unit testing.

Closes #195.

image


pastoriomarco:
I verified that mesh objects load correctly now (see #195 ), thank you!

PR #201, opened 2026-05-11
Add MuJoCo pick-and-place example with Franka Panda

### LINKED ISSUE #190: Create simple pick and place sequence example with MuJoCo
Something like: using known ground truth pose of a cube, plan a collision free trajectory including pre-grasp locations with Cartesian motions, etc.

Can connect to MuJoCo to demonstrate this tool with something that has physics

crisiumnih:
Ref - #190 
Uses the Franka Panda model from MuJoCo Menagerie for physics simulation, and the FR3 URDF from roboplan's example models for motion planning (IK, RRT, TOPP-RA). I found out that the two models are kinematically near-identical so plans transfer directly.

Full sequence: SimpleIK for grasp/lift/place waypoints, RRT for collision-free joint-space motion, TOPP-RA for time parameterization, and manual cube tracking during carry.

Doubts - 
- someone would have to clone MuJoCo Menagerie, to run the example, is this a valid option?
- shall i add randomize cube spawn and place target within a safe reachable region to demonstrate planning.
- Uses the MuJoCo passive viewer only, shall I add keep Viser as main or both?

https://github.com/user-attachments/assets/3fb37729-67cb-496b-ba46-df0d2c00eeb2

### DIFF
--- a/roboplan_examples/python/example_pick_place_mujoco.py
+++ b/roboplan_examples/python/example_pick_place_mujoco.py
@@ -10,7 +10,7 @@ import tyro
 import xacro
 
 from common import MODELS
-from roboplan.core import CartesianConfiguration, JointConfiguration, JointPath, Scene
+from roboplan.core import Box, CartesianConfiguration, JointConfiguration, JointPath, PathShortcutter, Scene
 from roboplan.example_models import get_package_share_dir
 from roboplan.rrt import RRT, RRTOptions
 from roboplan.simple_ik import SimpleIk, SimpleIkOptions
@@ -60,8 +60,15 @@ def main(
     rrt_max_nodes: int = 4000,
     rrt_max_connection_distance: float = 1.5,
     rrt_collision_check_step_size: float = 0.05,
-    rrt_goal_biasing_probability: float = 0.15,
+    rrt_goal_biasing_probability: float = 0.4,
     rrt_connect: bool = True,
+    rrt_shortcut_iters: int = 200,
+    # Wall obstacle (blocks transport path, forces RRT to route around)
+    wall_x: float = 0.5,
+    wall_y: float = 0.2,
+    wall_half_size_x: float = 0.15,
+    wall_half_size_y: float = 0.01,
+    wall_half_size_z: float = 0.15,
     # TOPP-RA
     toppra_dt: float = 0.02,
     toppra_mode: SplineFittingMode = SplineFittingMode.Hermite,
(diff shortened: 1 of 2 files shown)


sea-bass:
Very cool!

> someone would have to clone MuJoCo Menagerie, to run the example, is this a valid option?

This is totally fine; they will also need to pip install mujoco as well. Since it's an example and not the core library, it's okay.

However, it would be even better if we could instead have our own MJCF XML file in the repo pointing to the same meshes. There should be several options for automatic conversion, including in https://github.com/ros-controls/mujoco_ros2_control

> shall i add randomize cube spawn and place target within a safe reachable region to demonstrate planning.

That would be cool, but not required. Maybe more important would be to have more obstacles in the scene; for example a wall between the pick and place locations.

> Uses the MuJoCo passive viewer only, shall I add keep Viser as main or both?

Viser would be able to show the RRT, which is nice, but it also would make the example code busy. How about no Viser, but we display the resulting path in the mujoco viewer?

crisiumnih:
Added changes

https://github.com/user-attachments/assets/1b28cb56-e0cd-45b2-a0f3-e6facd4c849a

sea-bass:
So, it seems the block clips the obstacle. Do we need to include it as a collision geometry in this example and allow collisions between it and the fingers/table?

Also the stop and go behavior in the RRT is expected with the default "Hermite" mode of the TOPPRA parameterizer. If you change this to "Adaptive" it should on average move much smoother!

roboplan_examples/python/example_pick_place_mujoco.py
```diff
@@ -0,0 +1,421 @@
...
+    lift_traj = plan_linear(grasp_solution, lift_solution, "Lift")
+    transport_traj = plan_rrt(lift_solution, preplace_solution, "Transport")
+    place_traj = plan_linear(preplace_solution, place_solution, "Place descent")
+    retreat_traj = plan_rrt(place_solution, start_jc, "Retreat")
+
+    # --- MuJoCo body IDs for cube tracking ---
+    hand_body_id = mujoco.mj_name2id(model, mujoco.mjtObj.mjOBJ_BODY, "hand")
+    cube_jnt_addr = model.jnt_qposadr[model.body_jntadr[cube_body_id]]
+
+    def draw_ee_path(traj, color: np.ndarray) -> None:
```
sea-bass, line 314: I wonder if we should.make a mujoco utils file in the examples to keep this in?

sea-bass:
Closing due to inactivity, and there has been another related example merged in by now.

PR #203, opened 2026-05-12
Add CartesianTrajectory type

### LINKED ISSUE #199: Create `CartesianTrajectory` type and update action chunking example
Right now, the new action chunking example is treating Cartesian trajectories as loose lists of transforms and times; would be good to implement an equivalent to `JointTrajectory` for these purposes.

muslim-adedamola:
PR intended to fix issue #199


sea-bass (approved):
Looks great! Just need to update the stub files with the correct nanobind version by running:

```
pixi run -e ci build_all
```

and you should be good!

Issue #200, opened 2026-05-10
Make action chunking example support dual end effectors


sea-bass:
This also includes ensuring the joint action space plots trails for both end effectors for the dual Franka.

muslim-adedamola:
I have updated the action chunking example to support dual end effectors. 
also tested on single arms and it works.

<img width="720" height="432" alt="Image" src="https://github.com/user-attachments/assets/19ed5c32-39a1-4815-afc0-100d43f59a90" />
<img width="720" height="432" alt="Image" src="https://github.com/user-attachments/assets/4aaae421-f881-447b-bd93-b68ca1acc1cd" />

----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
I remember you mentioned something like this during the early review

"As a final comment, but not necessarily for this PR, it would be nice if there was an example showing switching between two action chunks online to see how the IK solver reacts, and then plotting the joint trajectories over time. That would really solidify the value of this."

Do you think I should add that feature here before opening the PR, or would it be better to keep this PR focused on dual end-effector support and make the plotting+switching a separate follow-up?

sea-bass:
Cool! It's always good to split up the functionality, if you're willing to do that. Nice (and fast!) progress.

Especially because the switching action chunks may require a little back and forth review to see how we can make it look good and understandable for users

muslim-adedamola:
Should I close this and we open an issue for this:

""As a final comment, but not necessarily for this PR, it would be nice if there was an example showing switching between two action chunks online to see how the IK solver reacts, and then plotting the joint trajectories over time. That would really solidify the value of this.""

sea-bass:
Yeah, this can be closed and the other piece of feedback can be implemented separately -- thanks!

PR #204, opened 2026-05-12
Support dual end effectors in action chunk example

muslim-adedamola:
## Summary

Updates `example_action_chunk_tracking.py` to support multiple end effectors. #200 

The example now:
- uses all `model_data.ee_names` instead of only the first ee
- creates one Cartesian `FrameTask` per ee
- generates and interpolates Cartesian target chunks per ee
- visualizes sparse and dense targets, executed traces, and current EE markers for each ee
- plots joint-space end-effector trails for both end effectors


roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -107,6 +107,71 @@ def make_mock_cartesian_target_poses(
...
+    Returns:
+        A pair of dictionaries keyed by end-effector frame name:
+            1. Sparse Cartesian target transforms.
+            2. Dense interpolated Cartesian target transforms.
+    """
+    sparse_targets_by_frame = {}
+    dense_targets_by_frame = {}
+
+    for ee_idx, ee_frame_name in enumerate(ee_frame_names):
+        sparse_targets = make_mock_cartesian_target_poses(
```
sea-bass: I think I might prefer not having this be 2 functions: just combine the contents of `make_mock_cartesian_target_poses` and `make_mock_cartesian_trajectories_by_frame` into one function.

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -107,6 +107,71 @@ def make_mock_cartesian_target_poses(
...
+                    dtype=float,
+                )
+
+        sparse_trajectory = CartesianTrajectory()
+        sparse_trajectory.base_frame = base_frame
+        sparse_trajectory.tip_frame = ee_frame_name
+        sparse_trajectory.times = [
+            idx * segment_time for idx in range(len(sparse_targets))
+        ]
+        sparse_trajectory.tforms = sparse_targets
```
sea-bass: nit: you can also use syntax like

```python
sparse_trajectory = CartesianTrajectory(
    base_frame=base_frame,
    tip_frame=ee_frame_name,
    times=...,
    tforms=...,
)
```

... more importantly, I missed this in the previous PR, but I think it actually might be preferable if `CartesianTrajectory` actually supported lists of base/tip frames and transforms so you can define a multi-end effector trajectory as one object. Do you think this makes sense?

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -142,51 +207,34 @@ def make_mock_joint_targets(
     return targets
 
 
-def compute_end_effector_positions(
+def compute_end_effector_positions_by_frame(
```
sea-bass: IMO no need to rename this function.

Also, can this just directly accept the `CartesianTrajectory` object now?

muslim-adedamola:
I think I fixed the comments.

let me know what you think.

For some reason, the format/pre-commit test is failing. Is this fault from my end?

sea-bass:
I think the pre-commit failure was a dead GitHub link based on the results, should be fine now.

bindings/src/roboplan/interpolation.py
```diff
@@ -153,9 +157,21 @@ def interpolateCartesianTrajectory(
                 continue
 
             alpha = step / steps_per_segment
-            dense_tforms.append(pin.SE3.Interpolate(start, end, alpha).homogeneous)
+            dense_times.append(trajectory.times[idx] + alpha * segment_time)
+
+            for frame_idx, tforms_for_frame in enumerate(trajectory.tforms):
+                start = pin.SE3(np.asarray(tforms_for_frame[idx]))
+                end = pin.SE3(np.asarray(tforms_for_frame[idx + 1]))
```
sea-bass: as with the previous PR, I think it's more efficient to just convert all the transforms to Pinocchio SE3s up front, and then just index them in the loop. This is basically doubling the conversions to SE3.

bindings/src/roboplan/interpolation.py
```diff
@@ -167,7 +183,12 @@ def interpolateSE3Waypoints(
 
     Prefer interpolateCartesianTrajectory() for timestamped Cartesian trajectories.
     """
-    trajectory = CartesianTrajectory()
-    trajectory.times = waypoint_times
-    trajectory.tforms = transforms
-    return interpolateCartesianTrajectory(trajectory, control_dt)
+    trajectory = CartesianTrajectory(
+        base_frames=[""],
+        tip_frames=[""],
```
sea-bass: empty strings are probably not right... my guess is we probably want to pass the base/tip frame names into this function?

I guess the "proper" thing to do would be to also have a `CartesianPath` type (since there is a `JointPath` too), and that will already have the base/tip frame names that can be copied into the trajectory?

roboplan/include/roboplan/core/types.hpp
```diff
@@ -172,19 +172,32 @@ struct JointTrajectory {
...
+  /// @brief Default constructor.
+  CartesianTrajectory() = default;
 
-  /// @brief The name of the tip (or target) frame.
-  std::string tip_frame;
+  /// @brief Constructor.
+  CartesianTrajectory(std::vector<std::string> base_frames, std::vector<std::string> tip_frames,
+                      std::vector<double> times, std::vector<std::vector<Eigen::Matrix4d>> tforms)
+      : base_frames(std::move(base_frames)), tip_frames(std::move(tip_frames)),
+        times(std::move(times)), tforms(std::move(tforms)) {}
```
sea-bass: these probably should be `const std::vector<T>&`, else they will be copied and that's inefficient

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -57,54 +57,79 @@
...
-    """Create sparse absolute Cartesian target poses as 4x4 transforms.
+) -> CartesianTrajectory:
+    """Create a sparse Cartesian target trajectory for all EE frames.
 
     Args:
-        scene: RoboPlan scene used to compute the starting end-effector pose.
+        scene: RoboPlan scene used to compute starting end-effector poses.
         q_start: Full starting robot configuration.
-        ee_frame_name: End-effector frame name.
+        ee_frame_names: End-effector frame names to create trajectories for.
```
sea-bass:
```suggestion
        ee_frame_names: End effector frame names for the trajectory.

File: bindings/src/modules/core.cpp [hunk 8]
```diff
@@ -126,11 +126,37 @@ void init_core_types(nanobind::module_& m) {
         return ss.str();
       });
 
+nanobind::class_<CartesianPath>(
```
sea-bass: i think the indentation is off here, but the formatter should fix this

bindings/src/roboplan/interpolation.py
```diff
@@ -42,7 +42,8 @@ def interpolateConfigurationWaypoints(
         Dense configuration waypoints sampled approximately every control_dt seconds.
     """
     if len(waypoints) < 2:
-        return waypoints.copy()
+        # Deep-copy to match the behaviour of the interpolated path (fresh arrays).
+        return [np.asarray(w).copy() for w in waypoints]
```
sea-bass: you can just do `[w.copy() for w in waypoints]` since it's already arrays, right?

bindings/src/roboplan/interpolation.py
```diff
@@ -87,6 +88,11 @@ def interpolateJointTrajectory(
     if len(trajectory.positions) < 2:
         return [np.asarray(position).copy() for position in trajectory.positions]
 
+    # Validate all timestamps upfront before doing any interpolation work.
+    for i in range(len(trajectory.times) - 1):
+        if trajectory.times[i + 1] - trajectory.times[i] <= 0.0:
+            raise ValueError("JointTrajectory times must be strictly increasing.")
```
sea-bass: you can just validate in the next loop as you go, it's easier

bindings/src/roboplan/interpolation.py
```diff
@@ -111,63 +114,181 @@ def interpolateJointTrajectory(
...
-        Dense Cartesian transforms sampled according to the trajectory times.
+        Dense Cartesian trajectory sampled according to the sparse waypoint times.
     """
     if control_dt <= 0.0:
         raise ValueError("control_dt must be positive.")
 
-    if len(trajectory.tforms) != len(trajectory.times):
-        raise ValueError(
-            "CartesianTrajectory tforms and times must have the same length."
+    _validate_cartesian_frames_and_tforms(
```
sea-bass: don't worry about making this a helper function yet if it's only used in 1 place here

bindings/src/roboplan/interpolation.py
```diff
@@ -111,63 +114,181 @@ def interpolateJointTrajectory(
...
+        tip_frames=path.tip_frames,
+        times=waypoint_times,
+        tforms=path.tforms,
+    )
 
-    return dense_tforms
+    return interpolateCartesianTrajectory(sparse_trajectory, control_dt)
 
 
 def interpolateSE3Waypoints(
```
sea-bass: it looks like this function isn't being used at all now -- so we can delete it!

roboplan/include/roboplan/core/types.hpp
```diff
@@ -170,21 +170,77 @@ struct JointTrajectory {
...
+  /// @brief Constructor.
+  CartesianPath(
+    const std::vector<std::string>& base_frames,
+    const std::vector<std::string>& tip_frames,
+    const std::vector<std::vector<Eigen::Matrix4d>>& tforms)
+    : base_frames(base_frames)
+    , tip_frames(tip_frames)
+    , tforms(tforms)
+  {
+  }
```
sea-bass: sorry I just realized... if this constructor doesn't do anything besides assign member variables, then there's no need for it! Just like `JointPath` only has a default constructor.

... or is this what enables the use of Python constructors such as:

```python
CartesianPath(base_frames=["frame1"], tip_frames=["frame2"], tforms=[])
```

because I noticed that `JointPath` bindings don't allow this. If that's the case maybe just leave it and we can address this for all the types in a follow-on issue.

roboplan/include/roboplan/core/types.hpp
```diff
@@ -170,21 +170,77 @@ struct JointTrajectory {
...
+/// @brief Constructor.
+CartesianTrajectory(
+    const std::vector<std::string>& base_frames,
+    const std::vector<std::string>& tip_frames,
+    const std::vector<double>& times,
+    const std::vector<std::vector<Eigen::Matrix4d>>& tforms)
+    : base_frames(base_frames),
+      tip_frames(tip_frames),
+      times(times),
+      tforms(tforms) {}
```
sea-bass: indent is also off here, so please run the formatter

roboplan/src/core/types.cpp
```diff
@@ -75,16 +75,66 @@ std::ostream& operator<<(std::ostream& os, const JointTrajectory& traj) {
...
+      os << "  no transforms\n";
+      continue;
+    }
+
+    for (size_t path_idx = 0; path_idx < path.tforms.at(frame_idx).size(); ++path_idx)
+    {
+      os << "  [idx=" << path_idx << "]\n";
+      os << path.tforms.at(frame_idx).at(path_idx) << "\n";
+    }
+  }
```
sea-bass: we aren't standardizing on adding a newline between loops and the opening bracket. Hopefully the formatter also fixes this?

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -143,50 +168,56 @@ def make_mock_joint_targets(
...
-    ee_frame_name: str,
-) -> np.ndarray:
-    """
-    Compute end-effector xyz positions for a sequence of full configurations.
+    trajectory: CartesianTrajectory | None = None,
+    scene: Scene | None = None,
+    configurations: list[np.ndarray] | None = None,
+    ee_frame_names: list[str] | None = None,
+) -> dict[str, np.ndarray]:
+    """Compute end-effector xyz positions from Cartesian targets or configurations.
```
sea-bass: rather than having to validate inputs that can be either `trajectory` or `configurations` / `ee_frame_names`, seems better to just make 2 separate functions to extract positions for the joint vs. Cartesian case.

sea-bass (commented):
Tested and all works great -- nice work! I also gave the code a closer look and have suggestions for consistency.

sea-bass:
I also think the stub generation check is failing because I merged a PR last night and I think you just rebased it correctly, but now need to re-generate the stubs with all the latest versions of the code. Sorry 😬

roboplan/include/roboplan/core/types.hpp
muslim-adedamola: Yes, it is what enables the python constructor style through the nanobindings

sea-bass (approved):
Awesome, thank you!

muslim-adedamola:
yes sure :sunglasses: 
don't know what's wrong but seems i am messing sth up in the pthon stubs even after regenerating them

sea-bass:
They seemed to be whitespace only changes now. I regenerated the stubs on my end.

muslim-adedamola:
oops
just seeing this.
okay, i'll just pause here so i don't change anything again

PR #205, opened 2026-05-14
Support Pinocchio 4.0

### LINKED ISSUE #181: Support Pinocchio 4.0
Pinocchio 4.0.0 was recently released, but for now it's only available on conda-forge.

This has some deprecations, including a full removal of hppfcl headers/namespaces, and needing to use coal instead.

We can probably get reasonably far right away with preprocessor conditionals on the Pinocchio version, which we can readily try with the pixi workflow.

Pinocchio 4 doesn't yet have nanobind ready to go, so that will be for later.

sea-bass:
Closes #181

This PR is mostly pixi lockfile changes.

It updates the pixi workflow to use Pinocchio 4.0.0, but leaves the option to stay compatible with Pinocchio 3.9.0. The ROS buildfarm and PyPi still haven't uniformly transitioned over, so it's good to leave this option.

Note that I found a problem with `toppra` segfaulting in newer versions than 0.6.4, which is super strange but seems to be driven by https://github.com/hungpham2511/toppra/pull/279. Then, when I unpinned the top version of Eigen it all worked? Magical ABI incompatibility stuff happening in conda land.

### DIFF
--- a/roboplan_examples/python/common.py
+++ b/roboplan_examples/python/common.py
@@ -2,7 +2,11 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import List
 
-import hppfcl
+try:
+    import coal
+except ModuleNotFoundError:
+    import hppfcl as coal
+
 import numpy as np
 from numpy.typing import NDArray
 import pinocchio as pin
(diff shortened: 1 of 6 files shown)


roboplan_examples/python/common.py
```diff
@@ -2,7 +2,11 @@
 from pathlib import Path
 from typing import List
 
-import hppfcl
+try:
+    import coal
+except ModuleNotFoundError:
+    import hppfcl as coal
```
eholum, line 8: I love compatibility solutions.

eholum (approved):

PR #206, opened 2026-05-18
Support priorities and nullspace projection in OInK tasks

### LINKED ISSUE #202: Consider adding priorities/nullspace projection to oink
The ConfigurationTask is adversarial to the FrameTask(s). This should really be done in the nullspace to avoid poor tracking.

The same concept extends to other such task combinations.

No idea how involved this would be to implement.

sea-bass:
I decided to give this a try... but this is not my strong suit.

Closes #202


roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
   c.setZero();
 
-  // Calculate accumulated Hessian and Gradient from tasks
-  for (const auto& task : tasks) {
-    auto objective_result = task->computeQpObjective(scene, task_H, task_c);
-    if (!objective_result.has_value()) {
-      return tl::make_unexpected(objective_result.error());
+  // Cumulative nullspace projector and Jacobian stack (only built up if there are 2+ priority
+  // levels — otherwise N stays at identity and nothing changes vs. the original single-level
+  // formulation).
```
sjahr: Nitpick
```suggestion
  // Cumulative nullspace projector and Jacobian stack (only built up if there are 2+ priority
  // levels — otherwise N stays at identity).
```

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
+
+  // Add each task's contribution to H and c, projecting its Jacobian through the cumulative
+  // nullspace from earlier priority levels. Priority 1 has N = I (no projection); subsequent
+  // priorities act only in the nullspace of higher priorities, so their contribution is
+  // structurally zero in the higher-priority directions and cannot fight them.
+  size_t level_start = 0;
+  for (size_t i = 0; i < sorted_tasks.size(); ++i) {
+    Task* task = sorted_tasks[i];
+
+    auto jacobian_result = task->computeJacobian(scene);
```
sjahr: Pre-allocate?

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
 
-  // Calculate accumulated Hessian and Gradient from tasks
-  for (const auto& task : tasks) {
-    auto objective_result = task->computeQpObjective(scene, task_H, task_c);
-    if (!objective_result.has_value()) {
-      return tl::make_unexpected(objective_result.error());
+  // Cumulative nullspace projector and Jacobian stack (only built up if there are 2+ priority
+  // levels — otherwise N stays at identity and nothing changes vs. the original single-level
+  // formulation).
+  Eigen::MatrixXd N = Eigen::MatrixXd::Identity(num_variables, num_variables);
```
sjahr: Pre-allocate

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
+      return tl::make_unexpected("Failed to compute error: " + error_result.error());
     }
 
-    H += task_H;
-    c += task_c;
+    //   min ||W J (N z) + W alpha e||^2   with delta_q = N z (z lives in the priority's
+    //   nullspace). We absorb the parameterization into a projected effective Jacobian and
+    //   keep the optimization variable as dq, so the same QP can be reused regardless of
+    //   priority count.
+    const Eigen::MatrixXd WJ_proj = task->weight * task->jacobian_container * N;
```
sjahr: Same here

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
+      return tl::make_unexpected("Failed to compute Jacobian: " + jacobian_result.error());
+    }
+    auto error_result = task->computeError(scene);
+    if (!error_result.has_value()) {
+      return tl::make_unexpected("Failed to compute error: " + error_result.error());
     }
 
-    H += task_H;
-    c += task_c;
+    //   min ||W J (N z) + W alpha e||^2   with delta_q = N z (z lives in the priority's
```
sjahr: Can you update the math in the docs accordingly?

roboplan_oink/src/optimal_ik.cpp
sjahr: Nitpick (Applies to all new Matrices / Vectors): I prefer verbose names to make it easier to understand what's going on except for H and c because these are the matrix names of the canonical QP form

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
+    //   nullspace). We absorb the parameterization into a projected effective Jacobian and
+    //   keep the optimization variable as dq, so the same QP can be reused regardless of
+    //   priority count.
+    const Eigen::MatrixXd WJ_proj = task->weight * task->jacobian_container * N;
+    const Eigen::VectorXd We = task->weight * (task->gain * task->error_container);
+
+    const double mu = task->lm_damping * We.squaredNorm();
```
sjahr: Why do we need to adjust the damping?

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
+      // damping smoothly preserves SPD-ness of (J J^T + lambda^2 I).
+      const double lambda_sq = regularization;
+      Eigen::MatrixXd JJt = jacobian_stack * jacobian_stack.transpose();
+      JJt.diagonal().array() += lambda_sq;
+      const Eigen::MatrixXd JJt_inv_J = JJt.llt().solve(jacobian_stack);
+      N.setIdentity(num_variables, num_variables);
+      N.noalias() -= jacobian_stack.transpose() * JJt_inv_J;
+
+      level_start = i + 1;
+    }
```
sjahr, line 236: This is a relatively large new block. I wonder if it should be it's own function or at least parts of it to make solveIk clearer

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
     barrier_H_contribution.resize(num_variables, num_variables);
     barrier_c_contribution.resize(num_variables);
   }
-
   for (const auto& barrier : barriers) {
-    auto obj_result =
-        barrier->computeQpObjective(scene, barrier_H_contribution, barrier_c_contribution);
-    if (obj_result.has_value()) {
-      // Add dense contribution to sparse H (convert to sparse for efficient addition)
+    auto r = barrier->computeQpObjective(scene, barrier_H_contribution, barrier_c_contribution);
```
sjahr: Did anything change here?

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
     }
   }
-
   H.makeCompressed();
 
-  // Query total constraint dimensions and cache sizes
-  constraint_sizes.reserve(constraints.size());
+  // Build inequality rows for constraints and barriers, in the original dq space. No
+  // transformation is needed because dq is the decision variable directly.
+  constraint_sizes.clear();
```
sjahr: Just double checking, is this change related, correct and what's the benefit?

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
   int total_constraint_rows = 0;
   for (const auto& constraint : constraints) {
-    int num_rows = constraint->getNumConstraints(scene);
-    constraint_sizes.push_back(num_rows);
-    total_constraint_rows += num_rows;
+    int n = constraint->getNumConstraints(scene);
```
sjahr: Unrelated change?

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
+    constraint_sizes.push_back(n);
+    total_constraint_rows += n;
   }
-
-  // Query total barrier dimensions and cache sizes
-  barrier_sizes.reserve(barriers.size());
+  barrier_sizes.clear();
```
sjahr: Same here

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -204,67 +205,130 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
...
   for (const auto& barrier : barriers) {
-    int num_rows = barrier->getNumBarriers(scene);
-    barrier_sizes.push_back(num_rows);
-    total_barrier_rows += num_rows;
+    int n = barrier->getNumBarriers(scene);
+    barrier_sizes.push_back(n);
+    total_barrier_rows += n;
   }
-
-  // Total inequality constraints = constraints (box) + barriers (one-sided)
```
sjahr, line 260: Keep?

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -274,61 +338,31 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
     last_barrier_rows = total_barrier_rows;
   }
 
-  // Fill constraint matrices block by block
```
sjahr, line 277: Are any changes related to barriers and constraints really necessary for the task priotization?

sjahr (commented):
I am really excited about this! Still need to think a bit about the math but in general this looks good. For now I see 3 things that need to be addressed before we merge this:
- Remove the changes unrelated to this PR for constraints & barriers
- Update the math in the docs to reflect these changes (Claude will be really helpful 😉)
- Clean-up the solveIk function a bit to make the code easier maintainable

roboplan_oink/src/optimal_ik.cpp
sea-bass: Apparently this is an error-dependent damping that is present in Pink: https://github.com/stephane-caron/pink/blob/eca52c023c18444cfc91ad73496a2f7a60acb46e/pink/tasks/task.py#L160

roboplan_oink/src/optimal_ik.cpp
sea-bass: this one probably doesn't need it since I'm sure it's optimized out being in the loop scope, but for the others yes.

roboplan_oink/src/optimal_ik.cpp
sea-bass: yes! I need to do this

sea-bass:
Thanks @sjahr -- indeed, I had Claude walk back some unnecessary changes but it seems like it didn't undo variable renames for some reason. Made it look closer to how it was before.

I also have some higher-level questions about the usability of this:

1. What did you think of the way that priorities are set simply in the options for each specific task?
2. The whole thing about automatically sorting all the tasks by priority every loop doesn't sit well with me, granted sorting a relatively small number of things is fast.

Combining both questions 1. and 2. above, another option might be to have more of a structured way to pass in tasks, e.g., a list of list of tasks, or a dictionary of priorities to tasks.

roboplan_oink/src/optimal_ik.cpp
```diff
@@ -239,17 +270,16 @@ Oink::solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& task
 
   H.makeCompressed();
 
-  // Query total constraint dimensions and cache sizes
-  constraint_sizes.reserve(constraints.size());
+  // Build inequality rows for constraints and barriers, in the original dq space. No
+  // transformation is needed because dq is the decision variable directly.
+  constraint_sizes.clear();
```
sjahr: Nice, makes sense to clear at the start instead at the end! Would it make sense to keep the reserve for a bit more memory efficiency?

sjahr (approved):
I like that the tasks have a priority assigned. From a simple API perspective I think this is also a nice way because list of list or dictionary sounds more complicated to use (but maybe worth the cost :shrug: ). One think that could help is to move the sorting somewho out of the loop e.g. the construction of a custom stack object. That way users who don't want to change the tasks every loop can ensure that the sorting only happens once. 

Maybe a simplification to just 1st order nullspace and then having an API like this could also work:
```
  solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& tasks,
          const std::vector<std::shared_ptr<Constraints>>& constraints,
          Eigen::Ref<Eigen::VectorXd, 0, Eigen::InnerStride<Eigen::Dynamic>> delta_q,
          const std::vector<std::shared_ptr<Task>>& nullspace_tasks
          double regularization = 1e-12);
```
We'd sacrifice nullspace levels for a simpler way to use it. How important do you think are multiple levels of nullspace projection?
Another though is that maybe if we pass these optional nullspace tasks, we could add the sorting there. Either by having a custom struct that sorts the nullspace tasks in the constructor and thus sorting is only necessary when you change something 
```
  solveIk(const Scene& scene, const std::vector<std::shared_ptr<Task>>& tasks,
          const std::vector<std::shared_ptr<Constraints>>& constraints,
          Eigen::Ref<Eigen::VectorXd, 0, Eigen::InnerStride<Eigen::Dynamic>> delta_q,
          const NullspaceTasks& nullspace_tasks
          double regularization = 1e-12);
```
or here we use a diary. That way we could maybe keep the generally simple interface for those who want to just use task weights and still have a multi-level nullspace hierachy.

But in general I'd be fine for now to keep the sort algorithm there. We can optimize more in a follow up. My guess is that many tasks are hard to juggle anyways so I'd guess the number of tasks is most of the time << 10

roboplan_oink/src/optimal_ik.cpp
sea-bass: yes it does. I even timed it with the collision branch and it helped

sea-bass:
Thanks! Let's keep it as is for now then.

I do not like the idea of limiting ourselves to just 2 priority levels, but something like assembling the stack once and freezing the sizes/priorities for runtime iterations could be promising. This is how I originally expected OInK to work anyhow.

I did address some performance things here that came out in the collision barrier PR, since it seemed better to put them in this PR. Namely that the barriers Jacobians/inequalities were being computed 2 times (3 if you count this `enforceBarriers`), so fixing that helped.

sjahr (approved):

PR #207, opened 2026-05-18
OInK self-collision barrier

### LINKED ISSUE #187: Add self-collision barrier to oink solver
This would be a really nice feature add for the solver.

Direct inspiration here: https://github.com/stephane-caron/pink/blob/main/pink/barriers/self_collision_barrier.py

sea-bass:
Closes #187


roboplan_examples/python/example_oink.py
```diff
@@ -119,6 +126,29 @@ def main(
 
     constraints = [position_limit, velocity_limit]
 
+    # Self-collision barrier: keep every collision pair declared in the SRDF at least
+    # `self_collision_d_min` meters apart. Skip when the model has no collision pairs.
+    n_pairs = 8
+    # len(collision_model.collisionPairs)
```
sjahr: Remove?

roboplan_examples/python/example_oink.py
```diff
@@ -196,7 +226,7 @@ def update_goals(_):
     for controls in transform_controls:
         controls.on_update(update_goals)
 
-    tasks = frame_tasks + [config_task]
+    tasks = frame_tasks  # + [config_task]
```
sjahr: Remove config task or keep it?

roboplan_oink/include/roboplan_oink/barriers/self_collision_barrier.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+#include <pinocchio/multibody/geometry.hpp>
+
+#include <roboplan_oink/optimal_ik.hpp>
+
+namespace roboplan {
+
+/// @brief Self-collision avoidance barrier based on hpp-fcl / coal collision pair distances.
+///
+/// Constrains the closest `n_collision_pairs` collision pairs in the scene to remain at least
```
sjahr, line 14: Make sure you add this to the docs page + a note that this can be computationally very costly and might have a significant impact on the solveIK time

roboplan_oink/include/roboplan_oink/barriers/self_collision_barrier.hpp
```diff
@@ -0,0 +1,125 @@
...
+/// `d_min` apart:
+///     h_i(q) = d(p_a, p_b)_i - d_min  for i in 0 .. n_collision_pairs - 1
+/// where the index `i` refers to the i-th closest pair at the current configuration.
+///
+/// The barrier Jacobian row is built from the witness points returned by hpp-fcl and the
```
sjahr: Nit
```suggestion
/// The barrier Jacobian row is built from the witness points returned by coal and the
```

roboplan_oink/include/roboplan_oink/barriers/self_collision_barrier.hpp
```diff
@@ -0,0 +1,125 @@
...
+  /// @param scene The scene containing the robot collision model and current state.
+  /// @return Expected void on success, error message on failure.
+  tl::expected<void, std::string> computeBarrier(const Scene& scene) override;
+
+  /// @brief Compute the Jacobian rows for the closest n_collision_pairs collision pairs.
+  ///
+  /// Assumes computeBarrier() has been called first (it caches the selected pair indices).
+  /// Each row uses the witness points and the parent-joint Jacobians of the two bodies in
+  /// the collision pair. If the witness points coincide, the row is zeroed (Jacobian is
+  /// undefined there). Any NaN values are clamped to zero.
```
sjahr, line 70:
```
 Any NaN values are clamped to zero.
```
Just double checking: Is that correct or dangerous?

roboplan_oink/src/barriers/self_collision_barrier.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <pinocchio/algorithm/kinematics.hpp>
+#include <pinocchio/collision/distance.hpp>
+#include <pinocchio/spatial/skew.hpp>
+
+namespace roboplan {
+
+namespace {
+
+// Skew-symmetric cross-product matrix.
+inline Eigen::Matrix3d skewSymmetric(const Eigen::Vector3d& v) {
```
sjahr: Can we pre-allocate this before so we avoid memory allocation in the main control loop

roboplan_oink/src/barriers/self_collision_barrier.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    const auto j1_id = go_1.parentJoint;
+    const auto j2_id = go_2.parentJoint;
+
+    const Eigen::Vector3d w1 = dr.nearest_points[0];
+    const Eigen::Vector3d w2 = dr.nearest_points[1];
+
+    // When the witness points coincide, the contact normal is undefined.
+    // Skip this row (left at zero) to mirror Pink's behaviour.
+    const Eigen::Vector3d diff = w1 - w2;
```
sjahr: Same here and later on

roboplan_oink/src/barriers/self_collision_barrier.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  // Joint Jacobians are needed for parent-joint Jacobian extraction below.
+  const Eigen::VectorXd& q = scene.getCurrentJointPositions();
+  scene.computeJointJacobians(q);
+  // scene.computeCollisionDistances(q);
+
+  jacobian_container.setZero();
+  for (int i = 0; i < n_collision_pairs; ++i) {
+    const std::size_t k = closest_pair_indices[i];
+    const auto& cp = geom_model.collisionPairs[k];
```
sjahr: Nitpick: Let's make the names more verbose for easier readability

roboplan_oink/src/barriers/self_collision_barrier.cpp
```diff
@@ -0,0 +1,189 @@
...
+tl::expected<void, std::string> SelfCollisionBarrier::computeJacobian(const Scene& scene) {
+  const auto& model = scene.getModel();
+  const auto& data = scene.getData();
+  const auto& geom_model = scene.getCollisionModel();
+  const auto& geom_data = scene.getCollisionData();
+
+  // Joint Jacobians are needed for parent-joint Jacobian extraction below.
+  const Eigen::VectorXd& q = scene.getCurrentJointPositions();
+  scene.computeJointJacobians(q);
+  // scene.computeCollisionDistances(q);
```
sjahr: Remove?

roboplan_oink/src/barriers/self_collision_barrier.cpp
```diff
@@ -0,0 +1,189 @@
...
+}
+
+tl::expected<double, std::string>
+SelfCollisionBarrier::evaluateAtConfiguration(const pinocchio::Model& model, pinocchio::Data& data,
+                                              const Eigen::VectorXd& q) const {
+  if (collision_model == nullptr) {
+    return tl::make_unexpected("SelfCollisionBarrier: collision_model pointer is null");
+  }
+
+  // TODO: Do not duplicate
```
sjahr: The one above is commented out I think. Is this TODO done?

sjahr (commented):
I tested it and it works 😍  Super cool, some minor cleanups necessary + an update for the docs and then let's go!

roboplan_examples/python/example_oink.py
sea-bass: Yep, I will re-add it as priority 2 once I rebase on that branch

roboplan_oink/include/roboplan_oink/barriers/self_collision_barrier.hpp
sea-bass: This again comes directly from Pink:

https://github.com/stephane-caron/pink/blob/eca52c023c18444cfc91ad73496a2f7a60acb46e/pink/barriers/self_collision_barrier.py#L219-L222

roboplan_oink/src/barriers/self_collision_barrier.cpp
sea-bass: yeah this is more about `computeBarrier` already calling `scene.computeCollisionDistances` so I'd rather reuse a member variable. Fixing this now

roboplan_examples/python/example_dual_arm.py
sea-bass, line 1: this example was useless so I drive-by removed it

docs/source/concepts/inverse_kinematics.rst
```diff
@@ -445,44 +505,81 @@ Usage Example
 .. code-block:: python
 
    import numpy as np
-   from roboplan_ext import Scene
-   from roboplan_ext.optimal_ik import (
-       Oink, FrameTask, FrameTaskOptions,
-       VelocityLimit, PositionLimit, PositionBarrier
+   from roboplan.core import Scene, CartesianConfiguration
```
sea-bass, line 508: note: this usage example was outdated since my previous oink refactor oops

sjahr (commented):
Thanks! The code looks great now. One thing I noticed is that it looks like the performance is significantly worse compared to yesterday :thinking: Any idea what could be the issue?

https://github.com/user-attachments/assets/176f4b17-2a0a-4499-a1dd-ebf588e42c6b

It looks like there is now a delay for the frame task being tracked that hasn't been there before. Does it behavior like this in your test too? I am using 2 collision pairs.

```
python3 roboplan_examples/python/example_oink.py --model dual --self_collision_num_pairs 2
```

sea-bass:
Yeah, this is pretty annoying, but actually the multi-priority problem (because I rebased this branch) is making it slower. If you remove the `config_task` from the list of tasks, you'll find it speeds up considerably... but is still slow.

The real problem is that the collision geometries aren't very optimized for fast collision checking in these examples.

If you run the single-arm Franka though, it should perform fine. This is what I'm seeing at least.

sea-bass:
Actually, I found that the collision distances were being computed twice per loop (3 times if you count calls to `enforceBarriers` -- I just halved the solve time with a quick caching in [this commit](https://github.com/open-planning/roboplan/pull/206/changes/cd59da85ca0f5899a6ec42cb7398fd365cd218ee), cleaned up to [this commit](https://github.com/open-planning/roboplan/pull/206/commits/3c49b3cde43d9ac0fcc1ecfe1e8a6f9583945e5d). Try now?

On my machine, the dual with 4 collision pairs has a ~17 ms solve time, which is still slower than the loop rate but it feels mostly responsive. If you change control_freq to 50 hz this could be fine.

[oink_collision_dual_50hz.webm](https://github.com/user-attachments/assets/3f373688-bd62-4a2b-9514-91b8c260b1a8)

sjahr (approved):
Nice work thank you! Let’s merge this. I agree that the performance bottleneck for the collision checking is the geometric collision object representation. We can look into adding spherical models in a follow up PR

PR #208, opened 2026-05-19
Add Stretch4 model

olaghattas:
Adds stretch4 robot by hello robot (URDF/SRDF/meshes, common.py, mobile base as 3 DOF joints for RRT)  as an example robot.

Fix1 : applyMimics in IK so telescoping segments stay synced.
arm_parts_disconnect

Fix2: 
clampJointLimits so arm extension stays within URDF bounds.

The links get disconnected

arm_extends


roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -111,6 +111,8 @@ bool SimpleIk::solveIk(const std::vector<CartesianConfiguration>& goals,
       }
 
       q = pinocchio::integrate(model, q, vel_ * options_.step_size);
+      scene_->clampJointLimits(q);
```
sea-bass: I'm not entirely sure it's a good idea to clamp the joints during integration, as sometimes an IK solver exceeds them before snapping back.

However, I can see how this would help with something with a floating base with unbounded joints at the base... hmm...

roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -111,6 +111,8 @@ bool SimpleIk::solveIk(const std::vector<CartesianConfiguration>& goals,
...
+      scene_->applyMimics(q);
```
sea-bass, line 125: Can this just be called once when returning a solution?

sea-bass (commented):
Thank you for this! Left a few comments.

One other major one: I've been meaning to better test actual planar URDF joints in this tool, so instead of making the URDF's base 3 separate 1 DOF joints, do you have any interest in switching to planar?

There will be some support needed to make this all work, so we could also do it in a follow-on PR and I can help with that.

roboplan_simple_ik/src/simple_ik.cpp
olaghattas: good point, i now only call this after convergence

roboplan_simple_ik/src/simple_ik.cpp
olaghattas: The clamp part yes, but the applyMimics has to be applied per-iteration because for the telescopic arm, only the leader joint is in the IK group so FK during the loop needs synced mimics, not just at the end. The robot started glitching when i only applied it after convergence.

olaghattas:
Hey Sebastian,

I did the 1dof thing because RRT doesnt support planar joints but i am down to switch to planar. Do you have any ideas of a good way to do that?

sea-bass:
> I did the 1dof thing because RRT doesnt support planar joints but i am down to switch to planar. Do you have any ideas of a good way to do that?

I do, but I don't want to make you do it as it requires a little familiarity with the dynotree library for setting up search spaces. So if you want to switch the URDF to a planar joint I can do the RRT/TOPPRA parts (I imagine TOPPRA will need some tweaks too). We can do this here or in a separate PR as you prefer.

roboplan_simple_ik/src/simple_ik.cpp
sea-bass: Makes sense, thanks!

Although the new place for the clamp may not be correct since you could converge to a solution, then after clamping it may no longer have FK that is within tolerance.

Maybe it is best to keep the clamping behavior in the iterations after all... hmmm

olaghattas:
Sounds good, let's do this in a separate PR. I will change the URDF and open a new PR and we can continue there.

roboplan_example_models/models/stretch4_robot_model/meshes/stretch4_sg4.urdf
sea-bass, line 1: there seems to be a copy of the URDF here. Make sure there is only one!

roboplan_examples/python/common.py
```diff
@@ -381,6 +381,67 @@ def get_model_data():
                 ),
             ],
         ),
+        "stretch4_sg4": RobotModelConfig(
```
sea-bass: can we just call this "stretch" or "stretch4" to be easier to write?

sea-bass (commented):
Hey @olaghattas -- I replaced the joint position clamping with something more efficient that extracts the joint info once and clamps at each iteration.

I also tried to put some breakpoints in the Viser visualizer and didn't find anything obvious there to fix the materials stuff... sad.

Besides that, just a few comments.

sea-bass:
there also seems to be a bug with the RRT implementation here...

[Screencast from 2026-05-19 07-49-48.webm](https://github.com/user-attachments/assets/62a949b6-5586-49e2-a2d7-ad8649c82cba)

roboplan_examples/python/common.py
olaghattas: there is another gripper that i want to support that is why i added the sg to differentiate them.

roboplan_example_models/models/stretch4_robot_model/meshes/stretch4_sg4.urdf
olaghattas: I am sorry i added that to view the urdf in the online urdf viewer https://gkjohnson.github.io/urdf-loaders/javascript/example/bundle/ and forgot to remove it.

sea-bass (approved):
Narrowed down the RRT issue to pixi doing weird stuff with ABI breakages in Eigen. After a few more fixes, this seems to work here now.

PR #209, opened 2026-05-19
Support planar joints

### LINKED ISSUE #186: Add a mobile manipulator example model
I think our example models should also have a robot with a 2D (planar) mobile base, and either 1 or 2 arms. One with a linear lift joint or something would be really cool too, but not necessary.

This issue may come with some bonus things like making sure our RRT, IK, etc. do support planar joints. They should with relative ease, since we already support continuous joints.

olaghattas:
Used the planar joint instead of 3 1 DOF joints.
The scene.cpp had to be adjusted. The changes add a dedicated planar branch that samples x and y within finite bounds (with a ±2 m fallback), samples yaw as an angle and writes cos/sin so rotation stays on the unit circle, and fix the generic loop to write positions(q_idx + idx) so multi-DOF joints use consecutive entries in q. Together, that makes random pose generation valid again so randomCollisionFreePositions() can succeed and example_ik no longer crashes on startup.

The error i got without the changes:
  File "/home/olagh48652/.local/lib/python3.12/site-packages/tyro/_cli.py", line 282, in cli
    out = run_with_args_from_cli()
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/olagh48652/roboplan_ws/install/roboplan_examples/lib/roboplan_examples/example_ik.py", line 150, in main
    randomize_position(None)
  File "/home/olagh48652/roboplan_ws/install/roboplan_examples/lib/roboplan_examples/example_ik.py", line 142, in randomize_position
    q_rand = scene.randomCollisionFreePositions()[q_indices]
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable


sea-bass:
@olaghattas I'm gonna merge your other PR now; when that is done, would you mind rebasing this?

I'm gonna be chasing down the Eigen 5.0.1 issue for now.

olaghattas:
I will do the rebasing now!

sea-bass:
> I will do the rebasing now!

Thanks! You can change the model name back to `stretch4_sg4` -- I realized my revert of that didn't make it into main

olaghattas:
i kept the name as stretch for now, i will change it back to stretch4_sg4 once i add the other gripper

sea-bass (approved):
Got the IK and RRT working with planar joints!

https://github.com/user-attachments/assets/70f4f330-009e-47f9-a9c8-3ce3bd2cc6ec

Just left you one very minor comment, that doesn't affect roboplan but may affect you elsewhere.

roboplan_example_models/models/stretch4_robot_model/stretch4_sg4.srdf
```diff
@@ -12,9 +10,7 @@
   </group>
 
   <group_state group="stretch4_arm" name="home">
-    <joint name="base_x_joint" value="0.0"/>
-    <joint name="base_y_joint" value="0.0"/>
-    <joint name="base_yaw_joint" value="0.0"/>
+    <joint name="mobile_base_planar_joint" value="0.0"/>
```
sea-bass: Does this value still work as a scalar given it's a 3 DOF joint? We don't use this in roboplan (yet), but just curious.

Issue #210, opened 2026-05-19
Viser upgrade past 1.0.24 breaks RRT visualization


sea-bass:
Making myself an issue here to keep track

sea-bass:
Update: This actually seems to not be an Eigen issue, but is actually something that changed on the Viser side. If you refresh the browser window, all the lines show up correctly.

However, of course we shouldn't need to refresh every time we plan a new path.

The issue is Viser 1.0.24 to 1.0.26+: See https://github.com/viser-project/viser/issues/719

sea-bass:
Seems like the maintainers fixed the issue in a Viser 1.0.29 release. When that's out on conda-forge we can use that as the lower pin.

PR #211, opened 2026-05-19
Pin viser to below 1.0.25

sea-bass:
Works around (but doesn't solve) https://github.com/open-planning/roboplan/issues/210


sea-bass:
@olaghattas FYI 🤦🏻

eholum (approved):
And it's only pixi? https://github.com/open-planning/roboplan/blob/main/.docker/ubuntu/Dockerfile#L24

sea-bass:
Good catch, I just tested the ROS Dockerfile and same deal.

PR #213, opened 2026-05-20
Keyboard teleoperation example

### LINKED ISSUE #149: Add teleop to roboplan examples
Providing a way to teleoperate a (simulated) robot using the keyboard, devices such as gamepads or 3D mice, or leader-follower systems.

## Desired output

A runnable teleop example that can move the UR and Franka example robots (end-effector jog) via:

- keyboard input (baseline, always supported)
- controller input (Playstation / Xbox)
- 3D mouse input (SpaceMouse)

## Proposed way to test it 

- roboplan_examples/python/example_teleop.py --model ur5 --device keyboard
- roboplan_examples/python/example_teleop.py --model franka --device keyboard
- later: --device gamepad, --device spacemouse, etc.

## Keyboard controls

Click/focus the running terminal window first.

**Translation:**

- w / s → +x / −x
- a / d → +y / −y
- q / e → +z / −z

**Rotation:**

- i / k → +roll / −roll
- j / l → +pitch / −pitch
- u / o → +yaw / −yaw

**Other:**

- space → pause/resume teleop
- r → reset robot to home position
- t → reset target to current end-effector pose
- x → quit teleop

## GUI controls

- Enable Teleop checkbox
- Pause / Resume button
- Linear Sensitivity slider
- Angular Sensitivity slider
- Control Frame dropdown (ee / world)
- Reset to Home button
- Reset Target to Current button

muslim-adedamola:
PR update for #149 and #150

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -248,6 +248,7 @@ compiler-rt = ">=19.1.7,<20"
 # Python test dependencies
 pytest = ">=6"
 pytest-benchmark = ">=5.1.0,<6"
+pynput = ">=1.8.2,<2"
 
 [pypi-dependencies]
 pycollada = ">=0.9.2, <0.10"
--- a/pixi.toml
+++ b/pixi.toml
@@ -227,6 +227,7 @@ tyro              = ">=0.9.26,<0.10"
 viser             = ">=1.0.21,<1.0.25"  # See https://github.com/viser-project/viser/issues/719
 zstandard         = ">=0.25.0,<0.26"
 plyfile           = ">=1.1,<2"
+pynput            = ">=1.8.2,<2"
 
 # We use pip as a workaround for pixi building the bindings before we build the core packages
 # In theory, it should be like this:
@@ -248,7 +249,6 @@ compiler-rt = ">=19.1.7,<20"
 # Python test dependencies
 pytest = ">=6"
 pytest-benchmark = ">=5.1.0,<6"
-pynput = ">=1.8.2,<2"
 
 [pypi-dependencies]
 pycollada = ">=0.9.2, <0.10"
--- a/roboplan_examples/python/common.py
+++ b/roboplan_examples/python/common.py
@@ -476,3 +476,44 @@ def get_octree():
         tform=pin.SE3(np.eye(3), np.array([0.0, 0.0, 0.0])).homogeneous,
         color=np.array([0.251, 0.878, 0.816, 1.0]),
     )
+
+
+def se3_to_viser_wxyz(transform: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+    """Extract Viser-compatible position and wxyz quaternion from an SE(3) matrix.
+
+    Args:
+        transform: 4x4 homogeneous transform.
+
+    Returns:
+        Tuple of ``(position, wxyz)`` as numpy arrays.
+    """
+    position = transform[:3, 3].copy()
+    quat = pin.Quaternion(transform[:3, :3])
+    wxyz = np.array([quat.w, quat.x, quat.y, quat.z])
+    return position, wxyz
+
+
+def get_home_configuration(
+    scene: Scene,
+    model_data: RobotModelConfig,
+) -> np.ndarray:
+    """Return the home configuration for the selected model.
+
+    Args:
+        scene: Scene used to read current joint positions.
+        model_data: Robot model configuration.
+
+    Returns:
+        Full starting joint configuration vector.
+    """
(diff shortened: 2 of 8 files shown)


pixi.toml
```diff
@@ -248,6 +248,7 @@ compiler-rt = ">=19.1.7,<20"
 # Python test dependencies
 pytest = ">=6"
 pytest-benchmark = ">=5.1.0,<6"
+pynput = ">=1.8.2,<2"
```
sea-bass: will we also need to install this in the other Docker containers under `.docker`?

also, can you move this here away from the test dependencies to whether the example deps go?

roboplan_examples/python/example_teleop.py
```diff
@@ -0,0 +1,634 @@
...
+    Args:
+        transform: 4x4 homogeneous transform.
+
+    Returns:
+        Tuple of ``(position, wxyz)`` as numpy arrays.
+    """
+    position = transform[:3, 3].copy()
+    quat = pin.Quaternion(transform[:3, :3])
+    wxyz = np.array([quat.w, quat.x, quat.y, quat.z])
+    return position, wxyz
```
sea-bass: should this be moved to a more general location / I also assume the same stuff is used in other examples

roboplan_examples/python/example_teleop.py
```diff
@@ -0,0 +1,634 @@
...
+    q_home = np.array(model_data.starting_joint_config)
+
+    if len(q_home) == len(q_full):
+        return q_home.copy()
+
+    print(
+        f"Warning: starting_joint_config size ({len(q_home)}) does not match "
+        f"model configuration size ({len(q_full)}). Using scene default instead."
+    )
+    return q_full.copy()
```
sea-bass: same here re: central utility since I believe you made this for the action chunking example. Maybe this can just go in `common.py`

sea-bass (commented):
Looks nice and simple! I haven't tested yet, but some initial comments.

One question: I know that in the regular OInK example there is use of the SE3 low-pass filtering utility to smooth out commands. Did you find that the robots looked jerky at all / should we incorporate that as an option? The OInK example simply accepts a CLI argument for the filter coefficient, which if 0 is not used.

muslim-adedamola:
I actually noticed some jerkiness sometimes especially with the Franka arms. I think it would be nice to add the low-pass utility then. Will do that.

roboplan_examples/python/example_teleop.py
muslim-adedamola: will do

pixi.toml
muslim-adedamola: I checked the docker files.
i don't think any dockerfile update is needed

sea-bass: As in, the docker containers already both have pynput available?

muslim-adedamola: Sorry, fixed it with a new push

roboplan_examples/python/common.py
```diff
@@ -421,3 +421,44 @@ def get_octree():
...
+    Args:
+        transform: 4x4 homogeneous transform.
+
+    Returns:
+        Tuple of ``(position, wxyz)`` as numpy arrays.
+    """
+    position = transform[:3, 3].copy()
+    quat = pin.Quaternion(transform[:3, :3])
+    wxyz = np.array([quat.w, quat.x, quat.y, quat.z])
+    return position, wxyz
```
sea-bass: nit: this one is probably best in `visualization.py`

muslim-adedamola:
Did this work as intended?

sea-bass (approved):
Sorry, I was waiting for a review re-request -- just tested and works great!

sea-bass:
Oh, I do have one comment (sorry I forgot) -- for the new utilities you created, can you modify your action chunking example to use them as well?

muslim-adedamola:
> Oh, I do have one comment (sorry I forgot) -- for the new utilities you created, can you modify your action chunking example to use them as well?

Yes, sure. Will hop on that

muslim-adedamola:
> Oh, I do have one comment (sorry I forgot) -- for the new utilities you created, can you modify your action chunking example to use them as well?

Should i include them in this PR as well?

sea-bass:
Yes please!

muslim-adedamola:
Done.
So I checked and only the home configuration utility from common.py had to be updated in the action_chunk script

sea-bass:
The Ubuntu 22.04 build failure is also happening on `main`, nothing to do with the PR. So I'm gonna go ahead and merge this. Thanks!

PR #214, opened 2026-05-21
Use native mimic joint functionality in Pinocchio

olaghattas:
addresses the OInK solver overusing the base instead of moving the telescoping arm by enabling the mimic in Pinocchio instead of having two different models.

before 

https://github.com/user-attachments/assets/fd06c7c9-477e-4b28-96c8-493bead4eac7

after 

https://github.com/user-attachments/assets/62fc427e-6196-4a29-a1a5-ab8caa50eb8d

### DIFF
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -869,4 +869,3 @@ std::ostream& operator<<(std::ostream& os, const Scene& scene) {
 }
 
 }  // namespace roboplan
-
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -72,7 +72,7 @@ Scene::Scene(const std::string& name, const std::string& urdf, const std::string
   size_t q_idx = 0;
   size_t v_idx = 0;
   joint_names_.reserve(model_.njoints - 1);
-  actuated_joint_names_.reserve(model_.njoints - 1);
+  actuated_joint_names_.reserve((model_.njoints - 1) - model_.mimicking_joints.size());
   for (int idx = 1; idx < model_.njoints; ++idx) {  // omits "universe" joint.
     const auto& joint_name = model_.names.at(idx);
     joint_names_.push_back(joint_name);
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -385,6 +385,17 @@ Eigen::VectorXd Scene::interpolate(const Eigen::VectorXd& q_start, const Eigen::
   return pinocchio::interpolate(model_, q_start, q_end, fraction);
 }
 
+void Scene::setJointPositions(const Eigen::VectorXd& positions) {
+  if (positions.size() != model_.nq) {
+    throw std::invalid_argument("setJointPositions: expected " + std::to_string(model_.nq) +
+                                " configuration values (model.nq), got " +
+                                std::to_string(positions.size()) +
+                                ". For robots with mimic joints, use the mimic-enabled model "
+                                "layout (not the expanded URDF DOF count).");
+  }
+  cur_state_.positions = positions;
+}
+
 Eigen::VectorXd Scene::integrate(const Eigen::VectorXd& q, const Eigen::VectorXd& v) const {
   return pinocchio::integrate(model_, q, v);
 }
@@ -462,6 +473,10 @@ Scene::getPositionLimitVectors(const std::string& group_name, const bool collaps
                                  maybe_joint_info.error());
     }
     const auto& joint_info = maybe_joint_info.value();
+    if (joint_info.mimic_info) {
+      // Mimic joints have nq=0; they do not occupy slots in the configuration vector.
+      continue;
+    }
 
     switch (joint_info.type) {
     case JointType::FLOATING:
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -56,9 +56,8 @@ Scene::Scene(const std::string& name, const std::string& urdf, const std::string
     package_paths_str.push_back(std::string(path));
   }
 
-  // Start by creating a Pinocchio model with mimic support to parse all the relationships.
-  pinocchio::Model mimic_model;
-  pinocchio::urdf::buildModelFromXML(urdf, mimic_model, /*verbose*/ false, /*mimic*/ true);
+  // Single model with Pinocchio native mimics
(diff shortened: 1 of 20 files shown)


roboplan_examples/python/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ install(
   example_dual_arm.py
   example_ik.py
   example_rrt.py
+  example_oink.py
```
sea-bass, line 7: Good catch. Can you add the oink with barriers example too?

roboplan_examples/python/common.py
```diff
@@ -395,27 +395,25 @@ def get_model_data():
             ee_names=["grasp_center_link"],
             base_link="base_link",
             starting_joint_config=[
-                # Planar base (x, y, cos(yaw), sin(yaw)).
+                # nq=17 with Pinocchio mimic joints (arm_l2/l3/l4 collapsed into arm_l1).
```
sea-bass, line 393: Great, but I feel like the kinova and maybe franka configs will need updating too? We'll need to check a lot of examples

roboplan_examples/python/example_oink.py
```diff
@@ -127,7 +127,7 @@ def main(
             f"configuration space dimension ({len(q_full)}), using current scene positions instead"
         )
         with scene_lock:
-            q_canonical = scene.getCurrentJointPositions()
+            q_canonical = np.array(scene.getCurrentJointPositions())
```
sea-bass: Is the np.array constructor necessary here?

sea-bass (commented):
This is great! We'll need to comb through the examples pretty carefully to check that things work, and the RRT benchmark seems to be failing (probably something minor)... but I'm glad to see this cleanup.

roboplan_examples/python/common.py
olaghattas: yeah I am looking into that. they currently run but because they are using the fallback

olaghattas: should be resolved in commit update models to remove mimic joints

roboplan/src/core/scene.cpp
```diff
@@ -72,13 +71,13 @@ Scene::Scene(const std::string& name, const std::string& urdf, const std::string
   // Create additional robot information.
   size_t q_idx = 0;
   size_t v_idx = 0;
-  joint_names_.reserve(mimic_model.njoints - 1);
-  actuated_joint_names_.reserve(mimic_model.njoints - mimic_model.mimicking_joints.size() - 1);
-  for (int idx = 1; idx < mimic_model.njoints; ++idx) {  // omits "universe" joint.
-    const auto& joint_name = mimic_model.names.at(idx);
+  joint_names_.reserve(model_.njoints - 1);
+  actuated_joint_names_.reserve(model_.njoints - model_.mimicking_joints.size() - 1);
```
sea-bass: Do we need this variable anymore?

roboplan_rrt/src/rrt.cpp
```diff
@@ -24,15 +24,18 @@ RRT::RRT(const std::shared_ptr<Scene> scene, const RRTOptions& options)
     throw std::runtime_error("Failed to instantiate RRT planner: " + maybe_collapsed_pos.error());
   }
 
-  const auto joint_names = scene_->getJointNames();
   std::vector<std::string> state_space_names;
-  state_space_names.reserve(joint_names.size());
+  state_space_names.reserve(joint_group_info_.joint_names.size());
```
sea-bass, line 28: Nice catch!

roboplan_rrt/src/rrt.cpp
```diff
@@ -220,9 +230,17 @@ std::optional<JointPath> RRT::joinTrees(const std::vector<Node>& nodes, const Kd
...
   const auto& q_last_added = last_added_node.config;
 
-  // Find the nearest node in the target tree.
+  // Find the nearest node in the target tree (search uses collapsed coordinates).
   const auto& q_indices = joint_group_info_.q_indices;
-  const auto& nn = target_tree.search(q_last_added(q_indices));
+  const auto maybe_q_collapsed =
+      collapseContinuousJointPositions(*scene_, options_.group_name, q_last_added(q_indices));
+  if (!maybe_q_collapsed) {
+    return std::nullopt;
```
sea-bass: In this and the next case, should we just throw an exception rather than returning nullopt? This would only fail in the event of a bug, not a planning failure.

sea-bass (commented):
This is looking good, and I will test later.

Just one more comment really, thinking more about the connection to ROS: Should we keep around a way to get the full expanded out configuration vector including the mimic joints? I hope Pinocchio has this information available, but if not then this is a possible way to repurpose the `applyMimics` function.

roboplan/src/core/scene.cpp
olaghattas: roboplan doesnt need it but roboplan_ros still uses it. I guess once we adjust roboplan_ros we can get rid of it.

sea-bass: I would say we can remove it right away and fix the `roboplan-ros` repo after. I assume we would just be able to call `getJointNames()` instead wherever we use that.

olaghattas: sounds good, i will do that.

roboplan/src/core/scene.cpp
```diff
@@ -72,13 +71,13 @@ Scene::Scene(const std::string& name, const std::string& urdf, const std::string
   // Create additional robot information.
   size_t q_idx = 0;
   size_t v_idx = 0;
-  joint_names_.reserve(mimic_model.njoints - 1);
-  actuated_joint_names_.reserve(mimic_model.njoints - mimic_model.mimicking_joints.size() - 1);
-  for (int idx = 1; idx < mimic_model.njoints; ++idx) {  // omits "universe" joint.
-    const auto& joint_name = mimic_model.names.at(idx);
+  joint_names_.reserve(model_.njoints - 1);
+  actuated_joint_names_.reserve(model_.njoints - 1);
```
sea-bass: Commented in the wrong place. This is over-reserving memory if there are mimics.

You may be able to subtract the size of the model's `mimicking_joints` member variable from here?

sea-bass:
All looks good, except I found one more possible issue.

In the `example_oink.py`, I noticed that the base was able to move around pretty readily. Because joints in URDF Just have one limit, these are probably getting ignored. So I threw in a small change to provide these limits from YAML config.

sea-bass (approved):

PR #215, opened 2026-05-22
Add Ubuntu 26.04 and ROS 2 Lyrical to CI

sea-bass:
This required a change to the ROS container's requirements file since Ubuntu 26.04 / ROS 2 Lyrical finally updates to NumPy 2.x and also has an incompatibility with `typing_extensions` pin for older distros.


sea-bass:
Of course it's never easy... osqp version being outdated strikes back: https://github.com/tier4/osqp_vendor/issues/26

And `tl_expected` was removed, which is fine: https://github.com/PickNikRobotics/generate_parameter_library/pull/355

For now, found a few workarounds and we shall move forward

.docker/ros/ros_python_requirements_humble.txt
```diff
@@ -0,0 +1,8 @@
+matplotlib
+nanobind
```
sea-bass: TODO: resolve conflicts by removing `nanobind` + updating Viser versions in all these

sea-bass:
BTW the rolling CI job is expected to fail since we're in a transition from Ubuntu 24.04 to 26.04 and the Docker image isn't yet up to date.

PR #216, opened 2026-05-22
Fix TOPPRA issues when using planar joints

sea-bass:
Found an issue in which TOPPRA wasn't doing the right thing with planar joints because of Lie group (SE2) interpolation in Pinocchio/dynotree vs. individual DOF in toppra. So we can "fix" it by more densely sampling in the planar joint case.

This looks bad if you use the Hermite TOPPRA mode (lots of stop and go knots), so I changed the default mode to Adaptive which is generally good to do anyway.

Additionally, it fixes a bug in which the continuous joint wraparounds were being done using joints, and not position DOFs, as indices.

image

### DIFF
--- a/roboplan_examples/python/example_rrt.py
+++ b/roboplan_examples/python/example_rrt.py
@@ -230,7 +230,7 @@ def main(
             )
         else:
             visualizeJointTrajectory(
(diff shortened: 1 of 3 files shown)


sjahr (approved):
Looks good! If it works, let's merge it 🚀

PR #219, opened 2026-05-29
Add toppra and nanobind-dev rosdep keys

sea-bass:
On some ROS distros (really, Ubuntu versions), the `nanobind-dev` dependency is too old to have stub generation. But we can simply skip it with a warning if that's the case to allow for buildfarm release, at least.

`toppra` is already on the buildfarm, though at the time of creating this PR has only been released on Humble/Lyrical since other distros haven't had syncs yet.


eholum (approved):
we can live with this, right?

sea-bass:
Yes. People generating stubs are being pointed to pixi anyway. Or you can pip install nanobind manually as a developer.

PR #221, opened 2026-05-30
Modularize Python bindings

### LINKED ISSUE #220: Modularize Python bindings
Right now, there is a single `roboplan_bindings` package that will continue to be a bottleneck as the number of packages increases.

There is appeal to have each package include its own bindings, all guarded by a `BUILD_PYTHON_BINDINGS` (or similar) build flag.

This will of course be a breaking change, but well worth trying before a potential 1.0.0.

sea-bass:
Big PR, but has upsides for scaling up -- both build times, and for people to not need to build every roboplan package to use Python bindings.

I wanted to get this out before a release that works with the ROS buildfarm so we don't end up renaming packages etc.

This does of course mean changing the conda packaging too, but it seems to work: https://github.com/conda-forge/roboplan-feedstock/pull/15

Closes #220


roboplan/bindings/python/roboplan/visualization.py
```diff
@@ -1,11 +1,19 @@
...
 import pinocchio as pin
 from pinocchio.visualize import ViserVisualizer
 
 from roboplan.core import Scene, computeFramePath, JointPath, JointTrajectory
-from roboplan.rrt import RRT
+
+if TYPE_CHECKING:
+    # Only needed for type checking; visualizeTree() uses the planner via duck
+    # typing, so roboplan-core does not depend on roboplan-rrt at runtime.
+    from roboplan.rrt import RRT
```
sea-bass: possibly want to move this utility to the `roboplan_rrt` bindings

.docker/ros/Dockerfile
```diff
@@ -33,7 +33,7 @@ COPY . src/roboplan
 # Install dependencies and build RoboPlan.
 RUN source /opt/ros/${ROS_DISTRO}/setup.bash \
     && rosdep install --from-paths src --ignore-src --rosdistro ${ROS_DISTRO} -y \
-    && colcon build --cmake-args -DBUILD_TESTING=ON
+    && colcon build --cmake-args -DBUILD_TESTING=ON -DGENERATE_PYTHON_STUBS=OFF
```
sea-bass: I disabled stub generation exclusively because there was an issue with Humble. I will keep looking into this to see if I can fix it instead of disabling it for all distros.

roboplan_oink/README.md
```diff
@@ -1,19 +1,14 @@
...
 
-## TODO
+The OInK solver uses Quadratic Programming (QP) to compute joint displacements that achieve multiple objectives while respecting constraints and safety barriers.
 
-- Support floating base robots (`nq != nv`) in position limits
-- Hierarchical task prioritization via null-space projection
-- Adaptive LM damping near singularities
-- Constraint scaling for mixed constraint types
-- Self-collision avoidance barrier
-- Singularity avoidance barrier
```
sea-bass, line 10: @sjahr removed this because between already-merged and in-progress PRs, I think we have basically all of them except the constraint scaling. Plus, this stuff is better suited for Git issues

roboplan/CMakeLists.txt
```diff
@@ -138,6 +138,13 @@ install(TARGETS roboplan
         INCLUDES DESTINATION include
         )
 
+# Build the Python bindings by default (for the scikit-build wheel, colcon, and
+# pixi). Override with -DBUILD_PYTHON_BINDINGS=OFF for a pure C++ build.
```
eholum, line 142: I guess a pure C++ build would be optimal? Though rare given our first class python bindings?

roboplan_rrt/bindings/python/roboplan/rrt/__init__.py
eholum, line 1: ooh yeah good change. I like keeping utils alongside their relevant packages!

eholum (commented):
This is a good (and painful change)! 

I'm assuming that like 90% of it is moving code? I took a close look at the updated CMakeLists for each of the bindings packages and that looks good! Is there any other piece of this that you have major concerns with? Or should we trust test coverage and yolo this and fix follow ons in post?

sea-bass:
> I'm assuming that like 90% of it is moving code?

Definitely. But I did test all the build workflows, including building the conda packages locally, to be safe. After much pain, it seems to work fine? 🤞🏻

roboplan/CMakeLists.txt
sea-bass: I was in fact speaking to an old colleague yesterday who was interested in a no-ROS. no-Python, C++ only build. Which may have precipitated this whole thing.

eholum (approved):
Assuming you like this then you have to approve this.

https://github.com/open-planning/roboplan-ros/pull/42

PR #222, opened 2026-05-30
Pin viser to 1.0.29 or newer

### LINKED ISSUE #210: Viser upgrade past 1.0.24 breaks RRT visualization

sea-bass:
Pushed the Viser version to conda-forge in https://github.com/conda-forge/viser-feedstock/pull/32, which lets us update across the board.

Closes #210

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -223,10 +223,12 @@ xacro             = ">=2.1.1,<3"
 
 # Dependencies for examples
 matplotlib        = ">=3.10.5,<4"
+nodeenv           = ">=1.10.0, <2"
 tyro              = ">=0.9.26,<0.10"
-viser             = ">=1.0.21,<1.0.25"  # See https://github.com/viser-project/viser/issues/719
+viser             = ">=1.0.29,<2.0"  # See https://github.com/viser-project/viser/issues/719
 zstandard         = ">=0.25.0,<0.26"
 plyfile           = ">=1.1,<2"
+pycollada         = ">=0.9.2, <0.10"
 pynput            = ">=1.8.2,<2"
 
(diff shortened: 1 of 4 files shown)


pixi.toml
```diff
@@ -250,10 +252,6 @@ compiler-rt = ">=19.1.7,<20"
 pytest = ">=6"
 pytest-benchmark = ">=5.1.0,<6"
 
-[pypi-dependencies]
-pycollada = ">=0.9.2, <0.10"
-nodeenv = ">=1.10.0, <2"
```
sea-bass, line 255: also took the opportunity to install these from conda-forge instead of PyPi since they are available.

pixi.toml
```diff
@@ -223,10 +223,12 @@ xacro             = ">=2.1.1,<3"
 
 # Dependencies for examples
 matplotlib        = ">=3.10.5,<4"
+nodeenv           = ">=1.10.0, <2"
 tyro              = ">=0.9.26,<0.10"
-viser             = ">=1.0.21,<1.0.25"  # See https://github.com/viser-project/viser/issues/719
+viser             = ">=1.0.29,<2.0"  # See https://github.com/viser-project/viser/issues/719
 zstandard         = ">=0.25.0,<0.26"
 plyfile           = ">=1.1,<2"
+pycollada         = ">=0.9.2, <0.10"
```
eholum, line 231: huh, what is using pycollada directly?

eholum (approved):

pixi.toml
sea-bass: the Pinocchio viser visualizer. It (or trimesh, which is underneath) doesn't enforce that you have pycollada installed, you simply hit issues if you load a collada file.

PR #224, opened 2026-06-02
Improve TOPP-RA behavior for planar joints without discretizing path

sea-bass:
Improves https://github.com/open-planning/roboplan/pull/216 without having to discretize the path, which was causing some slowness and also lots of stop-and-go behavior in pathological cases.

Effectively, for non-standard joints (i.e., the ones with more than 1 position DOF), we simply do the TOPP-RA time parameterization in normalized coordinates, so that joints are interpolated correctly on their respective geodesics. This approach appears to work well, as shown by additional tests and overlaying paths + trajectories for the RRT example.

image


sea-bass:
ended up causing some pretty bad bugs, guess my idea was flawed. never mind :)

PR #227, opened 2026-06-02
Add Reachback mobile manipulator model

stephanie-eng:
Add the Reachback, which is more-or-less a Doosan H2017 bolted on top of a Clearpath Ridgeback with some other bits added in.

I'm new to this project, so I'm not 100% sure how you guys like to run and test things, so if there are millions of conventions I'm violating, let me know. I just tried to run the RRT demo with the Reachback model:

https://github.com/user-attachments/assets/3411df0c-6098-49ca-ba6b-7105f8cd064e


roboplan_examples/python/common.py
```diff
@@ -436,6 +436,74 @@ def get_model_data():
...
+                    disabled_collisions=["test_box"],
+                ),
+                ObstacleConfig(
+                    name="ground_plane",
+                    geom=coal.Box(5.0, 5.0, 0.2),
+                    parent_frame="universe",
+                    tform=pin.SE3(np.eye(3), np.array([0.0, 0.0, -0.1])).homogeneous,
+                    color=np.array([0.5, 0.5, 0.5, 0.5]),
+                    disabled_collisions=[
+                        "base_link",
```
sea-bass: This seems like it should be a permament disable in the SRDF?

roboplan_example_models/models/reachback_robot_model/reachback.urdf
```diff
@@ -0,0 +1,1007 @@
...
+  </material>
+  <material name="clearpath_aluminum">
+    <color rgba="0.5 0.5 0.5 1"/>
+  </material>
+  <material name="clearpath_plastic">
+    <color rgba="0.1 0.1 0.1   1"/>
+  </material>
+
+  <link name="base_footprint" />
+  <joint name="base_joint" type="planar">
```
sea-bass, line 36: Cool, glad to see the planar worked!

sea-bass (commented):
Yay, thank you! Good to see this. Your approach looks totally sensical, though I have a few comments:

1. Testing wise, I would try out all the Python examples (IK, RRT, and Oink specifically)
2. The Rolling job is known to fail right now because they just cut over from 24.04 to 26.04 but the image isn't updated yet. So ignore it
3. As we start collecting more robot model, file size is going to keep being more of a problem. I have a standing issue to resolve this, but not something for today. But if you have any ideas... 
4. ~The one that maybe we do want to look at for this PR is that the Kinova model also has a Robotiq. So maybe we can find a way to reuse these meshes instead of duplicating them?~ wait different robotiq model nvm

roboplan_example_models/models/reachback_robot_model/README.md
```diff
@@ -0,0 +1,4 @@
+These files were generated from the [`clearpath_community_demos`](https://github.com/clearpathrobotics/clearpath_community_demos) repository.
+
+Additionally, the Ridgeback meshes come from the [`ros2_robotiq_gripper`](https://github.com/clearpathrobotics/clearpath_common) repository.
+The Doosan meshes come from the [`ros2_robotiq_gripper`](https://github.com/DoosanRobotics/doosan-robot2) repository.
```
stephanie-eng: Bad copy paste, need to fix

roboplan_example_models/models/reachback_robot_model/reachback.urdf
stephanie-eng: Haha, yes! When I first showed you the mobile manipulator with roboplan, I noticed you didn't have planar joint support so I just stacked x, y, and theta joints. But since that recently changed, I figured I should restore the planar joint, and it worked fantastically :D

stephanie-eng:
> As we start collecting more robot model, file size is going to keep being more of a problem. I have a standing issue to resolve this, but not something for today. But if you have any ideas...

I assume we don't want to use LFS? Barring that, I'd imagine we'd want to have the models stored and retrieved from somewhere else.

roboplan_examples/python/common.py
stephanie-eng: I'll admit this is mostly copy-paste from https://github.com/open-planning/roboplan/blob/main/roboplan_examples/python/common.py#L435

For the reachback, base_link has no geometry though,  chassis_link is the link with geometry. I can just as well remove base_link from here.

sea-bass:
> I assume we don't want to use LFS? Barring that, I'd imagine we'd want to have the models stored and retrieved from somewhere else.

Right. LFS helps retrieve the data, but it doesn't dimish the file size of the entire package. It would have some be something like the latter, which is done for example in https://github.com/robot-descriptions/robot_descriptions.py

roboplan_example_models/models/reachback_robot_model/meshes/gripper-visual.stl
sea-bass, line 1: this is a big boi

sea-bass (commented):
I tested this branch during my lunch break on all the examples:

* In the RRT example, if you add `--include-obstacles`, then there is no valid initial configuration. I think the wheels are also colliding with the `ground_plane` object and may need to be excluded. Alternatively you could just lower the ground plane collision object, which also lets you remove the `chassis_link` from the exclusion list.
* IK examples all worked great ✔️

roboplan_example_models/models/reachback_robot_model/meshes/gripper-visual.stl
stephanie-eng: Indeed, I deleted it and just used the collision since it looks the same to me anyway :shrug: 

Thanks for the pointers, got it working with `--include-obstacles` too. My bad for not checking that was even an option. I was kind of wondering where those test obstacles were used :sweat_smile: 

[reachback.webm](https://github.com/user-attachments/assets/7afbd26f-e39b-46c8-a3be-a95c30aed339)

sea-bass (approved):

sea-bass:
By the way, I just went down a perf rabbit hole that should help with this example tremendously: https://github.com/open-planning/roboplan/pull/232

PR #228, opened 2026-06-03
Scene parsing and joint configuration clamping improvements

sea-bass:
This PR does a handful of things:

1. Throws an exception if there are duplicate frame names of different types (e.g., joint + body)
2. Fixes a bug parsing SRDF chain groups if the base link isn't directly the parent of a joint, but is rather connected via fixed joint
3. Renames `Scene::isValidPose()` to `Scene::isValidConfiguration()` to be technically correct
4. Adds a new `Scene::clampToValidConfiguration()` method to handle numerical drift from hardware or simulation drivers
5. Makes RRT planning pre-validation error messages more granular to ease debugging

### DIFF
--- a/roboplan/include/roboplan/core/scene.hpp
+++ b/roboplan/include/roboplan/core/scene.hpp
@@ -122,7 +122,14 @@ public:
   /// @brief Checks if the specified joint positions are valid with respect to joint limits.
   /// @param q The joint positions.
   /// @return True if the positions respect joint limits, else false.
-  bool isValidPose(const Eigen::VectorXd& q) const;
+  bool isValidConfiguration(const Eigen::VectorXd& q) const;
+
+  /// @brief Clamps the specified joint positions to valid joint limits.
+  /// @details Bounded joints are clamped to their position limits, while continuous and planar
+  /// rotation representations are renormalized onto the unit circle.
+  /// @param q The joint positions.
+  /// @return A new vector of joint positions that respects joint limits.
+  Eigen::VectorXd clampToValidConfiguration(const Eigen::VectorXd& q) const;
 
   /// @brief Converts partial joint positions to full joint positions.
   /// @details This includes adding new joints.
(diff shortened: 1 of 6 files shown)


roboplan/src/core/scene.cpp
eholum: Oh right, this happens in `clampToValidConfiguration` but probably needs to happen here?

```suggestion
        if (q(q_idx + idx) < lo || q(q_idx + idx) > hi) {
```

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -122,7 +122,14 @@ class Scene {
   /// @brief Checks if the specified joint positions are valid with respect to joint limits.
   /// @param q The joint positions.
   /// @return True if the positions respect joint limits, else false.
-  bool isValidPose(const Eigen::VectorXd& q) const;
+  bool isValidConfiguration(const Eigen::VectorXd& q) const;
+
+  /// @brief Clamps the specified joint positions to valid joint limits.
```
eholum, line 127: What other use cases are there for this other than for annoying deltas coming out of (possibly poorly written) hardware interfaces? It's a bandaid, but I do feel like it's a bandaid that we regularly run into so seems nice to have here.

eholum (commented):
Minor notes but lgtm! Except I'm pretty sure that's a bug for planar joints...

roboplan/src/core/scene.cpp
sea-bass: oh woooow yeah, good catch -- also happens in the default case

roboplan/include/roboplan/core/scene.hpp
sea-bass: There probably are other use cases, but what I converged on here was not go down the old MoveIt route where it did the clamping for you, but rather letting people directly call this function if they do need to address fudge factors, use this in custom algorithms, etc.

eholum (approved):
:fire:

PR #231, opened 2026-06-03
Minor improvements to SimpleIK and forward kinematics

eholum:
* Optionally set a base_link in compute forward kinematics
* Can set and use the base_link in the IK solver, rather than relying on a "root" link
* Add option to keep iterating until a "best" solution is found in simpleIK


sea-bass (commented):
Looks good, though the behavior of "return first solution" vs. "use all your time budget to give me the nearest to start seed" should be configurable in the `SimpleIkOptions`. Can you add that?

sea-bass (approved):
Approving, though now that we have the no fast return option, maybe we should reduce the default `max_time` from `0.01`, which is quite large?

roboplan_simple_ik/include/roboplan_simple_ik/simple_ik.hpp
```diff
@@ -17,7 +17,7 @@ struct SimpleIkOptions {
   size_t max_iters = 100;
 
   /// @brief Max total computation time, in seconds.
-  double max_time = 0.01;
+  double max_time = 0.005;
```
sea-bass, line 20: i think the bindings also has a default value which sadly doesn't sync :(

roboplan_simple_ik/bindings/src/simple_ik.cpp
sea-bass: here it is

eholum: oh good shout, also need the default for fast_return...

PR #232, opened 2026-06-04
Speed up collision checking and RRT

sea-bass:
After a humbling benchmark against MoveIt / OMPL, I went down a crusade of improving performance here. This manifests itself in a few changes.

1. (Biggest change) Switched to using Pinocchio's broadphase manager which dramatically speeds up collision checking. To keep around ability to do concurrent collision checking on a shared scene, I introduced a collision context which is a little complex, but IMO worthwhile. This one helps with RRT, path shortcutting, and adaptive TOPP-RA.
2. Improved random state generation for RRT to only sample joints in the relevant group. This was a big difference for scenes that have a lot of extraneous joints to the planning group.
3. Updated the "bisection" approach to check collisions along a path, which previously required the number of points to be a power of 2. This has a not-great worst-case scenario, which makes even using a queue in the new implementation better.
4. Since RRT already validates endpoints when sampling nodes, added an option to check collisions along path that excludes endpoint checking. This one is the most modest change, but generally useful.
5. Fixes RRTConnect behavior by removing static goal biasing, and instead actually trying to grow trees towards each other.

See this (real time) video as proof.

[rrt-dual-speedup.webm](https://github.com/user-attachments/assets/3c6d9e8f-b555-4cb4-beb1-693b612a75a1)


roboplan/bindings/src/core.cpp
```diff
@@ -305,9 +306,18 @@ void init_core_path_utils(nanobind::module_& m) {
...
+  m.def(
+      "hasCollisionsAlongPath",
+      [](const Scene& scene, const Eigen::VectorXd& q_start, const Eigen::VectorXd& q_end,
+         double max_step_size, bool bisection, bool check_endpoints) {
+        // Build a one-off collision context for this query. This convenience binding is not on a
+        // hot path; algorithms that check many paths should own a long-lived CollisionContext.
+        const CollisionContext context(scene);
+        return hasCollisionsAlongPath(scene, context, q_start, q_end, max_step_size, bisection,
+                                      check_endpoints);
+      },
```
sea-bass: an alternative here is for the Scene model to store its own `CollisionContext` object internally that anyone using can poke at, but with a caveat that it won't be thread-safe... or it can at least be locked.

roboplan/test/test_scene.cpp
```diff
@@ -112,15 +113,16 @@ TEST_F(RoboPlanSceneTest, CollisionCheck) {
...
   q_start_free << 0.0, -1.57, 0.0, 0.0, 0.0, 0.0;
   Eigen::VectorXd q_end_free(6);
   q_end_free << 1.0, -1.57, 1.57, 0.0, 0.0, 0.0;
-  EXPECT_FALSE(hasCollisionsAlongPath(*scene_, q_start_free, q_end_free, 0.05));
+  EXPECT_FALSE(hasCollisionsAlongPath(*scene_, collision_context, q_start_free, q_end_free, 0.05));
 
   Eigen::VectorXd q_end_coll(6);
   q_end_coll << 0.0, -1.57, 3.0, 0.0, 0.0, 0.0;
-  EXPECT_TRUE(hasCollisionsAlongPath(*scene_, q_start_free, q_end_coll, 0.05));
+  EXPECT_TRUE(hasCollisionsAlongPath(*scene_, collision_context, q_start_free, q_end_coll, 0.05));
```
sea-bass: same deal here; if a Scene had its own internal collision context then this test could be reverted to its state on `main`

roboplan/include/roboplan/core/collision_context.hpp
```diff
@@ -0,0 +1,60 @@
...
+namespace roboplan {
+
+class Scene;
+
+/// @brief Per-consumer scratch space for collision queries (Data + GeometryData + broadphase tree).
+/// @details A Scene's collision-query path mutates shared scratch state (joint/frame placements,
+/// the geometry world transforms, and the broadphase AABB tree), so a single Scene cannot answer
+/// collision queries from multiple threads concurrently. A CollisionContext owns its own copy of
+/// that scratch over the Scene's *immutable* Model and GeometryModel (shared by pointer), so an
+/// algorithm such as the RRT can run its collision queries without contending with anything else.
```
eholum, line 28: Oh this is gonna dope when we need to run 24x simultaneous RRTs...

eholum (approved):
Straight fire.

PR #233, opened 2026-06-04
Fix the example IK python script

eholum:
To use the actual base frame for control markers, now that simple ik supports it.

https://github.com/user-attachments/assets/43396727-b202-497f-a523-636b30104df6


sea-bass (changes requested):
I don't think computing the base_fk once statically works when your base link is movable. See the `stretch` or `reachback` models:

image

Image

We either have to:
1. Just do everything in world frame (base link is `""`)
2. Do the FK every time you move the marker, and convert the marker's world position to the current base link frame

eholum:
Ah man what even is a mobile robot?

Option 1 is definitely the fastest, but I'd also like to have an example of making markers work for these use cases...

sea-bass (approved):
Works for me now.

One issue we face is that there is no validation is the `base_link` is actually above the base joint of the joint group. I am not entirely sure how worthwhile this is vs. having a separate validation utility for this that isn't in the IK solver hot loop.

eholum:
Ship it?

sea-bass:
> Ship it?

Ship it.

PR #235, opened 2026-06-05
support accel + jerk limits in urdf>=1.2 

### LINKED ISSUE #217: Support acceleration + jerk limits in URDF >= 1.2
URDF >= 1.2 adds acceleration and jerk limits.

While this is not yet supported in Pinocchio, we could try read them off the URDF files in addition to our special YAML file with limit overrides.

muslim-adedamola:
PR intended to fix #217 

Our current urdfdom only supports up to 1.1 and rejects version 1.2 so acceleration and jerk are read from limit tags regardless of version. Present means used, absent means unlimited and YAML still wins when both are set. TinyXML2 is used directly, same as the SRDF parsing

### DIFF
--- a/roboplan/src/core/scene.cpp
+++ b/roboplan/src/core/scene.cpp
@@ -1,6 +1,8 @@
 #include <algorithm>
 #include <cmath>
 #include <stdexcept>
+#include <string>
+#include <unordered_map>
 
 #include <tl/expected.hpp>
 
@@ -59,6 +61,7 @@ Scene::Scene(const std::string& name, const std::string& urdf, const std::string
 
   // Single model with Pinocchio native mimics
   pinocchio::urdf::buildModelFromXML(urdf, model_, /*verbose*/ false, /*mimic*/ true);
+  const auto urdf_extended_limits = parseUrdfExtendedJointLimits(urdf);
 
   YAML::Node yaml_config;
   if (!yaml_config_path.empty() && !std::filesystem::is_directory(yaml_config_path)) {
@@ -149,6 +152,7 @@ Scene::Scene(const std::string& name, const std::string& urdf, const std::string
         }
       }
     }
+    const auto urdf_extended_it = urdf_extended_limits.find(joint_name);
     for (int idx = 0; idx < joint.nv(); ++idx) {
       if (maybe_vel_limits) {
         info.limits.max_velocity[idx] = maybe_vel_limits.value()[idx].as<double>();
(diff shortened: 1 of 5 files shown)


roboplan/src/core/scene.cpp
```diff
@@ -24,6 +28,54 @@ constexpr double kUnitCircleTol = 1.0e-6;
...
 constexpr double kDefaultPlanarJointTranslationLimit = 2.0;
 
+/// @brief Holds acceleration and jerk limits parsed from a URDF 1.2 <limit> tag.
+struct Urdf12JointLimits {
+  std::optional<double> acceleration;
+  std::optional<double> jerk;
+};
+
+/// @brief Parses URDF 1.2 extended joint limits (acceleration, jerk).
+/// @details Returns an empty map if the URDF version is not 1.2, or if parsing fails.
```
sea-bass: Is it checking the version here? Although I frankly don't think checking versions is that important and we should just remove the "if the URDF version is not 1.2".

sea-bass (commented):
Looks great!

Just one comment: Instead of calling classes/variables `urdf12_limits` etc., can we use a more lasting name like `urdf_extended_limits` or something?

Also can you leave a note in the code that this is a temporary holdover until Pinocchio properly supports URDF 1.2 and its extra limits in its own parsers? You can leave a link to https://github.com/stack-of-tasks/pinocchio/issues/2893

roboplan/src/core/scene.cpp
```diff
@@ -24,6 +28,60 @@ constexpr double kUnitCircleTol = 1.0e-6;
 /// @details TODO Make this more configurable by users.
 constexpr double kDefaultPlanarJointTranslationLimit = 2.0;
 
+/// @brief Holds extended joint limits (acceleration, jerk) parsed from a URDF <limit> tag.
```
sea-bass: Actually is this stuff movable to `scene_utils.hpp/cpp`?

roboplan/test/test_urdf12_limits.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include <memory>
+
+#include <roboplan/core/scene.hpp>
+
+namespace {
+constexpr double kTolerance = 1e-6;
+constexpr double kUnlimited = std::numeric_limits<double>::max();
+}  // namespace
+
+const std::string SRDF = R"(
```
sea-bass: For all these variable names we use Google standard, e.g. `kSrdf`

roboplan/test/test_urdf_extended_limits.cpp
sea-bass, line 1: Make sure you rename this file / test target name too!

muslim-adedamola: i considered renaming it but i thought to leave it since this is targeted towards version 1.2 of urdf
will fix

sea-bass (approved):
Thank you!

PR #236, opened 2026-06-05
Speed up path shortcutting

sea-bass:
This PR takes a similar scalpel to path shortcutting as #232 did with RRT, and does the following:

1. Adds a method to remove colinear points which largely contributes to "micro-segments" that trip up trajectory timing.
2. Adds a parameter for early stopping after a number of no-progress iterations. Somehow we didn't have this, which is wild.
3. In the wake of more options, formally extracts out a `PathShortcuttingOptions` struct.
4. Caught and fixed a bug with the random seed logic being applied backwards.

### DIFF
--- a/roboplan/src/core/path_utils.cpp
+++ b/roboplan/src/core/path_utils.cpp
@@ -119,11 +119,12 @@ bool hasCollisionsAlongPath(const Scene& scene, const Eigen::VectorXd& q_start,
       max_step_size, bisection, check_endpoints);
 }
 
-PathShortcutter::PathShortcutter(const std::shared_ptr<Scene> scene, const std::string& group_name)
-    : scene_{scene} {
+PathShortcutter::PathShortcutter(const std::shared_ptr<Scene> scene,
+                                 const PathShortcuttingOptions& options)
+    : scene_{scene}, options_{options} {
 
   // Validate the joint group.
-  const auto maybe_joint_group_info = scene_->getJointGroupInfo(group_name);
+  const auto maybe_joint_group_info = scene_->getJointGroupInfo(options_.group_name);
   if (!maybe_joint_group_info) {
     throw std::runtime_error("Could not initialize path shortcutter: " +
                              maybe_joint_group_info.error());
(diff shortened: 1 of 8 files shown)


roboplan/src/core/path_utils.cpp
```diff
@@ -133,16 +134,18 @@ PathShortcutter::PathShortcutter(const std::shared_ptr<Scene> scene, const std::
...
+  const unsigned int max_convergence_iters = options_.max_convergence_iters;
 
   // Make a copy of the provided path's configurations.
   JointPath shortened_path = path;
   auto& path_configs = shortened_path.positions;
 
   // We sample in the range (0, 1] to prevent modification of the starting configuration.
   std::random_device rd;
-  std::mt19937 gen(seed < 0 ? seed : rd());
+  std::mt19937 gen(options_.seed < 0 ? rd() : static_cast<unsigned int>(options_.seed));
```
eholum, line 148: d'oh

roboplan/src/core/path_utils.cpp
```diff
@@ -153,18 +156,35 @@ JointPath PathShortcutter::shortcut(const JointPath& path, double max_step_size,
   auto q_start = q_full_;
   auto q_end = q_full_;
 
+  // Cadence (in iterations) at which to interleave the deterministic redundant-vertex removal pass
+  // that cleans up the micro-segments introduced by corner-cutting shortcuts.
+  constexpr unsigned int kRedundantRemovalCadence = 20;
```
eholum: Since we have the options should this be there? Or should it be derived from the iterations or something? I guess it's pretty innocuous but since it's here and we have the options...

eholum (approved):
The minor nit but def makes sense...

roboplan/src/core/path_utils.cpp
sea-bass: oh yeah that should def be an option

eholum (approved):

eholum:
approved presuming the build gets fixed...

Issue #229, opened 2026-06-03
Support specifying base frames in Scene methods such as FK/Jacobian


sea-bass:
We don't yet have this, but should.

PR #238, opened 2026-06-06
add base frame support to Jacobian

muslim-adedamola:
Intended to fixe #229.

`forwardKinematics` already supported `base_frame`. This brings `computeFrameJacobian` and `FrameTask` up to the same level.

**`Scene::computeFrameJacobian`** gains an optional `base_frame_id` (default `nullopt`, fully backward-compatible). When set, returns the relative Jacobian of the EE w.r.t. the base frame, computed directly from LWA Jacobians:

```
J_rel_lin = J_ee_lin - J_base_lin + skew(dp) * J_base_ang
J_rel_ang = J_ee_ang - J_base_ang
```
**`FrameTask`** resolves `CartesianConfiguration::base_frame` to a frame ID at construction. `computeError` converts the relative target to world frame via the base frame's current FK pose; `computeJacobian` passes `base_frame_id` through to the new API.


roboplan/src/core/scene.cpp
```diff
@@ -543,8 +507,72 @@ Eigen::Matrix4d Scene::forwardKinematics(const Eigen::VectorXd& q, const std::st
...
+  // This avoids toActionMatrix() convention issues between Pinocchio versions.
+  Eigen::MatrixXd J_ee_lwa = Eigen::MatrixXd::Zero(6, model_.nv);
+  Eigen::MatrixXd J_base_lwa = Eigen::MatrixXd::Zero(6, model_.nv);
+  pinocchio::computeFrameJacobian(model_, model_data_, q, frame_id, pinocchio::LOCAL_WORLD_ALIGNED,
+                                  J_ee_lwa);
+  pinocchio::computeFrameJacobian(model_, model_data_, q, base_id, pinocchio::LOCAL_WORLD_ALIGNED,
+                                  J_base_lwa);
+
+  // Copy oMf immediately after each call that sets it (avoids stale references).
+  const pinocchio::SE3 T_ee = model_data_.oMf.at(frame_id);
```
sea-bass: Does `computeFrameJacobian` already do the FK and updating frame placements internally? Because if not, this data structure may not necessarily be up to date.

sea-bass (commented):
Thank you! I am traveling this week so it will take me a while to go through the math in detail, but in the meantime, are you able to share any references to help with this, as well as some video results of using the Oink solver with different base frames to check that this works? Of particular interest would be controlling with respect to movable frames, and checking that the interactive markers are tracking properly.

I also believe Pinocchio offers a `toActionMatrix()` conversion that may help abstract some of the manual calculations, but not sure... worth checking: https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/structpinocchio_1_1SE3Base.html#aec388721bf6853ed1fbebcf242cca30e

On a completely different note, the boolean parameter on local vs. non-local makes the Jacobian code pretty complicated looking. I wonder if it might be easier to follow if we simply split this function into local and non-local versions?

muslim-adedamola:
> Thank you! I am traveling this week so it will take me a while to go through the math in detail, but in the meantime, are you able to share any references to help with this, as well as some video results of using the Oink solver with different base frames to check that this works? Of particular interest would be controlling with respect to movable frames, and checking that the interactive markers are tracking properly.
> 
> I also believe Pinocchio offers a `toActionMatrix()` conversion that may help abstract some of the manual calculations, but not sure... worth checking: https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/structpinocchio_1_1SE3Base.html#aec388721bf6853ed1fbebcf242cca30e
> 
> On a completely different note, the boolean parameter on local vs. non-local makes the Jacobian code pretty complicated looking. I wonder if it might be easier to follow if we simply split this function into local and non-local versions?

Hi sea-bass,

Thanks for the review and I wish you safe travels!

**On the math.** I followed the velocity composition rule for two frames in relative motion, specialised to world-aligned coordinates ([Siciliano](https://people.disim.univaq.it/~costanzo.manes/EDU_stuff/Robotics_Modelling,%20Planning%20and%20Control_Sciavicco_extract.pdf) et al., *Robotics: Modelling, Planning and Control*, section 3.1.1, Eq. 3.14; same result in [Featherstone](https://gaoyichao.com/Xiaotu/papers/2008%20-%20Rigid%20body%20dynamics%20algorithms.pdf), *Rigid Body Dynamics Algorithms*, sec 2.2 Eq. 2.1 + sec 2.8 Eq. 2.21):

```
v_ee_world = v_base_world + ω_base × (p_ee − p_base) + v_ee_rel_world
⇒ v_ee_rel_world = v_ee_world − v_base_world − ω_base × (p_ee − p_base)
   ω_rel = ω_ee − ω_base
```

Pinocchio's LWA Jacobians give `v_ee_world` and `v_base_world` per joint, so the implementation is the matrix form of those two lines.

**On `toActionMatrix()`.** I tried this first. The subtlety is which frame the resulting velocity ends up in. The standard form

```
J_rel_local = J_ee_local − Ad(T_rel⁻¹) · J_base_local
```

gives velocity in `T_rel`'s body frame (rotation `R_rel = R_base^T · R_ee`). Converting to LWA via `R_rel · J_rel_local` rotates into a frame aligned with `R_rel`, not world. Since `FrameTask` computes its error in the world frame the Jacobian needs to match. The direct LWA formulation gives that without an intermediate frame swap. I'll add an inline comment explaining this.

**On splitting the function.** Agreed. I can push a refactor with two public methods:

- `Scene::computeFrameJacobian(q, frame_id, ref_frame, jacobian)`: restored to the original signature
- `Scene::computeRelativeFrameJacobian(q, frame_id, base_frame_id, ref_frame, jacobian)`: new

`FrameTask::computeJacobian` will dispatch based on whether `base_frame` is set.

**On FK currency.** `pinocchio::computeFrameJacobian` calls `forwardKinematics` and `updateFramePlacement(frame_id)` internally, so `data.oMf[frame_id]` is current immediately after each call. The two `oMf` copies are taken right after their corresponding `computeFrameJacobian` calls. I'll add a comment to make that explicit.

A short Viser demo with `base_frame` set to a movable mid-arm link is coming in the next day or so.

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -179,13 +161,25 @@ class Scene {
...
-  void computeFrameJacobian(const Eigen::VectorXd& q, pinocchio::FrameIndex frame_id,
-                            pinocchio::ReferenceFrame reference_frame,
-                            Eigen::Ref<Eigen::MatrixXd> jacobian) const;
+  /// @param base_frame_id Optional Pinocchio frame ID of the base frame. If set, returns the
+  /// Jacobian of the EE velocity relative to the base frame. Defaults to std::nullopt.
+  void
+  computeFrameJacobian(const Eigen::VectorXd& q, pinocchio::FrameIndex frame_id,
+                       pinocchio::ReferenceFrame reference_frame,
+                       Eigen::Ref<Eigen::MatrixXd> jacobian,
+                       std::optional<pinocchio::FrameIndex> base_frame_id = std::nullopt) const;
```
sea-bass: to be consistent with other usage, this should be a string, not an ID. And the ID can be looked up in the method itself.

sea-bass (commented):
Just tested and it all seems to work fine!

Note that we made some updates in https://github.com/open-planning/roboplan/pull/233 particularly with the base frame names in `common.py`, because using the base links as the base frame for the `stretch` and `reachback` models meant that the Jacobian rows corresponding to the mobile bases were zero, which is not good. So these should be `"universe"` now.

However, make sure you make updates to `example_oink.py` (and its relatives for action chunks and barriers etc.) to properly account for the base frame in the same was `example_ik.py` already did.

For the references that you responded with in previous comments, if you can put them in as comments in the implementation, that would be helpful.

muslim-adedamola:
> Just tested and it all seems to work fine!
> 
> Note that we made some updates in #233 particularly with the base frame names in `common.py`, because using the base links as the base frame for the `stretch` and `reachback` models meant that the Jacobian rows corresponding to the mobile bases were zero, which is not good. So these should be `"universe"` now.
> 
> However, make sure you make updates to `example_oink.py` (and its relatives for action chunks and barriers etc.) to properly account for the base frame in the same was `example_ik.py` already did.
> 
> For the references that you responded with in previous comments, if you can put them in as comments in the implementation, that would be helpful.


Thanks for testing it, and apologies for the silence. I got pulled onto some other lab work that ate more time than expected. I'll update the examples (oink + action chunks + barriers) to handle the base frame the way example_ik.py does, and move the references into comments. Should have it up soon

muslim-adedamola:
All updated and pushed.

1. Split: `computeFrameJacobian` is back to its original signature, with a new `computeRelativeFrameJacobian`. `FrameTask::computeJacobian` dispatches on whether a base frame is set.

2. Base frame by name: `computeRelativeFrameJacobian` now takes the base frame as a string and looks up the ID internally, matching `fk`.

3. Examples: `example_oink.py`, `example_oink_position_barriers.py`, and `example_action_chunk_tracking.py` now express targets in the task's base frame before passing them to the solver, same as `example_ik.py`. No-op for the `"universe"` models, correct for a real base frame.

4. References: Moved into a comment above the relative-Jacobian computation in `scene.cpp`.

5. Demo: Added `example_oink_base_frame_demo.py --model dual`: left arm tracks a world marker, right arm holds a fixed pose relative to the left hand, so dragging the left hand moves the right arm's base frame and it tracks along. Optional self-collision barrier via `--self-collision-num-pairs`. Clips below (Dual FR3 showing the feature, plus stretch/reachback/ur5 to show nothing regressed)

-----------------------------------------------------------------------------------------------------------------------------------------------------------------------
dual_no_collision_demo


------------------------------------------------------------------------------------------------------------------------------------------------------------------------
stretch_demo


------------------------------------------------------------------------------------------------------------------------------------------------------------------------
reachback_demo


------------------------------------------------------------------------------------------------------------------------------------------------------------------------
ur5_demo

roboplan/bindings/src/core.cpp
```diff
@@ -234,12 +234,32 @@ void init_core_scene(nanobind::module_& m) {
             }
             const auto reference_frame =
                 local ? pinocchio::ReferenceFrame::LOCAL : pinocchio::ReferenceFrame::WORLD;
+
             Eigen::MatrixXd jacobian = Eigen::MatrixXd::Zero(6, self.getModel().nv);
             self.computeFrameJacobian(q, maybe_frame_id.value(), reference_frame, jacobian);
             return jacobian;
           },
-          "Computes the frame Jacobian for a specific frame.", "q"_a, "frame_name"_a,
+          "Computes the world-rooted frame Jacobian for a specific frame.", "q"_a, "frame_name"_a,
```
sea-bass: nit: Change this to "Compute the frame Jacobian for a specific frame, expressed in world frame".

Same in the generated `.pyi` file.

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -178,15 +178,34 @@ class Scene {
   Eigen::Matrix4d forwardKinematics(const Eigen::VectorXd& q, const std::string& frame_name,
                                     const std::string& base_frame = "") const;
 
-  /// @brief Computes the frame Jacobian for a specific frame.
+  /// @brief Computes the world-rooted frame Jacobian for a specific frame.
```
sea-bass: same comment as above here re: "world-rooted".

roboplan/src/core/scene.cpp
```diff
@@ -547,6 +547,81 @@ void Scene::computeFrameJacobian(const Eigen::VectorXd& q, pinocchio::FrameIndex
...
+  const pinocchio::SE3 T_base = model_data_.oMf.at(base_id);
```
sea-bass: I get that this is trying to be thread-safe, but it's anyways not fully safe because `model_data_` could still be modified between executing these lines.

We can continue to use references here for now, and thread safety can be handled in a more principled way down the line.

roboplan_examples/python/example_oink_base_frame_demo.py
sea-bass, line 1: thanks for this example as a proof of concept, but I think it might have too much redundancy with the other ones. The animations were great to see as confirmation though!

It's probably best to just let people modify the base frame in `example_oink.py` themselves. So we can just delete this.

sea-bass (commented):
Looks good -- only have a few remaining comments and then I think this is ready. Thanks again @muslim-adedamola!

roboplan/bindings/src/core.cpp
muslim-adedamola: Done.

roboplan/include/roboplan/core/scene.hpp
muslim-adedamola: Done

sea-bass (approved):

PR #240, opened 2026-06-13
Cartesian path planner

### LINKED ISSUE #55: Create Cartesian path planning example
Consider doing this using differential IK, not position based IK, to be able to better trace approximate paths.

This could either be using oink or by exposing a single-step variation on simple_ik

sea-bass:
This PR adds a Cartesian Path planner with 2 modes:

1. A `Bounded` mode, which tries to adhere to max Cartesian speed/accelerations unless joint limits are violated; in which case it tries to slow down until/unless it can't.
2. A `TimeOptimal` mode which is time-optimal. This borrows from how MoveIt does things with TOTG, using a new TOPP-RA parameterization that does linear segments with circular blends. This gives really nice behavior, and it is bang-bang on the acceleration.

Some modifications that came with this, which are minor yet breaking changes:

1. TOPP-RA parameterizer had way too many arguments, so also made an options struct here.
2. Modified `plotJointTrajectory` to optionally show velocities/accelerations, and used collapsed continuous positions
3. Fixed the way gtests are added in the non-ROS case, since `gtest_discover_tests` was actually running the tests at build time! We wanted `gtest_add_tests` instead woops

[cartesian_hello_world.webm](https://github.com/user-attachments/assets/b94c87c7-5ff1-40e0-8923-9048a822f660)

[cartesian-dual.webm](https://github.com/user-attachments/assets/c82616f5-89a7-4ce9-88de-3eb949e35a92)

[cartesian-rounded-ur5.webm](https://github.com/user-attachments/assets/ce7eb890-2e49-4964-918a-aa64fb07741a)

Closes #55


docs/source/concepts/cartesian_planning.rst
```diff
@@ -0,0 +1,53 @@
...
+
+The ``roboplan_cartesian_planning`` package traces a Cartesian path in joint space using the :ref:`OInK <oink-solver>` optimal IK solver.
+
+Approach
+--------
+
+The planner builds an arc-length SE(3) reference from the waypoints (linear interpolation for position, SLERP for orientation).
+It then runs an offline servo loop using the OInK solver.
+At each control step, the planner advances the reference, solves one differential-IK step, and integrates the result.
+When the robot cannot keep up (e.g., near a singularity or joint velocity limit), the reference feedrate is throttled so that every committed sample stays within tolerance of the geometric path.
```
sjahr, line 17: Is there a point at which the planner fails? E.g. if the Cart. error gets to large?

roboplan_cartesian_planning/bindings/python/roboplan/cartesian_planning/_cartesian_ext.pyi
```diff
@@ -0,0 +1,233 @@
...
+
+
+class CartesianSpeedMode(enum.Enum):
+    """Selects how the planner assigns speed/timing along the path."""
+
+    Constant = 0
+    """Trace the path at a (roughly) constant Cartesian tool speed."""
+
+    Toppra = 1
+    """Time-optimal re-timing respecting joint limits (not yet implemented)."""
```
sjahr: Can you open an issue for this if it is not yet implemented and link the issue in the code?

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
```diff
@@ -0,0 +1,312 @@
...
+  /// @brief The achieved Cartesian path length (meters) of the resulting motion.
+  double achieved_path_length = 0.0;
+
+  /// @brief The fraction of control steps that ran at the full commanded feedrate
+  /// (i.e. were not throttled). 1.0 means the commanded speed was held throughout.
+  double feedrate_efficiency = 1.0;
+
+  /// @brief Peak |joint velocity| / velocity-limit ratio over the trajectory.
+  /// Values <= 1.0 mean the joint velocity limits are respected.
+  double peak_velocity_ratio = 0.0;
```
sjahr: Do we need these as return value or should we make them member functions of the joint trajectory so you can calculate them on demand independent of the planner you're using

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+  /// @brief Peak |joint acceleration| / acceleration-limit ratio over the trajectory.
+  /// The Constant mode is velocity-level and does not bound acceleration,
+  /// so this can exceed 1.0; use the Toppra mode to keep it within limits.
+  double peak_acceleration_ratio = 0.0;
```
sjahr: Same here

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
```diff
@@ -0,0 +1,312 @@
...
+/// The planner drives the motion by repeatedly updating each tracking FrameTask's target pose,
+/// so one tracking task must be provided per end-effector in the CartesianPath. All other
+/// tasks/constraints/barriers are passed to the solver unchanged on every step. The same
+/// objects are reused across all plan() calls; the planner never rebuilds or mutates them
+/// (other than the tracking tasks' targets), so any q_start-dependent setup (e.g. seeding a
+/// ConfigurationTask) is the caller's responsibility.
+struct CartesianPlannerComponents {
+  /// @brief The OInK solver to use.
+  /// @details Must be constructed for the same scene and joint group as the planner.
+  /// Must not be null.
```
sjahr, line 154: Do we check this later?

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
```diff
@@ -0,0 +1,312 @@
...
+  /// @brief Constraints applied at every control step (e.g. VelocityLimit, PositionLimit).
+  /// May be empty.
+  std::vector<std::shared_ptr<Constraints>> constraints;
+
+  /// @brief Control barrier functions applied at every control step. May be empty.
+  std::vector<std::shared_ptr<Barrier>> barriers;
+};
+
+/// @brief Offline Cartesian path planner that traces a CartesianPath in joint space.
+/// @details Uses the Oink optimal IK solver as a differential-IK tracker. See the
```
sjahr: Is the package README the right place to find more info?

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
```diff
@@ -0,0 +1,312 @@
...
+                       const CartesianPlannerComponents& components);
+
+  /// @brief Plans a joint trajectory that traces the provided Cartesian path.
+  /// @details Supports one or more end-effector frames (each entry in the path's
+  /// base_frames/tip_frames/tforms is traced simultaneously by its own FrameTask).
+  /// @param path The Cartesian waypoint path to trace.
+  /// @param q_start The seed/start configuration, as a full model configuration
+  /// (size model.nq). The robot should already be at (or near) the first waypoint.
+  /// @return The plan result on success, else a string describing the error.
+  tl::expected<CartesianPlanResult, std::string> plan(const CartesianPath& path,
```
sjahr: How about making q_start optional and if not provided, the current scene state is used? Or would that be to imprecise?

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
```diff
@@ -0,0 +1,312 @@
...
+                                                      const JointConfiguration& q_start);
+
+private:
+  /// @brief A single arc-length time-parameterized SE(3) reference built from waypoints.
+  struct Reference {
```
sjahr: Maybe call it ReferencePath since the term Reference is very overloaded in computer science 😅

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+  }
+  // Pin the exact endpoints.
+  result.front() = positions.front();
+  result.back() = positions.back();
+  return result;
+}
+
+/// @brief Computes the position (meters) and orientation (radians) error between two
+/// SE(3) transforms expressed in the same frame.
+void poseError(const Eigen::Matrix4d& a, const Eigen::Matrix4d& b, double& position_error,
```
sjahr: Would it make sense to move this somewhere where it can be reused? Looks like this might be reusable 🤔

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+    tasks_.reserve(components_->tracking_tasks.size() + components_->extra_tasks.size());
+    tasks_.insert(tasks_.end(), components_->tracking_tasks.begin(),
+                  components_->tracking_tasks.end());
+    tasks_.insert(tasks_.end(), components_->extra_tasks.begin(), components_->extra_tasks.end());
+    constraints_ = components_->constraints;
+    barriers_ = components_->barriers;
+
+    // Verify against the scene's physical velocity limits (unscaled) if available; the caller
+    // owns whatever velocity constraint they supplied, so this is only a hard-limit sanity net.
+    const auto maybe_velocity_limits = scene_->getVelocityLimitVectors(options_.group_name);
```
sjahr: Maybe I am missing something here, but why do you throw later when there are no velocity limits provided and and here you don't. Maybe, we can entirely rely on the OInK solver to handle limits if it is provided by the user? 
This might be overly nitpicky, but I wonder if we can avoid calling getVelocityLimitVectors twice

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+                             maybe_joint_group_info.error());
+  }
+  joint_group_info_ = maybe_joint_group_info.value();
+  buildStaticSolverComponents();
+}
+
+void CartesianPathPlanner::buildStaticSolverComponents() {
+  const int num_variables = oink_->num_variables;
+
+  if (components_) {
```
sjahr: What if instead of keeping components_ as a member you just pass it in the constructor as optional argument to buildStaticSolverComponents(). That way you could get rid of the components_ member variable as far as I can see at this point 😅

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+  constraints_ = {std::make_shared<VelocityLimit>(*oink_, options_.dt, v_max),
+                  std::make_shared<PositionLimit>(*oink_, options_.position_limit_gain)};
+  verify_v_max_ = std::move(v_max);
+  has_velocity_check_ = true;
+}
+
+tl::expected<CartesianPlanResult, std::string>
+CartesianPathPlanner::plan(const CartesianPath& path, const JointConfiguration& q_start) {
+  // Validate options.
+  if (options_.dt <= 0.0) {
```
sjahr: What if this is a member function of struct called validate()? That would assign this check directly to options_

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+    return tl::make_unexpected("q_start must be a full model configuration of size model.nq (" +
+                               std::to_string(model.nq) + "), got " +
+                               std::to_string(q_start.positions.size()) + ".");
+  }
+
+  switch (options_.speed_mode) {
+  case CartesianSpeedMode::Constant:
+    return planConstantSpeed(path, q_start.positions);
+  case CartesianSpeedMode::Toppra:
+    return planToppra(path, q_start.positions);
```
sjahr: Oh I think the comment about one option not implemented earlier is not correct in case you haven't already updated it

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+    }
+    try {
+      world_T_base.at(f) = scene_->forwardKinematics(q_start_full, path.base_frames.at(f));
+    } catch (const std::exception& e) {
+      return tl::make_unexpected(std::string("Could not resolve base frame '") +
+                                 path.base_frames.at(f) + "': " + e.what());
+    }
+  }
+
+  // Common timeline: a single reference time `s` advances all frames together; each reference
```
sjahr: Nitpick: In my understanding `s` referes to the path variable so maybe use `t` for reference time?

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+    }
+    delta_q.setZero();
+    const auto result =
+        oink.solveIk(*scene_, tasks_, constraints_, barriers_, delta_q, options_.regularization);
+    if (!result) {
+      return tl::make_unexpected(result.error());
+    }
+    delta_q_full.setZero();
+    delta_q_full(oink.v_indices) = delta_q;
+    q_candidate = scene_->integrate(q, delta_q_full);
```
sjahr: Just sanity checking; Do we need to run some update frame placement here?

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+  for (size_t f = 0; f < num_frames; ++f) {
+    previous_positions.at(f) =
+        scene_->forwardKinematics(q, path.tip_frames.at(f)).block<3, 1>(0, 3);
+  }
+
+  // Timed servo loop with feedrate throttling for tolerance enforcement.
+  double s = 0.0;
+  int total_steps = 0;
+  int throttled_steps = 0;
+  const double eps = 1e-9;
```
sjahr: Move to top as constant

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+}
+
+tl::expected<CartesianPlanResult, std::string>
+CartesianPathPlanner::planToppra(const CartesianPath& path, const Eigen::VectorXd& q_start_full) {
+  // Resolve a dense geometric joint path that hugs the Cartesian path. The resolution
+  // speed only sets sampling density and tracking tightness here (TOPP-RA assigns the
+  // final timing), so use a deliberately low speed that essentially any robot can follow
+  // without lagging: a faster resolution speed makes the robot trail the reference by up
+  // to the tolerance band, and TOPP-RA would then faithfully reproduce that lag.
+  constexpr double kResolutionLinearSpeed = 0.05;   // m/s
```
sjahr, line 827: Move to top as constants

roboplan_cartesian_planning/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>roboplan_cartesian_planning</name>
+  <version>0.4.0</version>
+  <description>Cartesian path planner for RoboPlan.</description>
+  <maintainer email="sebas.a.castro@gmail.com">Sebastian Castro</maintainer>
```
sjahr, line 7: Super nit: How about you add yourself as author and maintainer so even if the maintainer changes sometime in the future, you're still credited as author?

roboplan_cartesian_planning/README.md
```diff
@@ -0,0 +1,3 @@
+# roboplan_cartesian_planning
+
+A Cartesian path planner for RoboPlan.
```
sjahr, line 3: Any references worth quoting here?

roboplan_oink/test/CMakeLists.txt
```diff
@@ -97,11 +97,11 @@ if ( ament_cmake_gtest_FOUND )
...
 else()
     include(GoogleTest)
-    gtest_discover_tests(test_oink)
-    gtest_discover_tests(test_frame_task)
-    gtest_discover_tests(test_configuration_task)
-    gtest_discover_tests(test_velocity_limit)
-    gtest_discover_tests(test_position_limit)
-    gtest_discover_tests(test_position_barrier)
-    gtest_discover_tests(test_self_collision_barrier)
+    gtest_add_tests(TARGET test_oink)
```
sjahr, line 113: Just out of curiosity, why is this necessary?

sjahr (commented):
Super cool! I tested it and it works. Bunch of nitpicks and questions but nothing major. Let's merge this soon 🚀

roboplan_oink/test/CMakeLists.txt
sea-bass: this was a bug in how we were defining tests -- see the PR description:

> Fixed the way gtests are added in the non-ROS case, since gtest_discover_tests was actually running the tests at build time! We wanted gtest_add_tests instead woops

I can split this out if you prefer? Was considering it but then got lazy.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+  Eigen::VectorXd q = q_start_full;
+
+  // Runs one differential-IK step that retargets every frame to its pose at reference time
+  // `s` from the committed configuration `q`, writing the candidate configuration and the
+  // worst-case (max over frames) world-frame FK pose error. Does not commit.
+  auto solve_step = [&](double s, Eigen::VectorXd& q_candidate, double& position_error,
+                        double& orientation_error) -> tl::expected<void, std::string> {
+    // Refresh the scene state to the committed configuration so the Oink tasks read the
+    // correct current pose, and retarget every tracking task.
+    scene_->setJointPositions(q);
```
eholum: This means that this isn't specifically thread safe, right? Unlike the fancy new collision context for RRT? It is making me wonder if it would be crazy dumb to provide a copy constructor or something for the whole darn class. Anyway, not a big deal because I think that's the default expectation for the `Scene` class.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -0,0 +1,698 @@
...
+  toppra_options.acceleration_scale = options_.acceleration_scale;
+  toppra_options.max_blend_deviation = options_.toppra_blend_deviation;
+  auto maybe_trajectory = toppra_.generate(joint_path, toppra_options);
+  if (!maybe_trajectory) {
+    return tl::make_unexpected("TOPP-RA time parameterization failed: " + maybe_trajectory.error());
+  }
+
+  CartesianPlanResult plan_result;
+  plan_result.trajectory = std::move(maybe_trajectory.value());
+  // Re-timing does not change the geometric path, so reuse the resolved length.
```
eholum: This is ignorance, but doesn't it change it slightly if there is non zero blending applied?

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
```diff
@@ -0,0 +1,312 @@
...
+  double max_position_error = 0.005;
+
+  /// @brief Maximum allowed orientation deviation from the path, in radians.
+  double max_orientation_error = 0.01;
+
+  /// @brief Oink FrameTask position cost weight.
+  double position_cost = 1.0;
+
+  /// @brief Oink FrameTask orientation cost weight.
+  double orientation_cost = 1.0;
```
eholum, line 80: Ok, if I understand these orientation costs correctly, presumably setting `max_orientation_error` to infinity and `orientation_cost` to 0 will enable free rotation along the cartesian path?

eholum (commented):
Just reading for my own edification since I haven't been as involved! No comments, just questions that I don't know the answer to...

The examples are straight dope.

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
sea-bass: exactly! And non-zero but lower means that position is higher-weighted than orientation.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: not with `scene_->integrate()`, this just adds a delta q, it's at the next time forward kinematics is called that the frame placements get updated I think?

roboplan_cartesian_planning/package.xml
sea-bass: `package.xml` doesn't have an `author` field. It's all maintainer.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: ooh yeah, all of oink is not thread safe because it calls `getCurrentJointPositions()` ... and because it uses a roboplan scene, lol. Therefore this planner as well.

The problem with a copy constructor is if you're doing something like connecting to ROS where you're receiving the latest scene state from some middleware, it gets funky.

docs/source/concepts/cartesian_planning.rst
sea-bass: exactly -- good sentence for me to add now!

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
sea-bass: now we do (thank you)

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
sea-bass: Eh, I think I'll leave it as-is for now. People can easily grab the scene's current positions and pass them in.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: Done, good idea!

roboplan_cartesian_planning/README.md
sea-bass: I think these are all in the `rst` docs.

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
sea-bass: nice, done! Removes the need for this extra result struct too.

roboplan_cartesian_planning/include/roboplan_cartesian_planning/cartesian_path_planner.hpp
sea-bass: This struct doesn't exist anymore, or at least is absorbed into this `FrameTrack` struct which I agree is less confusingly named.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: yeah that was weird, fixed

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: yep! good one

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: I think you're right, but there have fortunately been enough changes that this is now computed right at the end of producing the final trajectory!

sjahr (approved):
I like it, let's merge this 🚿 (once you addressed the CI failures :wink:)

PR #242, opened 2026-06-15
Allow overriding joint position limits in YAML config

sea-bass:
Been meaning to do this for a while... and in the process I also factored out the ballooning block of code.

For example, this is one way to pad the joint position limits from a URDF if -- hypothetically -- a simulator's contact model caused them to go slightly outside their range.


sjahr (approved):
Makes sense to me 👍

PR #246, opened 2026-06-15
Add RRT* and fast-return mode

### LINKED ISSUE #243: Consider adding RRT* implementation
Sometimes shortcutting is good, sometimes doing tree rewiring on the fly is good. Who knows?

I'm also genuinely curious how well rewiring will work given we have `dynotree`'s k-d tree implementation at our disposal.

sea-bass:
This PR adds both an RRT* implementation, and a "fast return" option which uses all the time budget vs. returns the first available RRT. More knobs to play with!

image

### DIFF
--- a/roboplan_examples/python/example_rrt.py
+++ b/roboplan_examples/python/example_rrt.py
@@ -35,8 +35,11 @@ def main(
     collision_check_use_bisection: bool = False,
     goal_biasing_probability: float = 0.15,
     max_nodes: int = 1000,
-    max_planning_time: float = 5.0,
+    max_planning_time: float = 2.0,
     rrt_connect: bool = False,
+    rrt_star: bool = False,
+    rewire_distance: float = 5.0,
+    fast_return: bool = True,
     include_shortcutting: bool = False,
     max_shortcutting_iters: int = 100,
     toppra_mode: SplineFittingMode = SplineFittingMode.Adaptive,
@@ -59,6 +62,9 @@ def main(
         max_nodes: The maximum number of nodes to add to the search tree.
         max_planning_time: The maximum time (in seconds) to search for a path.
         rrt_connect: Whether or not to use RRT-Connect.
+        rrt_star: Whether or not to use RRT*, which keeps optimizing until the node or time budget is exhausted and returns the lowest-cost path. Can be combined with `rrt_connect`.
+        rewire_distance: The configuration-distance radius used to find neighbors for RRT* rewiring (only used when `rrt_star` is true). Should generally be at least `max_connection_distance`.
+        fast_return: If true, return on the first path found; if false, plan until the node or time budget is exhausted and return the lowest-cost path. Set to false to get RRT*'s asymptotically optimal behavior.
         include_shortcutting: Whether or not to include path shortcutting for found paths.
         max_shortcutting_iters: The maximum number of path shortcutting iterations.
         toppra_mode: The trajectory generation mode for TOPP-RA. Can be `Hermite`, `Cubic`, or `Adaptive` (default).
@@ -128,6 +134,9 @@ def main(
         goal_biasing_probability=goal_biasing_probability,
         max_planning_time=max_planning_time,
         rrt_connect=rrt_connect,
+        rrt_star=rrt_star,
+        rewire_distance=rewire_distance,
+        fast_return=fast_return,
     )
     rrt = RRT(scene, options)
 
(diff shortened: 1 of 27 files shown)


roboplan_examples/python/example_rrt.py
```diff
@@ -35,8 +35,11 @@ def main(
     collision_check_use_bisection: bool = False,
     goal_biasing_probability: float = 0.15,
     max_nodes: int = 1000,
-    max_planning_time: float = 5.0,
+    max_planning_time: float = 2.0,
```
eholum, line 38: :rocket:

roboplan_rrt/bindings/test/test_rrt.py
```diff
@@ -41,3 +41,30 @@ def test_plan(test_scene: Scene) -> None:
...
+    options.group_name = "arm"
+    options.max_connection_distance = 1.0
+    options.collision_check_step_size = 0.05
+    options.rrt_star = True
+    options.rewire_distance = 2.0
+    # Disable fast_return so RRT* optimizes, and cap the budget so the test does not run too long.
+    options.fast_return = False
+    options.max_planning_time = 1.0
+
+    rrt = RRT(test_scene, options)
```
eholum: Could we plan twice and verify that the rrt* produced path is shorter than the plain rrt?

roboplan_rrt/src/rrt.cpp
```diff
@@ -224,9 +251,20 @@ bool RRT::growTree(KdTree& kd_tree, std::vector<Node>& nodes, const Eigen::Vecto
     }
 
     grew_tree = true;
-    auto new_id = nodes.size();
-    kd_tree.addPoint(collapse(q_extend(q_indices)), new_id);
-    nodes.emplace_back(q_extend, parent_id);
+    int new_id;
+    if (options_.rrt_star) {
+      new_id = rewire(kd_tree, nodes, q_extend, parent_id, collision_context);
```
eholum, line 256: Not suggesting a change just commenting because it's interesting... but if `greedy` is set, we rewire every time an intermediate node gets added? I bet it ends up doing some redundant work depending on the neighbors distance / max_connection_distance, but I also bet it's negligible.

roboplan_rrt/bindings/src/rrt.cpp
```diff
@@ -41,7 +44,14 @@ void init_rrt(nanobind::module_& m) {
       .def_rw("max_planning_time", &RRTOptions::max_planning_time,
               "The maximum amount of time to allow for planning, in seconds.")
       .def_rw("rrt_connect", &RRTOptions::rrt_connect,
-              "If true, use the RRT-Connect algorithm to grow the search trees.");
+              "If true, use the RRT-Connect algorithm to grow the search trees.")
+      .def_rw("rrt_star", &RRTOptions::rrt_star,
+              "If true, use the RRT* algorithm to grow asymptotically optimal trees.")
+      .def_rw("rewire_distance", &RRTOptions::rewire_distance,
+              "The configuration-distance radius used to find neighbors for RRT* rewiring.")
```
eholum: Just inconsistent with the use above, 

```suggestion
              "The configuration distance radius used to find neighbors for RRT* rewiring.")
```

roboplan_rrt/test/test_rrt.cpp
```diff
@@ -77,6 +77,93 @@ TEST_F(RoboPlanRRTTest, PlanRRTConnect) {
...
+
+TEST_F(RoboPlanRRTTest, FastReturnUsesFullBudget) {
+  // fast_return is independent of the planner mode: with plain RRT, disabling it should keep
+  // planning past the first solution until the node budget is exhausted.
+  const auto plan_and_count = [this](bool fast_return) {
+    RRTOptions options;
+    options.group_name = "arm";
+    options.max_connection_distance = 0.5;
+    options.max_nodes = 200;
+    options.fast_return = fast_return;
```
eholum, line 166: Maybe worth capping the max planning time to 1.0s or less here, too? I know there's a default but might as well be explicit.

```suggestion
    options.fast_return = fast_return;
    options.max_planning_time = 1.0;
```

eholum (commented):
Minor nits and questions but otherwise this looks great!

roboplan_rrt/src/rrt.cpp
sea-bass: Yeah interesting... should we rewire at each intermediate or when the "extend" operation completes? There's appeal for either, but the intermediate solution is the one closest to "asymptotically optimal"

eholum (approved):
:rocket:

PR #247, opened 2026-06-16
adding cmeel based pip installation

### LINKED ISSUE #101: Add python bindings to pypi
It could be helpful to pip install the planner without building the bindings manually.

TomCC7:
Thanks for the awesome project! Here is a fully vibed but working solution that makes the project pip installable (related to #101 ). Can try and test directly:
```bash
uv pip install 'roboplan @ git+https://github.com/TomCC7/roboplan'
```
which will pull and build the python binding from source.

Hope this can be a good reference!


sea-bass:
Thank you for sharing this! I will try find some time in the next few days to try this out.

BTW the ROS Rolling job is known to be failing because the official ROS Docker image hasn't cut over to Ubuntu 26.04 now.

sea-bass:
So I just tried the `pip install` command locally on a fresh environment and got this. Any ideas?

```
$ pip install --break-system-packages 'roboplan @ git+https://github.com/TomCC7/roboplan'
Defaulting to user installation because normal site-packages is not writeable
WARNING: Skipping /usr/lib/python3/dist-packages/pytest_repeat.egg-info due to invalid metadata entry 'name'
WARNING: Skipping /usr/lib/python3/dist-packages/pytest_repeat.egg-info due to invalid metadata entry 'name'
Collecting roboplan@ git+https://github.com/TomCC7/roboplan
  Cloning https://github.com/TomCC7/roboplan to /tmp/pip-install-itp7gt2_/roboplan_decece5e15e44c989be32c9117eedf36
  Running command git clone --filter=blob:none --quiet https://github.com/TomCC7/roboplan /tmp/pip-install-itp7gt2_/roboplan_decece5e15e44c989be32c9117eedf36
  Resolved https://github.com/TomCC7/roboplan to commit 10e01c34fd3e187b285fa1a0d0699b0ee44dff80
  Running command git submodule update --init --recursive -q
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
  WARNING: Generating metadata for package roboplan produced metadata for project name roboplan-dimos. Fix your #egg=roboplan fragments.
Discarding git+https://github.com/TomCC7/roboplan: Requested roboplan-dimos from git+https://github.com/TomCC7/roboplan has inconsistent name: expected 'roboplan', but metadata has 'roboplan-dimos'
ERROR: Could not find a version that satisfies the requirement roboplan (unavailable) (from versions: none)
ERROR: No matching distribution found for roboplan (unavailable)
```

---

same with `uv` in a fresh env

```
(.test_venv) sebastian@rev07:~$ uv pip install 'roboplan @ git+https://github.com/TomCC7/roboplan'
Using Python 3.12.3 environment at: .test_venv
    Updated https://github.com/TomCC7/roboplan (10e01c34fd3e187b285fa1a0d0699b0ee44dff80)
  × Failed to download and build `roboplan @ git+https://github.com/TomCC7/roboplan`
  ╰─▶ Package metadata name `roboplan-dimos` does not match given name `roboplan`
```

TomCC7:
Oops sorry I made a couple of changes after opening the PR so that I can publish to pypi with a different name. You can either try to install from a previous commit (https://github.com/open-planning/roboplan/commit/05db7d4190665125035326b63edd205837fb0c24), or just remove the "roboplan" package name in the install command since the package name on my branch was renamed to roboplan-dimos.

sea-bass:
oh wow, yes -- confirming that removing the `roboplan @` worked and I could run a few examples. Very, very cool.

sea-bass:
Review-wise, I think I'm a little terrified to put so many custom shell scripts and `pyproject.toml` files in the main repo.

What I'm thinking right now is: what if we make a separate repository in the `open-planning` organization that pulls in the code, adds the necessary patches/modifications, and has a CI job we can kick off to grab the latest `roboplan` version and publish it to PyPi?

It seems complicated, but it's one way we can explore not sticking a ton of files into here... thoughts?

TomCC7:
> Review-wise, I think I'm a little terrified to put so many custom shell scripts and `pyproject.toml` files in the main repo.
> 
> What I'm thinking right now is: what if we make a separate repository in the `open-planning` organization that pulls in the code, adds the necessary patches/modifications, and has a CI job we can kick off to grab the latest `roboplan` version and publish it to PyPi?
> 
> It seems complicated, but it's one way we can explore not sticking a ton of files into here... thoughts?

Hey tried to cleanup a bit so patches don't spam up original cmake files. I think it's better to just keep everything together to make release easier. With the current release ci file, new pypi packages will be built and published automatically whenever you create a new release on github. If you still prefer using a separate repo I'm also happy to help :)

packaging/cmeel/libroboplan/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+
+[tool.cmeel]
+source = "../../python"
+has-sitelib = false
+run-tests = false
+configure-args = [
+  "-DROBOPLAN_CMEEL=ON",
+  "-DBUILD_PYTHON_BINDINGS=OFF",
+  "-DBUILD_TESTING=OFF",
+  "-DBUILD_TESTING_OINK=OFF",
```
sea-bass: I believe this flag is disabled based on `BUILD_TESTING`. If you omit this everywhere, does it still work?

sea-bass (commented):
It does make sense to keep it all together if possible! Will take another look when I can.

packaging/python/cmake/roboplan_python_packaging.cmake
```diff
@@ -0,0 +1,230 @@
...
+  )
+  set(ROBOPLAN_CMEEL_PREFIX "${ROBOPLAN_CMEEL_PREFIX}" PARENT_SCOPE)
+  if(ROBOPLAN_CMEEL_PREFIX)
+    list(PREPEND CMAKE_PREFIX_PATH "${ROBOPLAN_CMEEL_PREFIX}")
+    set(CMAKE_PREFIX_PATH "${CMAKE_PREFIX_PATH}" PARENT_SCOPE)
+  endif()
+endfunction()
+
+function(roboplan_ensure_hpp_fcl_target)
+  find_package(hpp-fcl CONFIG QUIET)
```
sea-bass: For this whole bit, hppfcl is the old version of the `coal` library / headers name. 

Also we are looking for files in `/opt/ros` which should have nothing to do with standalone packaging.

I wonder if we can omit a bunch of lines from here knowing we don't need to build on old pinocchio versions / with ROS?

packaging/python/cmake/roboplan_python_packaging.cmake
```diff
@@ -0,0 +1,230 @@
...
+    ALIASES roboplan_example_models::roboplan_example_models=roboplan_example_models)
+  roboplan_register_build_tree_package(roboplan
+    ALIASES roboplan::roboplan=roboplan roboplan::filters=filters)
+  roboplan_register_build_tree_package(roboplan_simple_ik
+    ALIASES roboplan_simple_ik::roboplan_simple_ik=roboplan_simple_ik)
+  roboplan_register_build_tree_package(roboplan_oink
+    ALIASES roboplan_oink::roboplan_oink=roboplan_oink)
+  roboplan_register_build_tree_package(roboplan_rrt
+    ALIASES roboplan_rrt::roboplan_rrt=roboplan_rrt)
+  roboplan_register_build_tree_package(roboplan_toppra
```
sea-bass, line 54: Yesterday we also added a new library /  package, `roboplan_cartesian_planning`. Please check it.

packaging/python/cmake/roboplan_python_packaging.cmake
```diff
@@ -0,0 +1,230 @@
...
+      yaml-cpp)
+    find_library(ROBOPLAN_UNIFIED_${library}_LIBRARY NAMES ${library})
+    if(ROBOPLAN_UNIFIED_${library}_LIBRARY)
+      install(FILES "${ROBOPLAN_UNIFIED_${library}_LIBRARY}" DESTINATION lib)
+      file(REAL_PATH "${ROBOPLAN_UNIFIED_${library}_LIBRARY}" ROBOPLAN_UNIFIED_${library}_REAL_LIBRARY)
+      install(FILES "${ROBOPLAN_UNIFIED_${library}_REAL_LIBRARY}" DESTINATION lib)
+    endif()
+  endforeach()
+
+  foreach(library IN ITEMS
```
sea-bass: The hard-coded version numbers here are not ideal... any way to avoid this?

packaging/python/README.md
```diff
@@ -0,0 +1,17 @@
+# Python packaging
```
sea-bass, line 1: Can you add instructions on how to run/test this process locally, for developers /contributors?

TomCC7:
hi @sea-bass , addressed your concerns. also updated package names to target `roboplan`. here's the instruction on pypi.org registration:
1. Go to https://pypi.org/manage/account/publishing/
2. add a new pending publisher.
Use these values for upstream:
PyPI project name: roboplan
Owner: open-planning
Repository name: roboplan
Workflow name: release.yml
Environment name: pypi

after this and the branch is merged. whenever you create a new release on github the release flow will run and publish automatically.

.github/workflows/build-pythonbinding.yml
```diff
@@ -0,0 +1,111 @@
...
+              raise SystemExit(f"Unexpected package name: {project['name']}")
+          print(f"version={project['version']}")
+          PY
+      - name: Check tag matches package version
+        if: startsWith(github.ref, 'refs/tags/v')
+        env:
+          PACKAGE_VERSION: ${{ steps.package.outputs.version }}
+          RELEASE_TAG: ${{ github.ref_name }}
+        run: |
+          TAG_VERSION="${RELEASE_TAG#v}"
```
scastro-nasa: unfortunately we don't use `v` as the tag; our tags are named `0.4.0`, etc.

https://github.com/open-planning/roboplan/tags

... also, if someone for some reason decided to make a tag that is is called `victory`, let's say, this would trigger it too right?

Is there a way to detect that the tag is named `<NUMBERS>.<NUMBERS>.<NUMBERS>`, to match our current style? And how the ROS build process generates them.

.github/workflows/build-pythonbinding.yml
```diff
@@ -0,0 +1,111 @@
...
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+          persist-credentials: false
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
```
scastro-nasa: is this version absolutely necessary, especially since it's not one of the stable "even" ones like 3.12, 3.14, etc.

.github/workflows/build-pypi-wheels.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+          persist-credentials: false
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v4.1.0
+        env:
+          CIBW_ARCHS_LINUX: x86_64
+          CIBW_BUILD: "cp310-* cp311-* cp312-* cp313-* cp314-*"
```
scastro-nasa, line 124: wonderful ✔️

.github/workflows/build-pypi-wheels.yml
```diff
@@ -0,0 +1,111 @@
...
+          CIBW_SKIP: "pp* *-musllinux*"
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux_2_28
+          CIBW_ENVIRONMENT: >-
+            CMAKE_BUILD_PARALLEL_LEVEL=2
+            MAKEFLAGS="-j2"
+            NINJAFLAGS="-j2"
```
scastro-nasa, line 131: if these are specified in here, then the `env` stuff at the top of the file can be removed yeah?

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+          - pypi
+        default: testpypi
+
+permissions:
+  contents: read
+
+jobs:
+  build_python_bindings:
+    name: Build Python bindings
+    uses: ./.github/workflows/build-pythonbinding.yml
```
scastro-nasa: nit: Can we rename this to `build-pypi-wheels.yml` or something?

because many other workflows also build the python bindings; this is specifically for PyPi releasing.

packaging/cmeel/roboplan/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+
+[tool.cmeel]
+source = "../../python"
+run-tests = false
+configure-args = [
+  "-DROBOPLAN_CMEEL=ON",
+  "-DBUILD_STANDALONE_PYTHON_BINDINGS=ON",
+  "-DBUILD_PYTHON_BINDINGS=ON",
+  "-DBUILD_TESTING=OFF",
+  "-DBUILD_TESTING_OINK=OFF",
```
scastro-nasa: can this be removed?

packaging/python/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+  endif()
+  if(TARGET roboplan_rrt::roboplan_rrt AND NOT TARGET roboplan_rrt)
+    add_library(roboplan_rrt ALIAS roboplan_rrt::roboplan_rrt)
+  endif()
+  if(TARGET roboplan_toppra::roboplan_toppra AND NOT TARGET roboplan_toppra)
+    add_library(roboplan_toppra ALIAS roboplan_toppra::roboplan_toppra)
+  endif()
+  if(TARGET roboplan_cartesian_planning::roboplan_cartesian_planning AND NOT TARGET roboplan_cartesian_planning)
+    add_library(roboplan_cartesian_planning ALIAS roboplan_cartesian_planning::roboplan_cartesian_planning)
+  endif()
```
scastro-nasa: why do we need these conditionals? What causes this to be true sometimes; false other times?

packaging/cmeel/roboplan/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+
+[project.urls]
+Homepage = "https://github.com/sea-bass/roboplan"
+
+[tool.cmeel]
+source = "../../python"
+run-tests = false
+configure-args = [
+  "-DROBOPLAN_CMEEL=ON",
+  "-DBUILD_STANDALONE_PYTHON_BINDINGS=ON",
```
scastro-nasa: what are these args for? Everywhere in this PR I see that they are always ON, so if we don't have a reason to set them to OFF, is it worth removing these and also a lot of the branching logic that comes with them? Or am I missing something.

.github/workflows/build-pypi-wheels.yml
```diff
@@ -0,0 +1,111 @@
...
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+          persist-credentials: false
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v4.1.0
+        env:
+          CIBW_ARCHS_LINUX: x86_64
```
scastro-nasa: am I correct in seeing that this will only build wheels for Linux on x86, but no other platforms/OS? This is fine, just checking how easy it would be e.g. to add macOS / arm64 support, for example... given that this already works in conda/pixi, it won't be impossible.

scastro-nasa (commented):
Thanks, you're right that this is trending in a good direction where we can have the packaging in here. I left some small comments throughout, but have 3 major ones to look at:

1. We don't use `vX.Y.Z` tagging, it's just `X.Y.Z`, so these rules will have to change.
2. Should we be continuously testing that the PyPi packaging doesn't get broken? I think it will be very easy to break with changes. I'm thinking maybe the packaging CI job runs with every PR, but then it doesn't publish to `testpypi` or `pypi`; it just stops at running the tests.
3. Are some of these variables like `BUILD_STANDALONE_PYTHON_BINDINGS` and `BUILD_CMEEL` ever turned off in these scripts? If not, can we just remove them?

.github/workflows/release.yml
scastro-nasa: wait, if we do it this way will it re-run all the jobs in the other YAML file, or is it smart enough to know it should just take the completion of the previous jobs and continue from here?

However, I think this answers one of my top-level questions about the wheels always being built, just that release happens on a tagged version.

sea-bass:
ugh I accidentally used my work account. sorry

.github/workflows/build-pythonbinding.yml
TomCC7: changed to 3.12

.github/workflows/build-pypi-wheels.yml
TomCC7: Hey should be good? I guess they don't have a default macos x86 runner anymore lol

sea-bass:
it might be the case that macOS x86 doesn't really have any github active runners. Probably pretty OK to remove that since all modern macs use M series chips (which are ARM)

tests/unified_python_package_test.py
```diff
@@ -129,6 +91,8 @@ def test_root_cmake_superbuild_adds_all_python_binding_packages_in_order() -> No
 
     assert "roboplan_configure_scikit_build_prefix()" in source
     assert "roboplan_configure_unified_python_wheel()" in source
+    assert "ROBOPLAN_CMEEL" not in source
+    assert "BUILD_STANDALONE_PYTHON_BINDINGS" not in source
```
sea-bass: if these are not even options anymore, you dont even need to assert them. Glad to see all these cleanups btw

packaging/python/README.md
```diff
@@ -0,0 +1,74 @@
...
+Release wheels are built by `.github/workflows/build-pypi-wheels.yml` with
+cibuildwheel and are smoke-tested by importing the `roboplan` namespace and
+every compiled submodule. That workflow runs on pull requests so wheel build
+breakage is caught before release. `.github/workflows/release.yml` invokes the
+same build workflow from the tagged commit, downloads those fresh artifacts, and
+publishes them through PyPI trusted publishing.
+
+The wheel target matrix is Linux x86_64, Linux aarch64, macOS x86_64, and macOS
+arm64. Linux wheels use `manylinux_2_28`; macOS wheels are built on native Intel
+and Apple Silicon runners. Every wheel job runs the same import-smoke validation.
```
TomCC7:
```suggestion
```

packaging/python/README.md
```diff
@@ -0,0 +1,74 @@
...
+```bash
+export CMAKE_BUILD_PARALLEL_LEVEL=2
+export MAKEFLAGS=-j2
+export NINJAFLAGS=-j2
+```
+
+Source-build and import-test the unified wheel path:
+
+```bash
+uv venv --seed --python 3.13 /tmp/roboplan-wheel-check
```
TomCC7:
```suggestion
uv venv --seed /tmp/roboplan-wheel-check
```

packaging/python/README.md
```diff
@@ -0,0 +1,74 @@
...
+on hosted runners:
+
+```bash
+CIBW_ARCHS_LINUX=x86_64 \
+CIBW_BUILD='cp313-*' \
+CIBW_SKIP='pp* *-musllinux*' \
+CIBW_MANYLINUX_X86_64_IMAGE=manylinux_2_28 \
+CIBW_ENVIRONMENT='CMAKE_BUILD_PARALLEL_LEVEL=2 MAKEFLAGS="-j2" NINJAFLAGS="-j2"' \
+uvx --from cibuildwheel cibuildwheel --platform linux
+```
```
TomCC7, line 51:
```suggestion
```

sea-bass (approved):
@TomCC7 I just went through the local testing and made a few changes, and this seems good to go. Local build and test worked, and I added a few more README instructions to get an example running.

Also added a PyPi trusted publisher re: your instructions, so that should be set once we queue up a 0.5.0 release. Though I may tag a dummy version for now just to test it.

Finally, THANK YOU for doing this very unglamorous task. This is great. I'd be happy to hear how you guys are using this library for dimensionalOS some time.

TomCC7:
Glad to help! We've done some benchmarking and get a very positive result on roboplan. Can share more thoughts once we use roboplan extensively for a while!

sea-bass:
ok I got an example version of RoboPlan on here: https://pypi.org/project/roboplan/

I installed it and had to clear 2 issues.

First, I had issues with my host's `tinyxml2` version when importing pinocchio:

```
Python 3.12.3 (main, Mar 23 2026, 19:04:32) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pinocchio
  File "<stdin>", line 1, in <module>
  File "/home/sebastian/.local/lib/python3.12/site-packages/cmeel.prefix/lib/python3.12/site-packages/pinocchio/__init__.py", line 19, in <module>
    from .pinocchio_pywrap_default import *
ImportError: libtinyxml2.so.11: cannot open shared object file: No such file or directory
```

I was able to resolve this by creating a symlink:

```
sudo ln -s /usr/lib/x86_64-linux-gnu/libtinyxml2.so.10 /usr/lib/x86_64-linux-gnu/libtinyxml2.so.11
```

Secondly, I got errors again importing Pinocchio due to my having NumPy 2.5.0 installed. But when I downgraded to NumPy 1.2.6, I was able to run an example!

EDIT: If I install it in a clean `uv` venv it all works perfectly out of the box... so yeah, will be recommending that to people :)

PR #248, opened 2026-06-18
Add pixi Support for osx-arm64 and missing docs pages

eholum:
A couple of things are/were missing...

* Didn't have the platform in the pixi.toml
* Broke linux specific deps into their own logic
* Fixing test discovery and models lookup

I quick sanity checked this in my linux VM, but might be worth having more people step through this if anyone else out there has a mac? Verified,

```
pixi run build_all
pixi run test_all
pixi run test_benchmarks
pixi run roboplan_examples/python/example_oink.py 
```


eholum:
Jazzy just be borking?

```
11.74 Fetched 151 MB in 2s (64.1 MB/s)
11.74 E: Failed to fetch http://packages.ros.org/ros2/ubuntu/pool/main/r/ros-jazzy-pinocchio/ros-jazzy-pinocchio_3.9.0-1noble.20260225.233616_amd64.deb  404  Not Found [IP: 64.50.233.100 80]
11.74 E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
11.74 ERROR: the following rosdeps failed to install
11.74   apt: command [apt-get install -y ros-jazzy-pinocchio] failed
```

sea-bass:
> Jazzy just be borking?

Jazzy had a sync today, time to blow up some caches

roboplan/test/CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ if ( ament_cmake_gtest_FOUND )
     ament_add_gtest_test(test_urdf_extended_limits)
 else()
     include(GoogleTest)
+    set(CMAKE_GTEST_DISCOVER_TESTS_DISCOVERY_MODE PRE_TEST)
```
sea-bass: note that the use of `gtest_discover_tests` was bad, so I switched to `gtest_add_tests` in https://github.com/open-planning/roboplan/pull/240/changes#r3416354530 ... but like, should I just pull that out of that massive PR, or do you just want to bring it in here? Either / or

roboplan_example_models/src/resources.cpp
```diff
@@ -19,7 +20,16 @@ std::filesystem::path get_install_prefix() {
...
+  const auto prefix = lib_path.parent_path().parent_path();
+
+  // For compiled installs with symlinks, dladdr may follow and break (e.g. on MacOS).
+  // If the path doesn't exist we must rely on a compile time prefix.
+  if (std::filesystem::exists(prefix / "share" / "roboplan_example_models")) {
+    return prefix;
+  }
+
+  // Fall back to compile-time install prefix.
+  return ROBOPLAN_INSTALL_PREFIX;
```
sea-bass, line 32: oh man, who knows if this will break in other build cases... like conda... but let's cross that bridge when we get there

sea-bass (commented):
LGTM, just the thing about `gtest_discover_tests` not being necessary

roboplan/test/CMakeLists.txt
eholum: oooooh yeah ok that makes sense.

There are dumb differences in how macOS loads these libraries which was borking test discovery. Let me just sanity check those changes here too...

roboplan_example_models/src/resources.cpp
eholum: Ugh. _Theoretically_  it shouldn't break any "normal" install mechanisms, since that directory should exist unless the lookup follows odd symlink installs. But also it might break everything because why not?

sea-bass (approved):

mini-1235:
> I quick sanity checked this in my linux VM, but might be worth having more people step through this if anyone else out there has a mac?

Hi! I can confirm this works on my macOS, thanks for adding osx-arm64 support, it makes my life a lot easier :)

Question: should we add a macOS job to CI, so future PRs don't accidentally break this?

sea-bass:
> Question: should we add a macOS job to CI, so future PRs don't accidentally break this?

Yes, this would be good to have!

eholum:
Are macOS runners free for public repos? If so then yes... would probably be worth adding!

I'm also still poking at an odd issue with `qhull` not being found. I ran into on one machine but not the other... and it seems to be transient. I don't think it's a blocker because a reinstall just fixed it.

sea-bass:
> Are macOS runners free for public repos? If so then yes... would probably be worth adding!

Apparently yes! 

https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories

> I'm also still poking at an odd issue with `qhull` not being found. I ran into on one machine but not the other... and it seems to be transient. I don't think it's a blocker because a reinstall just fixed it.

Hmm, is this a host machine resolution thing? Like do we need to add a new dependency? Or it could be that the old machine was frozen to e.g. an older Pinocchio version

eholum:
> Hmm, is this a host machine resolution thing? Like do we need to add a new dependency?

I thought there was an issue with `qhull-static` but I'm going to ask the people that know a lot more about this than me: https://github.com/conda-forge/qhull-feedstock/issues/24

> Apparently yes!

Unbelievable. Let's go: https://github.com/open-planning/roboplan/actions/runs/27875833214/job/82495048150?pr=248

eholum:
Also I blame myself, https://github.com/open-planning/roboplan/actions/runs/27875833214/job/82495048150?pr=248, of course this isn't guaranteed...

sea-bass:
Blargh, sorry for the flaky new RRT tests. I will fix these later by making them deterministic.

eholum:
> Blargh, sorry for the flaky new RRT tests. I will fix these later by making them deterministic.

There's a question here of RRT* should always outperform RRT, but RRT*-connect _might_ not always produce a better result than RRT-connect given finite time / number of nodes?

PR #250, opened 2026-06-18
Make `ConfigurationTask` runtime tunable and add `AccelerationLimit`

### LINKED ISSUE #249: Allow OInK to be used to track joint position targets
To do this, you can remove the use of a `FrameTask` and make `ConfigurationTask` tunable at runtime in the same way.

Then, we want a demo showing following a joint action chunk but sticking to limits (let's say collision, position, and velocity). Modifying the existing action chunk demo to use OInK in this way will be good.

sea-bass:
This PR started by ensuring we could use OInK as just a joint tracker with constraints, as described in #249 (which works).

Also added a reimplementation of the acceleration limit from Pink. The `AccelerationLimit` causes some gnarly oscillations, even when I turned up the maximum values. So I've left them disabled by default.

Closes #249

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -152,6 +152,8 @@ jobs:
           fi
           echo "✅ .pyi stub files look good!"
       # TODO: Decide on a method for automated benchmark performance testing.
+      # TODO: These are currently running in debug mode due to `-e ci`. We should fix this.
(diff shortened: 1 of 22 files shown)


sjahr (approved):
Looks good to me ✔️  I like that we keep the acceleration limits optional for the example.

PR #252, opened 2026-06-20
Fix RRT connect tree joining for RRT*

eholum:
Tracked down the failures from https://github.com/open-planning/roboplan/pull/248#issuecomment-4758819431 to issues in joining trees for some set of start / goal poses. Basically, if the connected nodes ended up being the same (because of the extend), then the re-parenting logic that was there could end up connecting nodes that were actually further apart. 

This change checks if the joining is happening in the same place, then just side steps the whole parenting thing completely. Existing tests seem to cover this pretty well...

Also left hints about the seeds that causes failures for some start / goal poses.


sea-bass (approved):
Makes sense, thank you for digging into this edge case!

PR #253, opened 2026-06-21
Store list of link names in group info

sea-bass:
This PR allows storing link information in the scene's per-group `JointGroupInfo`.

This is useful for visualization applications where we only want to render markers for a specific subset of a full robot model. See https://github.com/open-planning/roboplan-ros/issues/48 for more info, and https://github.com/open-planning/roboplan-ros/pull/50 for the `roboplan-ros` PR.

(EDIT: some more flakes in RRT testing squashed in the process, some of which were already there but RRT* made them far more likely due to incorporating 500-node limits)

### DIFF
--- a/roboplan/test/test_joints.cpp
+++ b/roboplan/test/test_joints.cpp
@@ -83,6 +83,9 @@ TEST_F(RoboPlanJointTest, SceneProperties) {
   ASSERT_EQ(full_group_info.v_indices.size(), 2);
   ASSERT_EQ(full_group_info.nq_collapsed, 2);
   ASSERT_TRUE(full_group_info.has_continuous_dofs);
+  // The default group should contain every link in the model.
+  EXPECT_THAT(full_group_info.link_names,
+              ::testing::UnorderedElementsAre("base_link", "link1", "link2", "link3"));
 
   const auto arm_group_info = scene_->getJointGroupInfo("arm").value();
   const std::vector<std::string> expected_arm_joint_names = {"revolute_joint", "mimic_joint"};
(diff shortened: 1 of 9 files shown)


roboplan/test/test_joints.cpp
```diff
@@ -91,6 +94,33 @@ TEST_F(RoboPlanJointTest, SceneProperties) {
   ASSERT_EQ(arm_group_info.v_indices.size(), 1);
   ASSERT_EQ(arm_group_info.nq_collapsed, 1);
   ASSERT_FALSE(arm_group_info.has_continuous_dofs);
+  // The arm group lists revolute_joint and mimic_joint, whose child links are link2 and link3.
+  EXPECT_THAT(arm_group_info.link_names, ::testing::UnorderedElementsAre("link2", "link3"));
```
eholum, line 98: There are so many more utility functions in `::testing::` than I ever realized!

eholum (approved):
Neat! Tested the corresponding ROS PR and seems good to ship both all the way down.

I get why pinocchio doesn't want to deal with any of this. Joint groups are a mess.

sea-bass:
> I get why pinocchio doesn't want to deal with any of this. Joint groups are a mess.

Pinocchio has this `buildReducedModel()` function that turns a bunch of joints you don't want to deal with into fixed joints, making all the data structures / computations more efficient.

This is a great feature, but it ignores the fact that one could be moving non-group joints through other means in the midst of usage. So yeah, on our side we're choosing to deal with it at the expense of using full-size configuration vectors, Jacobians, etc. all the time.

No free lunch.

PR #254, opened 2026-06-21
Add broadphase culling to collision checking and self-collision barrier

sea-bass:
Inspired by https://github.com/kevinzakka/mink/pull/151, I realized we could also do broadphase culling using Pinocchio's broadphase manager / our `CollisionContext` we added to similarly speed up RRT. The difference here is this uses axis-aligned bounding boxes (AABB) instead of bounding spheres.

Doing some tests, with a broadphase margin of 0.5 meters, the Franka Dual collision barrier was sped up ~2x and the Kinova one (which has pretty fine collision meshes) was ~39x! So yeah a definite win.

It also inches us closer to having thread-safe OInK, though forward kinematics needs to follow suit.

### DIFF
--- a/roboplan/include/roboplan/core/collision_context.hpp
+++ b/roboplan/include/roboplan/core/collision_context.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <optional>
+#include <vector>
 
 #include <Eigen/Dense>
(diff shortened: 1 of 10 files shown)


sjahr (approved):
This is a very cool idea, thanks for adding it.

PR #255, opened 2026-06-21
Run clang-tidy in a separate CI Job

sea-bass:
it's so painfully slow we _have_ to decouple it


sea-bass:
oh, it requires builds. sucks.

PR #256, opened 2026-06-23
Fix OInK instability with failures to converge

sea-bass:
I started seeing the OInK instability behavior again, and pinned it down due to an OSQP state that is somehow not a failure to solve, but a failure to converge?

So now we handle that case and properly zero out the resulting delta_q.

In doing so, the solver was getting "stuck" in these states because of a clamping issue in the `PositionLimit` task.

Other stability/jerk fixes have to do with tweaking the examples and decoupling the animation loop with the IK solving.

### DIFF
--- a/roboplan_examples/python/example_oink.py
+++ b/roboplan_examples/python/example_oink.py
@@ -32,7 +32,7 @@ def main(
     model: str = "ur5",
     task_gain: float = 1.0,
     lm_damping: float = 0.01,
-    regularization: float = 1e-6,
+    regularization: float = 1e-3,
     control_freq: float = 100.0,
     reference_filter_tau: float = 0.1,
     self_collision_num_pairs: int = 0,
@@ -259,8 +259,14 @@ def main(
     def control_loop():
         delta_q = np.zeros(num_variables)
         delta_q_full = np.zeros(model_pin.nv)
+        # Visualization is throttled and runs outside the scene lock so the Viser push to
+        # the browser cannot stretch the control period. The solver assumes a fixed dt, so
+        # keeping the control loop at a steady rate prevents jitter.
+        display_period = max(dt, 1.0 / 30.0)
+        last_display = 0.0
         while running:
             loop_start = time.time()
+            q_to_display = None
 
             # Thread-safe scene access for IK solving
             if not paused:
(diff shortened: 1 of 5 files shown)


roboplan_simple_ik/bindings/python/roboplan/simple_ik/py.typed
sea-bass, line 1: incidentally this file got deleted along the way and kept being re-added when we built

sjahr (changes requested):
In general I think this looks great and I think the fixes to the limits can be merged immediately. However when I tested it with the dual arm setup I ran into a strange bug (or feature!) where it was possible to get a situation where the robots did not move any longer at all until I hit the reset button (see video). I tested main and the issue does not seem to happen there.
Can you try to reproduce it on your end? Maybe it is related to the newly added logic in the python example control loop.

https://github.com/user-attachments/assets/881d5df1-68fd-4483-94c1-6c858b797421

sea-bass:
Nice find! This is exactly the same issue I found with the position limits that required the clipping there, but I didn't realize the barriers were susceptible to the same problem. I'll take a look when I get a chance.

sea-bass:
OK @sjahr so I took a look.

First off: Worth noting that I did go back to `main` and was able to reproduce the problem there as well. So it's not unique to this PR, unfortunately. But good to find it regardless.

If you're curious, this is the issue: https://github.com/open-planning/roboplan/blob/7401cfa97b16bcaa7a651830f483e08efdd28b36/roboplan_oink/src/optimal_ik.cpp#L483-L486

---

I committed a fix that works for you to test; however, it does impact the performance of `enforceBarriers()` in that it re-evaluates the barriers and jacobians more. See video of the fixed behavior:

[Screencast from 2026-06-30 18-23-18.webm](https://github.com/user-attachments/assets/7c2574eb-ce7a-4c8e-86e7-17cdf5a6b88f)

So some questions for you:

1. Do you want this feature knowing it can cost performance? (see last section below)
2. Should we maybe provide an option for users to configure which mode of enforcement they want; either "if any is violated, zero out" vs. "zero out only in the direction barrier function increase"

---

Finally, this is my (Claude's) analysis of this strategy with the collision barrier, which is the expensive one:

  Per-step timing (dual model, nv=14, 100 Hz)

```
  example_oink.py self-collision barrier (16 pairs) — the case you asked about:

  ┌───────────────────────────┬───────────────┬─────────────────┐
  │                           │    solveIk    │ enforceBarriers │
  ├───────────────────────────┼───────────────┼─────────────────┤
  │ Baseline (global zeroing) │ ~0.35–0.39 ms │ ~0.07–0.09 ms   │
  ├───────────────────────────┼───────────────┼─────────────────┤
  │ With masking              │ ~0.38–0.40 ms │ ~0.14–0.16 ms   │
  └───────────────────────────┴───────────────┴─────────────────┘
```

`solveIk` is untouched by this feature (the variation is noise). The masking roughly doubles `enforceBarriers`, ~+0.07 ms/step — but only on the veto path, where it does a second FK evaluation at the current config plus a recompute of the 16-pair barrier Jacobian. When the barrier is satisfied, masking early-continues and costs the same as baseline (~0.07 ms).

This benchmark is the worst case: vetoes fired on 100% of steps. In absolute terms +0.07 ms is <1% of the 10 ms period at 100 Hz (~3% at 400 Hz).

Per-arm `PositionBarrier`s (the example the fix actually targets): `enforceBarriers` is 0.0012 ms → 0.0020 ms — negligible, because position barriers are cheap and rarely hit the veto path.

sjahr (approved):
Thanks for investigating this! I'd be fine with the computation times for now. In order to optimize the solver, I think we need to do a proper profiling anyways. 
I'd be in favor of merging this because it fixes a crucial issue

PR #257, opened 2026-06-23
Add unit tests for roboplan_simple_ik package

### LINKED ISSUE #179: Add unit tests for `roboplan_simple_ik` package
I just realized they don't really exist... hehe

sea-bass:
Adds a few basic tests, plus performs some minor cleanup.

Closes #179

### DIFF
--- a/roboplan/bindings/test/CMakeLists.txt
+++ b/roboplan/bindings/test/CMakeLists.txt
@@ -1,0 +1,11 @@
+# Register the Python binding tests (colcon picks these up via ament). They run
+# against the installed package, so the roboplan namespace is importable.
+find_package(ament_cmake_pytest REQUIRED)
+set(BINDINGS_TESTS_PY
+  test_import
+  test_scene
+  test_filters
+)
+foreach(test_name ${BINDINGS_TESTS_PY})
+  ament_add_pytest_test(${test_name} ${test_name}.py TIMEOUT 300)
+endforeach()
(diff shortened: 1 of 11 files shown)


roboplan_simple_ik/bindings/python/roboplan/simple_ik/py.typed
sea-bass, line 1: incidentally this file was missing from a previous PR that accidentally clobbered it. Adding back here, since otherwise build keeps re-adding it.

roboplan/bindings/test/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# Register the Python binding tests (colcon picks these up via ament). They run
```
eholum, line 1: and `pytest` #justworks.

roboplan_simple_ik/test/test_simple_ik.cpp
```diff
@@ -0,0 +1,122 @@
...
+  goal.base_frame = kBaseFrame;
+  goal.tip_frame = "not_a_frame";
+
+  JointConfiguration start;
+  start.positions = Eigen::VectorXd::Zero(6);
+  JointConfiguration solution;
+  EXPECT_THROW(ik->solveIk(goal, start, solution), std::runtime_error);
+}
+
+TEST_F(RoboPlanSimpleIkTest, CollisionChecking) {
```
eholum, line 92: Philosophically the python bindings tests do the exact same thing, right? Just the downside is if anything changes it's two tests that might need to change.

eholum (approved):
woohoo

roboplan_simple_ik/test/test_simple_ik.cpp
sea-bass: Yeah, just in case something broke in the bindings, for example default param values or bad lambdas etc.

PR #258, opened 2026-06-30
Enable modernize and performance clang-tidy checks

sjahr:
- [x] Address findings once CI flagged them


sea-bass (approved):
it actually seems to pass with no issues 🚀

PR #259, opened 2026-07-02
Switch OInK backend from OSQP to ProxQP

### LINKED ISSUE #245: Consider switching Oink's QP solver backend from OSQP to ProxQP
We have heard this is better, and secondarily its ROS vendor is actively maintained. We've been hitting some walls with non-responsiveness.

sjahr:
Closes https://github.com/open-planning/roboplan/issues/245

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -225,11 +225,10 @@ sccache     = ">=0.10.0,<0.11"
 yaml-cpp    = ">=0.8.0,<0.9"
 
 # Core dependencies
-# Note: the OInK QP solver (proxsuite/ProxQP, header-only) is fetched at a pinned version
-# via CMake FetchContent in roboplan_oink, so it is intentionally not listed here.
 eigen      = ">=3.4.0"
 libtoppra  = ">=0.6.4,<0.7"
 pinocchio  = ">=4.1.0,<5"
+proxsuite = ">=0.7.3,<0.8"
 
 # Python bindings dependencies
 python            = ">=3.11,<3.12"
--- a/pixi.toml
+++ b/pixi.toml
@@ -225,10 +225,11 @@ sccache     = ">=0.10.0,<0.11"
 yaml-cpp    = ">=0.8.0,<0.9"
 
 # Core dependencies
+# Note: the OInK QP solver (proxsuite/ProxQP, header-only) is fetched at a pinned version
+# via CMake FetchContent in roboplan_oink, so it is intentionally not listed here.
 eigen      = ">=3.4.0"
 libtoppra  = ">=0.6.4,<0.7"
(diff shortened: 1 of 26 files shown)


pixi.toml
```diff
@@ -225,10 +225,11 @@ sccache     = ">=0.10.0,<0.11"
 yaml-cpp    = ">=0.8.0,<0.9"
 
 # Core dependencies
+# Note: the OInK QP solver (proxsuite/ProxQP, header-only) is fetched at a pinned version
```
sea-bass: Pretty sure proxqp is available via conda

roboplan_oink/CMakeLists.txt
```diff
@@ -21,69 +22,29 @@ find_package(roboplan REQUIRED)
...
+set(BUILD_WITH_VECTORIZATION_SUPPORT OFF CACHE INTERNAL "Disable proxsuite vectorization")
+set(BUILD_TESTING OFF)
+
+FetchContent_Declare(
+  proxsuite
+  URL https://github.com/Simple-Robotics/proxsuite/archive/refs/tags/v0.7.3.tar.gz
+  URL_HASH SHA256=e634babff534c8812c6dbc6b022b1d01f9a2a4a9a73cf87a6dc299a4de996904
+  DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+)
+FetchContent_MakeAvailable(proxsuite)
```
sea-bass: proxsuite is available on the ROS buildfarm, why not try use that first? Otherwise the same issues we have with osqpeigen will persist and we won't be able to ship this via binaries

sea-bass (approved):
OK this is awesome @sjahr -- thanks for kicking it off, and I think I just finished it up!

* Don't need to depend on osqp_vendor (whose maintainers are unresponsive) and osqp_eigen (which is not yet on the buildfarm and would anyway rely on osqp vendoring or an alternative solution)
* proxsuite is already on the ROS buildfarm (though not yet synced to kilted, lyrical, rolling)
* Solve times are about 2x faster, and you can feel it!

I did find a regression with the Cartesian planner which was addressed, but besides that it's very solid. I'm honestly good to ship it, but will let you take another look first.

Besides that, I notice that the version of proxsuite that is on the ROS buildfarm (0.6.5) leads to some failures / poorer behavior compared to 0.7.3, so I've asked the maintainers to bloom a new one in https://github.com/Simple-Robotics/proxsuite/issues/472.

sjahr (commented):
Cool, thanks for bringing this over the finish line. I tested your changes and it works as expected. I found one minor thing I addressed: When the solver fails, we should reset the warm-start seed so we make it more likely to successfully solve the next time. Let's merge this 🚀

PR #261, opened 2026-07-09
Fix SimpleIk return logic on timeout

sea-bass:
I was testing out the ROS interactive marker example and found that a lot of my IK calls as I dragged the marker around were failing!

Turns out there was a bug in the logic when the timeout is hit... and it so happened that `fast_return = False` was being used in those demos.

For funsies, test the RoboPlan ROS example with and without this change


eholum (approved):
ooooo d'oh, nice catch...

PR #264, opened 2026-07-09
Make nanobind-dev and python3-dev build dependencies in package.xml

sea-bass:
Saw a buildfarm failure because of missing `python3-dev` on a particular Linux distro. So I guess it's best to be explicit.


eholum (approved):
is this possibly more idl fun?

sea-bass:
> is this possibly more idl fun?

It would possibly touch that too, but no in this case it was that `nanobind` operations failed because there were no Python 3 dev headers. So it seems reasonable, though more explicit than what may be needed by most other linux distros that may already grab these headers transitively (or have them already installed).

PR #268, opened 2026-07-19
Add mujoco_ros2_control option to Franka xacro

sea-bass:
In preparation for a `mujoco_ros2_control` example on the `roboplan-ros` side.

### DIFF
--- a/roboplan_example_models/models/franka_robot_model/fr3.xacro
+++ b/roboplan_example_models/models/franka_robot_model/fr3.xacro
@@ -1,8 +1,10 @@
 <?xml version='1.0' encoding='utf-8'?>
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="fr3">
 
-  <!-- Macro definition for FR3 robot with prefix and parent_link -->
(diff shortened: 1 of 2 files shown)


roboplan_example_models/models/franka_robot_model/fr3.xacro
eholum, line 390: Yeah only downside to lumping these in is that position actuators will complain about these command interfaces. E.g.,

```
[ros2_control_node-3] [ERROR] [1784477158.723351136] [MujocoSystemInterface]: Velocity command interface for the joint : fr3_joint1 is not supported with position actua
tor      
```

Probably doesn't matter enough for now... but could either conditionally include them, or just make the ros2_control xacro optional and pass it in from the mujoco config's urdf. Or we could make the actuators generals or motors...

That said it works so ship it?

eholum (approved):

roboplan_example_models/models/franka_robot_model/fr3.xacro
sea-bass: IMO it's not worth adding conditionals and I'd rather just eat the warnings.

PR #270, opened 2026-07-19
ci: isolate clang-tidy and RRT benchmark jobs

### LINKED ISSUE #267: Improve organization of CI jobs
Right now our RRT benchmarks and clang-tidy checks are sort of thrown in wherever they fit.

Additionally, the pixi `ci` environment builds in debug mode to catch Eigen assertions we may miss in other forms, and the RRT benchmarks are consequently running in debug mode which slows them down a lot unnecessarily.

This issue tracks that we should do passes that deliberately isolate these 3 separate things:

1. `clang-tidy`
2. RRT Benchmarks
3. Building in debug mode

w3lld1:
## Summary

- move clang-tidy into its own Ubuntu job with an independent build and cache
- move RRT benchmarks into a dedicated job that uses the default `RelWithDebInfo` Pixi environment instead of the Debug `ci` environment
- keep the existing Debug build-and-test matrix focused on its normal test and stub checks

## Validation

- `pixi run -e lint lint`
- `pixi install --locked -e default`
- parsed the workflow YAML and verified the three job command paths locally
- compared actionlint output with `upstream/main`; no new diagnostics were introduced

The full C++ builds, clang-tidy run, and RRT benchmark execution are deferred to the GitHub Actions jobs this change isolates.

Fixes #267

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -154,13 +154,61 @@ jobs:
             exit 1
           fi
           echo "✅ .pyi stub files look good!"
+
+  clang_tidy:
+    timeout-minutes: 30
+    runs-on: ubuntu-24.04
+    env:
+      SCCACHE_DIR: ${{ github.workspace }}/.sccache
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Cache sccache
+        uses: actions/cache@v4
+        with:
+          path: ${{ env.SCCACHE_DIR }}
+          key: ubuntu-24.04-clang-tidy-sccache-${{ github.head_ref || github.ref_name }}
+          restore-keys: |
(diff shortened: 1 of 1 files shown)


sea-bass (commented):
Thank you! For the benchmarks I think it makes perfect sense as a separate test for now, even if it's rebuilding, because of the release mode.

For the clang-tidy, maybe we should just reuse the pixi jobs that already exist and just add a boolean variable to the CI matrix, setting it true for 26.04? That way we save ourselves another full pixi rebuild.

w3lld1:
Thanks — I updated the pixi matrix so the Ubuntu 26.04 entry runs clang-tidy after the existing build and test steps, and removed the separate clang-tidy job. This keeps the release-mode RRT benchmark job independent while avoiding a second pixi rebuild for clang-tidy.

Validation:
- workflow YAML parses successfully
- `git diff --check`

sea-bass (approved):
Thanks!

PR #271, opened 2026-07-21
Windows support through Pixi

### LINKED ISSUE #132: Provide alternative to `dlfcn` header usage for Windows builds
I tried enabling conda builds for Windows and ran into issues because of the `dlfcn.h` header in the `roboplan_example_models` package, which is a UNIX only header.

https://github.com/open-planning/roboplan/blob/2805eea484a41f7f0e34334c9f475f92618bd16b/roboplan_example_models/src/resources.cpp#L2

There are alternatives with the `windows.h`  header for Windows systems, so this issue is just to track enabling this with some conditionals.

sea-bass:
This PR adds a new Pixi setup/CI job for Windows, as well as making necessary changes for windows. These include:

* Bringing in https://github.com/dlfcn-win32/dlfcn-win32 for the example models resource lookup
* Minor C++ fixes, e.g., replacing `M_PI` etc. with `std::numbers::pi` and converting `std::filesystem::path` to stringd
* Windows-related CMake fixes

While iterating on this PR, I also setup the conda feedstock PR and that seems to be building well too: https://github.com/conda-forge/roboplan-feedstock/pull/20

Closes https://github.com/open-planning/roboplan/issues/132

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -154,6 +154,11 @@ jobs:
           manifest-path: pixi.toml
       - name: Build and test
         run: |
+          if [ "$RUNNER_OS" = "Windows" ]; then
+            # Bounded parallelism: the heavy bindings translation units can
+            # exhaust the memory of the GitHub Windows runners.
+            export CMAKE_BUILD_PARALLEL_LEVEL=2
+          fi
           pixi run -e ci install_all
           pixi run -e ci build_tests
           pixi run -e ci test_all
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -146,7 +146,7 @@ jobs:
           restore-keys: |
             ${{ matrix.cache-key }}-sccache-main
             ${{ matrix.cache-key }}-sccache
-      - uses: prefix-dev/setup-pixi@v0.8.14
+      - uses: prefix-dev/setup-pixi@v0.10.0
         with:
           cache: true
           cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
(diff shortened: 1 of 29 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -123,8 +123,16 @@ jobs:
           - os: macos-14
             cache-key: macos-arm64
             clang_tidy: false
+          - os: windows-latest
```
eholum, line 126: Ok it is actually extraordinary that this just works...

roboplan_oink/test/tasks/test_frame_task.cpp
```diff
@@ -132,7 +133,7 @@ TEST_F(FrameTaskTest, ErrorWithRotation) {
   pinocchio::SE3 current_pose(current_tform);
 
   // Create target pose with 90 degree rotation around z-axis
-  Eigen::AngleAxisd rotation(M_PI / 2.0, Eigen::Vector3d::UnitZ());
```
eholum, line 135: Silly question but does `M_PI` just not exist on windows or are we just hopping on the numbers train (I'm in)?

eholum (approved):
I... cannot say that I test it. But I believe CI, and native MacOS isn't broken so that's all that really matters.

roboplan_oink/test/tasks/test_frame_task.cpp
sea-bass: it's a POSIX thing, so it's not available by default with Windows/MSVC, but there is a `_USE_MATH_DEFINES` preprocessor conditional you can add. Or, you can use `std::numbers` which probably does this under the hood for us.

PR #272, opened 2026-07-23
Fix Python bindings installs for ROS + Windows

sea-bass:
Fixes a lingering issue with Python bindings installs on ROS + Windows, found with `roboplan-ros`.

Apparently even Pinocchio does stuff like this: https://github.com/stack-of-tasks/pinocchio/blob/f101785266e5367969c22103e2b5a3b59e5ace27/bindings/python/pinocchio/windows_dll_manager.py#L19

Also adds a smoke test to verify this problem pre-emptively.

### DIFF
--- a/.github/workflows/build-pypi-wheels.yml
+++ b/.github/workflows/build-pypi-wheels.yml
@@ -11,6 +11,11 @@ on:
 permissions:
   contents: read
 
+# Ensures that only one workflow runs at a time for this branch
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check_version:
     name: Check package version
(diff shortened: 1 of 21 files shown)


.github/workflows/build-pypi-wheels.yml
```diff
@@ -11,6 +11,11 @@ on:
 permissions:
   contents: read
 
+# Ensures that only one workflow runs at a time for this branch
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
sea-bass, line 18: sneaking this in because I noticed it wasn't the case

sea-bass:
Tested on Windows with this PR...

image-9

eholum (commented):
LGTM!

PR #274, opened 2026-07-23
Look at task

msavchen-nasa:
Trying my hand at making new tasks!

### DIFF
--- a/roboplan_oink/include/roboplan_oink/tasks/look_at.hpp
+++ b/roboplan_oink/include/roboplan_oink/tasks/look_at.hpp
@@ -1,0 +1,142 @@
+#pragma once
+
+#include <limits>
+#include <memory>
+#include <string>
+
+#include <Eigen/Dense>
+
+#include <roboplan/core/scene.hpp>
+#include <roboplan/core/types.hpp>
+#include <roboplan_oink/optimal_ik.hpp>
+
+namespace roboplan {
+/// @brief Optional parameters for FrameTask configuration.
+struct LookAtTaskOptions {
+  /// @brief Cost weight for position error (default: 1.0).
+  double position_cost = 1.0;
+
+  /// @brief Cost weight for orientation error (default: 1.0).
+  double orientation_cost = 1.0;
+
+  /// @brief Proportional gain for error feedback (default: 1.0).
+  double task_gain = 1.0;
+
+  /// @brief Levenberg-Marquardt damping for regularization (default: 0.0).
+  double lm_damping = 0.0;
+
+  /// @brief Maximum position error magnitude in meters (default: unlimited).
+  /// Limits the position error norm to prevent large jumps that can invalidate
+  /// CBF linearization. Recommended: 0.1-0.2m for systems with barriers.
+  double max_position_error = std::numeric_limits<double>::infinity();
+
+  /// @brief Maximum rotation error magnitude in radians (default: unlimited).
+  /// Limits the rotation error norm to prevent large jumps.
+  /// Recommended: 0.5-1.0 rad for systems with barriers.
+  double max_rotation_error = std::numeric_limits<double>::infinity();
+
+  /// @brief Priority level (default: 1). Tasks at higher priority numbers are projected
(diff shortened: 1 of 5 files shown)


roboplan_oink/src/tasks/look_at.cpp
```diff
@@ -0,0 +1,146 @@
...
+      transform_world_to_target.translation() - transform_world_to_frame.translation();
+  frame_to_target.normalize();
+
+  Eigen::Matrix3d rotation_error =
+      Eigen::Quaterniond::FromTwoVectors(look_at_axis, frame_to_target).toRotationMatrix();
+
+  Eigen::Vector3d e_rot = transform_world_to_frame.rotation() * pinocchio::log3(rotation_error);
+
+  error_container.head<3>() = Eigen::Vector3d{0.0, 0.0, 0.0};
+  error_container.tail<3>() = e_rot;
```
scastro-nasa, line 84: you can probably cut down on number of lines -- but more importantly, number of data type conversions -- by using the log6 method as shown here: https://github.com/open-planning/roboplan/blob/7c0df250f49897435733418fc4d898438d36d061/roboplan_simple_ik/src/simple_ik.cpp#L109-L119

roboplan_oink/src/tasks/look_at.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  // Soft saturation of rotation error using tanh for smooth gradients
+  if (std::isfinite(max_rotation_error)) {
+    Eigen::Vector3d rot_error = error_container.tail<kOrientationDimension>();
+    const double rot_norm = rot_error.norm();
+    if (rot_norm > kMinNormForSaturation) {
+      const double scale = max_rotation_error * std::tanh(rot_norm / max_rotation_error) / rot_norm;
+      error_container.tail<kOrientationDimension>() = rot_error * scale;
+    }
+  }
```
scastro-nasa, line 108: how necessary was this?

and question about the comment on CBF linearization, since this is a task and not a barrier?

roboplan_oink/include/roboplan_oink/tasks/look_at.hpp
```diff
@@ -0,0 +1,142 @@
...
+#include <string>
+
+#include <Eigen/Dense>
+
+#include <roboplan/core/scene.hpp>
+#include <roboplan/core/types.hpp>
+#include <roboplan_oink/optimal_ik.hpp>
+
+namespace roboplan {
+/// @brief Optional parameters for FrameTask configuration.
```
scastro-nasa, line 14: a lot of the comments here are still copy-paste from the FrameTask, take a look!

scastro-nasa (commented):
nice!

high level question: Should the "look at" task have a position error component at all? Or is this purely something to keep rotation somewhat controlled? I guess updating the comments in the header file / the PR description to explain the desired behavior of this task would be helpful for me to understand.

msavchen-nasa:
> nice!
> 
> high level question: Should the "look at" task have a position error component at all? Or is this purely something to keep rotation somewhat controlled? I guess updating the comments in the header file / the PR description to explain the desired behavior of this task would be helpful for me to understand.

Technically no, I copy pasted the Frame task and did the smallest number of modification to get the cost function for the Look At task, since I still dont understand all of the moving pieces in the code.

but theoretically there is a position component to the task itself, for example being too close to actually see the object...

roboplan_oink/src/tasks/look_at.cpp
msavchen-nasa: thats part of the micro Lie Theory paper right?
Im in the process of learning about it...

roboplan_oink/src/tasks/look_at.cpp
msavchen-nasa: I will let you know as soon as I understand this code fully.

scastro-nasa: oh, I realize now this is already existing from the frametask implementation -- sorry!

PR #275, opened 2026-07-24
Fix MacOS rpath issues

eholum:
So that colcon doesn't bork


sea-bass (approved):
I saw pics so it happened

eholum:
Screenshot 2026-07-24 at 9 55 48 AM

It's just nicer on native osx

PR #276, opened 2026-07-25
Improve joint limit saturation in SimpleIk

sea-bass:
This makes some fixes to SimpleIk to make it work better with the NASA CLR (UR + linear rails).

The whole investigation uncovered that blindly clamping joint limits could cause bad behavior in IK; specifically that cutting off joints to their position limits while solving the IK integration doesn't properly redistribute the error to the other joints. This was causing the bulk of our crazy joint flips while dragging the interactive marker too!

See the new behavior:

[clr-fixed-ik.webm](https://github.com/user-attachments/assets/f68fc4df-c03a-4637-957d-9913de34615f)

### DIFF
--- a/roboplan_cartesian_planning/src/cartesian_path_planner.cpp
+++ b/roboplan_cartesian_planning/src/cartesian_path_planner.cpp
@@ -618,7 +618,6 @@ CartesianPathPlanner::runServoLoop(const std::vector<FrameTrack>& tracks,
     // real speed (a throttled step naturally slows the profile).
     s = committed_s;
     feedrate = feedrate_eff;
-    // Clamp away solver-epsilon overshoot of the position limits, since the QP only satisfies
     q = q_candidate;
 
     // The velocity constraint keeps each step within the joint velocity limits inside the QP;
--- a/roboplan_cartesian_planning/src/cartesian_path_planner.cpp
+++ b/roboplan_cartesian_planning/src/cartesian_path_planner.cpp
@@ -618,7 +618,9 @@ CartesianPathPlanner::runServoLoop(const std::vector<FrameTrack>& tracks,
     // real speed (a throttled step naturally slows the profile).
     s = committed_s;
     feedrate = feedrate_eff;
-    q = q_candidate;
+    // Clamp away solver-epsilon overshoot of the position limits, since the QP only satisfies
+    // its constraints to within the solver tolerance.
+    q = scene_->clampToValidConfiguration(q_candidate);
 
     // The velocity constraint keeps each step within the joint velocity limits inside the QP;
     // verify the committed step actually does, to guard against solver constraint relaxation or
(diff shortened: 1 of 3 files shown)


roboplan_simple_ik/src/simple_ik.cpp
```diff
@@ -164,12 +190,18 @@ bool SimpleIk::solveIk(const std::vector<CartesianConfiguration>& goals,
       jjt_.noalias() = jacobian_ * jacobian_.transpose();
       jjt_.diagonal().array() += options_.damping;
       vel_(v_indices) = -jacobian_.transpose() * jjt_.ldlt().solve(error_);
+      saturateStep(q);
       if (vel_.hasNaN()) {
         break;
       }
 
       q = pinocchio::integrate(model, q, vel_ * options_.step_size);
-      q(q_indices) = q(q_indices).cwiseMax(lower_position_limits_).cwiseMin(upper_position_limits_);
```
sea-bass, line 172: this was the primary offender!

FYI @olaghattas -- turns out a seemingly harmless change for Stretch had some ripple effects. But with this change, it still works as intended.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -618,7 +618,9 @@ CartesianPathPlanner::runServoLoop(const std::vector<FrameTrack>& tracks,
     // real speed (a throttled step naturally slows the profile).
     s = committed_s;
     feedrate = feedrate_eff;
-    q = q_candidate;
+    // Clamp away solver-epsilon overshoot of the position limits, since the QP only satisfies
+    // its constraints to within the solver tolerance.
+    q = scene_->clampToValidConfiguration(q_candidate);
```
sea-bass: putting in another PR

```suggestion
    q = q_candidate;

File: roboplan_cartesian_planning/src/cartesian_path_planner.cpp [hunk 3]
```diff
@@ -618,6 +618,7 @@ CartesianPathPlanner::runServoLoop(const std::vector<FrameTrack>& tracks,
     // real speed (a throttled step naturally slows the profile).
     s = committed_s;
     feedrate = feedrate_eff;
+    // Clamp away solver-epsilon overshoot of the position limits, since the QP only satisfies
```
sea-bass:
```suggestion
```

roboplan_simple_ik/src/simple_ik.cpp
olaghattas: good catch, I wonder how we didnt see issues with the other robots

sea-bass: I think it still works okay when you're doing live dragging, but now we've started using `SimpleIk` to solve for far-away targets for motion planning and some of this stuff is cropping up.

eholum (approved):
I believe you... also tested and things seem waaaaay more consistent!

PR #277, opened 2026-07-25
Improve Cartesian path planner

sea-bass:
Majorly refactors the Cartesian planner to make it more robust for longer trajectories -- again found on the NASA CLR case.

I guess how do I explain this change? It shifted things from doing a live servo loop with OInK that did slowdowns on the fly, to first doing a fully geometric pass and then afterwards using TOPPRA for trajectory timing and further resolving of Cartesian speeds/accelerations (if using `Bounded` mode).

Had a side benefit that it's an overall code _deletion_, despite burning a few extra lines on adding a `Scene::difference()` method.

[clr_cartesian_plan.webm](https://github.com/user-attachments/assets/77e49b0b-d48b-4646-a8f9-910575364339)

### DIFF
--- a/roboplan_cartesian_planning/src/cartesian_path_planner.cpp
+++ b/roboplan_cartesian_planning/src/cartesian_path_planner.cpp
@@ -54,6 +54,12 @@ constexpr size_t kMaxToppraWaypoints = 10000;
 /// whose estimate is tiny, against tripping the safety cap on benign sub-dt fragments and retries.
 constexpr int kHardCapStepMargin = 1000;
 
+/// @brief Smallest reference advance, relative to the total reference duration, that a throttling
+/// attempt may command.
+/// @details Set to a small value, so that a throttle attempt is considered stalled if it would
+/// re-solve an almost identical problem.
+constexpr double kMinRelativeAdvance = 1e-12;
+
 /// @brief Maximum feedrate (fraction of full speed) the reference may carry through the corner at
 /// `cur` between `prev` and `next`, keeping the Cartesian acceleration within the commanded maxima.
 /// @details The reference is piecewise-linear (linear position interpolation, SLERP orientation),
@@ -226,7 +232,13 @@ void CartesianPathPlanner::buildStaticSolverComponents(
                              "velocity limits: " +
                              maybe_velocity_limits.error());
   }
-  Eigen::VectorXd v_max = maybe_velocity_limits->second.cwiseAbs() * options_.velocity_scale;
+  // The velocity scale belongs to whichever stage owns the final timing:
+  // - Bounded mode emits the traced motion directly, so its steps must respect the scaled limits.
+  // - TimeOptimal mode only resolves path geometry and hands it to TOPP-RA, which applies the scale
+  //   during re-timing.
+  const double resolution_velocity_scale =
+      options_.speed_mode == CartesianSpeedMode::TimeOptimal ? 1.0 : options_.velocity_scale;
+  Eigen::VectorXd v_max = maybe_velocity_limits->second.cwiseAbs() * resolution_velocity_scale;
   if (v_max.size() != num_variables) {
     throw std::runtime_error(
         "Could not initialize Cartesian path planner: velocity limit vector size (" +
@@ -589,8 +601,16 @@ CartesianPathPlanner::runServoLoop(const std::vector<FrameTrack>& tracks,
     bool committed = false;
     double feedrate_eff = feedrate_cmd;
     double committed_s = s;
+    // Halving the feedrate only helps while it still moves the reference. Once an attempt would
+    // advance `s` by less than this, the retry re-solves an unchanged problem (and committing a
+    // step that does not advance `s` at all would spin the loop without progress), so the ladder
+    // stops there and the stall below is declared instead of burning the remaining attempts.
+    const double min_advance = kMinRelativeAdvance * total_time;
     for (int attempt = 0; attempt < options_.max_attempts_per_step; ++attempt) {
       const double s_try = std::min(s + feedrate_eff * options_.dt, total_time);
+      if (s_try < total_time && s_try - s < min_advance) {
+        break;
+      }
       const auto step =
           solveStep(tracks, q, s_try, q_candidate, delta_q, position_error, orientation_error);
       if (!step) {
@@ -618,7 +638,9 @@ CartesianPathPlanner::runServoLoop(const std::vector<FrameTrack>& tracks,
     // real speed (a throttled step naturally slows the profile).
     s = committed_s;
     feedrate = feedrate_eff;
-    q = q_candidate;
+    // Clamp away solver-epsilon overshoot of the position limits, since the QP only satisfiesExpand
+    // commentComment on line R621Resolved its constraints to within the solver tolerance.
+    q = scene_->clampToValidConfiguration(q_candidate);
 
     // The velocity constraint keeps each step within the joint velocity limits inside the QP;
     // verify the committed step actually does, to guard against solver constraint relaxation or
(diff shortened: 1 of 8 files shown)


roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -771,117 +702,65 @@ CartesianPathPlanner::planBounded(const CartesianPath& path, const Eigen::Vector
...
-  toppra_options.velocity_scale = options_.velocity_scale;
-  toppra_options.acceleration_scale = options_.acceleration_scale;
-  toppra_options.max_blend_deviation = options_.toppra_blend_deviation;
-  auto maybe_trajectory = toppra_.generate(joint_path, toppra_options);
-  if (!maybe_trajectory) {
-    return tl::make_unexpected("TOPP-RA time parameterization failed: " + maybe_trajectory.error());
-  }
-  return std::move(maybe_trajectory.value());
+/// @brief Resolves a joint group, throwing the planner's construction error if it does not exist.
+JointGroupInfo resolveJointGroup(const Scene& scene, const std::string& group_name) {
```
eholum: I think this is duplicated in the anonymous namespace above

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -401,95 +366,21 @@ CartesianPathPlanner::buildFrameTracks(const CartesianPath& path,
...
-                                double& position_error, double& orientation_error) {
+CartesianPathPlanner::solveStep(const std::vector<FrameReference>& references,
+                                const Eigen::VectorXd& q, double s, Eigen::VectorXd& q_candidate,
+                                Eigen::VectorXd& delta_q, double& position_error,
+                                double& orientation_error) {
   Oink& oink = *oink_;
 
   // Refresh the scene state to the committed configuration so the Oink tasks read the correct
   // current pose, and retarget every tracking task.
   scene_->setJointPositions(q);
```
eholum, line 428: Right I think we called this out before but this means the scene _must_ be static while this is running, right (ie scene is not thread safe)? Is that documented somewhere?

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -504,200 +395,240 @@ CartesianPathPlanner::solveStep(const std::vector<FrameTrack>& tracks, const Eig
...
+tl::expected<std::vector<Eigen::VectorXd>, std::string>
+CartesianPathPlanner::resolvePath(const CartesianPath& path, const Eigen::VectorXd& q_start_full) {
+  auto references = buildFrameReferences(path, q_start_full);
+  if (!references) {
+    return tl::make_unexpected(references.error());
+  }
+
+  // Walk the path in small steps so every IK solve starts from a near neighbour. The step is a
+  // seeding concern, not an output resolution: a solve that had to cross the whole path at once
+  // would need thousands of iterations and could land on a different IK branch. Which of these
```
eholum: "IK branch" == joint flips or otherwise large joint deltas right? We're basically trying to give oink the best possible chance to find smooth ik solutions for the entirety of the path?

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -504,200 +395,240 @@ CartesianPathPlanner::solveStep(const std::vector<FrameTrack>& tracks, const Eig
...
-            std::to_string(verify_v_max_.maxCoeff()) + ").");
+    parameters.at(k) = s;
+    walked.at(k) = q;
+  }
+
+  // Choose the fewest waypoints that still describe the motion. Every waypoint handed to the time
+  // parameterization becomes a corner it must slow through, and its blend radius is capped by the
+  // adjacent segment, so waypoint count sets the achievable speed directly: an unnecessary sample
+  // costs a near-stop and buys nothing. Keep a sample only when dropping it would let the tool
+  // stray, measured the way the trajectory will actually travel.
```
eholum: No need to accept this just had to do it to check my understanding of this section...

```suggestion
  // Choose the fewest waypoints that still describe the motion. Every waypoint handed to the time
  // parameterization becomes a corner it must slow through, and its blend radius is capped by the
  // adjacent segment, so waypoint count sets the achievable speed directly: an unnecessary sample
  // costs a near-stop and buys nothing. Keep a sample only when dropping it would let the tool
  // stray, measured the way the trajectory will actually travel.
  //
  // Starting from just the first and last sample, we recursively check whether a straight joint-space
  // interpolation between two kept samples stays "close enough" to the reference at all the
  // intermediate samples. If the worst deviation exceeds the budget, that sample must be kept and
  // we split the span. Otherwise every intermediate sample is redundant and can be removed.
  //
  // The budget is half the user's pose tolerance, reserving the other half for the corner-rounding
  // that TOPP-RA applies during time parameterization.
```

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -771,117 +702,65 @@ CartesianPathPlanner::planBounded(const CartesianPath& path, const Eigen::Vector
...
-                             ? resampleUniform(trace_positions, num_target_waypoints, *scene_,
-                                               joint_group_info_.q_indices)
-                             : trace_positions;
-  if (joint_path.positions.size() < 2) {
-    return tl::make_unexpected(
-        "Resolved joint path has fewer than 2 waypoints; the Cartesian path may be degenerate "
-        "or too short to time-parameterize.");
+  const auto resolved = resolvePath(path, q_start_full);
+  if (!resolved) {
+    return tl::make_unexpected(resolved.error());
```
eholum: These are probably simple enough to just lump into `plan` right? Not that the switch is that big a deal just slightly harder to jump around in the source file, and really they have the same fundamental steps, right? Both paths resolve the path and time parameterize... and then in the bounded case we have the extra work? Might make the logic slightly more clean without the hard branch.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
```diff
@@ -504,200 +395,240 @@ CartesianPathPlanner::solveStep(const std::vector<FrameTrack>& tracks, const Eig
...
+    angular_speeds.reserve(trajectory.positions.size());
+
+    q_full(joint_group_info_.q_indices) = trajectory.positions.front();
+    Eigen::Matrix4d previous = scene_->forwardKinematics(q_full, tip_frame);
+    for (size_t i = 1; i < trajectory.positions.size(); ++i) {
+      q_full(joint_group_info_.q_indices) = trajectory.positions.at(i);
+      const Eigen::Matrix4d current = scene_->forwardKinematics(q_full, tip_frame);
+      const auto [linear_distance, angular_distance] = poseError(previous, current);
+      linear_speeds.push_back(linear_distance / options_.dt);
+      angular_speeds.push_back(angular_distance / options_.dt);
```
eholum, line 665: Right, so the "peaks" are determined by the average velocity between consecutive waypoints? Given we're timing in joint space could it be that consecutive waypoints have a wide ranging tip velocity between them? Like could it move at 0.01 m/s for a second and then spike to 100 m/s for an instant?

eholum (commented):
Read through the new version to understand the change in approach... it took a bit and through in some comments and questions. I kind of read it as three steps (or two with a +). Is this basically it?

1. Figure out what the path is geometrically with oink
2. Time it with toppra
3. If running bounded, check for spikes and smooth it out

I am definitely not familiar enough with these methods to comment on the effectiveness or tradeoffs vs servoing... But it sure _seems_ reasonable to me?

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: oopsie

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: yeah, it's that. De-verbosified it.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: yes but the peaks are computed for a trajectory that is already pretty densely timed to `dt`, so it's not like we're looking at huge changes in Cartesian speed (hopefully...), unless near gnarly singularities.

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: yeah, good idea -- done!

roboplan_cartesian_planning/src/cartesian_path_planner.cpp
sea-bass: added to the header!

eholum (approved):
Nice! Tested this out with CLR, and bounded paths are usually around 0.04s vs time_optimal in 0.02s. So significantly slower but man I do enjoy the silky smooth motions.

Clicking the green button here! But take it with a grain a of salt, which is equivalent to how much I know about how this works...

PR #278, opened 2026-07-28
Pose constraints in RRT

### LINKED ISSUE #234: Consider adding constraint projection as an option to RRT
Source: https://personalrobotics.cs.washington.edu/publications/berenson2009cbirrt.pdf

sea-bass:
[constrained-rrt-ur.webm](https://github.com/user-attachments/assets/47fdee96-35b3-4aca-bd0b-fe9c83f21f80)

Closes https://github.com/open-planning/roboplan/issues/234

### DIFF
--- a/docs/source/concepts/sampling_based_planning.rst
+++ b/docs/source/concepts/sampling_based_planning.rst
@@ -13,6 +13,7 @@ RoboPlan implements the **Rapidly-exploring Random Tree (RRT)** family of planne
    RRT planning for a dual-arm Franka system.
 
 The classic single-tree RRT, the bidirectional RRT-Connect variant, and the asymptotically optimal RRT* are all supported — and RRT* can be combined with RRT-Connect.
+Any of them can additionally be run under :ref:`pose constraints <constrained-planning>` that every configuration on the path must satisfy, such as holding a gripper upright while it moves.
 There are many other variants of RRT, as well as other types of sampling-based planning algorithms.
 We would always appreciate new contributions!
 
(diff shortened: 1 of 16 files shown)


docs/source/concepts/sampling_based_planning.rst
```diff
@@ -86,6 +87,155 @@ With plain RRT or RRT-Connect — which never rewire — disabling ``fast_return
...
+  than silently moving the endpoints you asked for; project them yourself first.
+- **Budget more nodes.** Every projected hop becomes a node, so constrained planning wants a smaller
+  ``max_connection_distance`` and a much larger ``max_nodes`` than unconstrained planning.
+- **``rrt_star`` is worth enabling.** Walking in small hops leaves a visibly zigzagging path.
+  Rewiring reconnects nodes with straight segments, and a long one between two configurations on a
+  curved constraint set leaves it and is rejected — but the constrained extension packs nodes only
+  ``path_step_size`` apart, so most candidates are short enough to be accepted. Those short
+  rewires are enough to collapse the zigzag, and produce roughly a quarter shorter paths on the
+  bundled example.
+- **Do not shortcut a constrained path.** :doc:`path_shortcutting` splices in straight
```
sjahr, line 177: Maybe add a comment here if there are any post processing / optimization steps that work. Is there even one available in Roboplan? 🤔

roboplan_examples/python/example_rrt.py
```diff
@@ -32,7 +32,7 @@ def main(
     model: str = "ur5",
     max_connection_distance: float = 3.0,
     collision_check_step_size: float = 0.05,
-    collision_check_use_bisection: bool = False,
+    collision_check_use_bisection: bool = True,
```
sjahr, line 35: Is this change related to this PR?

roboplan_rrt/include/roboplan_rrt/constraints.hpp
```diff
@@ -0,0 +1,335 @@
...
+
+#include <roboplan/core/scene.hpp>
+#include <roboplan/core/types.hpp>
+
+namespace roboplan {
+
+/// @brief A six-dimensional task space vector, ordered [x, y, z, roll, pitch, yaw].
+/// @details Translations are in meters and rotations in radians. The rotation triplet is the
+/// roll-pitch-yaw decomposition R = Rz(yaw) * Ry(pitch) * Rx(roll), to match Pinocchio.
+using TaskSpaceVector = Eigen::Matrix<double, 6, 1>;
```
sjahr, line 19: Should we have a core type for this?

roboplan_rrt/include/roboplan_rrt/constraints.hpp
```diff
@@ -0,0 +1,335 @@
...
+  std::shared_ptr<Scene> scene_;
+
+  /// @brief The joint group whose degrees of freedom a projection may move.
+  std::string group_name_;
+
+  /// @brief The name of the constrained frame.
+  std::string frame_name_;
+
+  /// @brief The name of the frame the region transform is expressed in ("" means world).
+  std::string reference_frame_;
```
sjahr, line 183: Would it make sense to collect them all in a config struct?

roboplan_rrt/include/roboplan_rrt/constraints.hpp
sjahr, line 1: Nice! Maybe link a reference paper somewhere here as well if people want to understand the concepts better

roboplan_rrt/src/constraints.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+    if (satisfied) {
+      return true;
+    }
+    if (iter >= options_.max_iters) {
+      return false;
+    }
+
+    // Damped least-squares step that cancels the stacked violations.
+    const auto active_jacobian = active_jacobian_.topRows(num_active);
```
sjahr: Nit: Would it make sense to pre-allocate these matrices or use references?

roboplan_rrt/src/constraints.cpp
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    if (satisfied) {
+      return true;
+    }
+    if (iter >= options_.max_iters) {
+      return false;
+    }
+
+    // Damped least-squares step that cancels the stacked violations.
```
sjahr, line 281: Maybe we should have this as a general helper function somewhere. Aren't we doing the same for simple IK?

sjahr (approved):
So cool! 😎  Most of my comments are only nitpicks and minor suggestions, so they shouldn't block the merge. The example works as expected although it took me a bit to understand that the goal pose is randomized and changes when you press the button.

docs/source/concepts/sampling_based_planning.rst
sea-bass: nothing besides "pass the raw path to TOPP-RA" at the moment...

roboplan_examples/python/example_rrt.py
sea-bass: it's not, but I realized in doing this PR that the defaults were inconsistent across the repo... so this PR just puts them all to True!

roboplan_rrt/include/roboplan_rrt/constraints.hpp
sea-bass: it's in the docs but I will link it here too, in the relevant part in the docstring that already mentions the paper -- good to have!

roboplan_rrt/include/roboplan_rrt/constraints.hpp
sea-bass: I'm thinking if we ever use this again in another place, we can make it a core type.

This is just an alias for a matrix anyhow, nothing fancy.

roboplan_rrt/src/constraints.cpp
sea-bass: I looked into this: these are already "views" into the original Eigen matrices, so I think no need to further optimize!

roboplan_rrt/include/roboplan_rrt/constraints.hpp
sea-bass: It's tempting because we've been doing config structs for everything else, like planners, IK solvers, OInK tasks, etc... but it seems this one is right on the threshold where it might make it harder to work with than just loose constructor args.

roboplan_rrt/src/constraints.cpp
sea-bass: oh, nice -- good idea! Implementing.

PR #279, opened 2026-08-01
0.6.0

sea-bass:
Includes an AI-assisted docs/metadata sweep of the repo where it pointed out inconsistencies and I fixed them.


sea-bass:
NOTE: I am landing these changes with the proxsuite related failures because proxsuite 0.7.3 has already made it onto the buildfarm. Let's find out whether I'm being too hasty 😬

PR #282, opened 2026-08-03
fix missing braces warnings

Mattolo4:
## Problem

`std::array<T, N>` holds a single C-array member, so a brace-elided initializer like `{a, b}` triggers `-Wmissing-braces` on GCC. `-Wall` is already enabled repo-wide, so today these are warnings in upstream CI; in any downstream build that promotes them, they are hard build failures.

I hit this building the workspace with `colcon` on ROS 2 Humble at PAL Robotics, where `ament_cmake_pal` injects `-Werror=missing-braces` into every package:

```
roboplan_toppra/src/toppra.cpp:144:42: error: missing braces around initializer for
'std::__array_traits<toppra::BoundaryCond, 2>::_Type' {aka 'toppra::BoundaryCond [2]'}
[-Werror=missing-braces]
  144 |   toppra::BoundaryCondFull bc_full{bc, bc};
      |                                          ^
```

`roboplan_toppra` fails first, which aborts the rest of the workspace. With that fixed, `roboplan_oink` fails the same way at two sites in `barriers/position_barrier.cpp`.

## Change

Explicit inner braces at the three affected sites (`toppra::BoundaryCondFull` is `std::array<BoundaryCond, 2>`). Purely syntactic — the resulting objects are identical, no behaviour change.

## Testing

Ubuntu 22.04, GCC 11.4, ROS 2 Humble, `colcon build`:

- Full workspace builds; previously it stopped at `roboplan_toppra`.
- `colcon test --packages-select roboplan_toppra` passes.
- `clang-format` clean.

Two `roboplan_oink` tests fail in my environment — `OinkTest.InfeasibleQpSolvesClosestFeasibleProblem` and `PositionBarrierTest.BarrierLimitsMotion`, both with *"QP solver reached the maximum number of iterations"*. That is unrelated to this change (which cannot affect runtime behaviour) and looks like a proxsuite version difference on my side. Happy to open a separate issue if useful.

## Note

These sites can regress silently, since they are only warnings under the current CI flags. If you want them enforced, `-Werror=missing-braces` on the CI job would catch it.

### DIFF
--- a/roboplan_toppra/src/toppra.cpp
+++ b/roboplan_toppra/src/toppra.cpp
@@ -141,8 +141,6 @@ PathParameterizerTOPPRA::generateCubicSpline(const toppra::Vectors& path_pos_vec
 
   // Set boundary conditions to zero velocity and acceleration at both endpoints.
   toppra::BoundaryCond bc{2, Eigen::VectorXd::Zero(path_pos_vecs.at(0).size())};
-  // BoundaryCondFull is a std::array, so the inner braces initialize the wrapped C array
-  // explicitly. Relying on brace elision here raises -Wmissing-braces on GCC.
   toppra::BoundaryCondFull bc_full{{bc, bc}};
 
   const auto spline = toppra::PiecewisePolyPath::CubicSpline(path_pos_vecs, times, bc_full);
--- a/roboplan_toppra/src/toppra.cpp
+++ b/roboplan_toppra/src/toppra.cpp
@@ -141,7 +141,9 @@ PathParameterizerTOPPRA::generateCubicSpline(const toppra::Vectors& path_pos_vec
 
   // Set boundary conditions to zero velocity and acceleration at both endpoints.
   toppra::BoundaryCond bc{2, Eigen::VectorXd::Zero(path_pos_vecs.at(0).size())};
-  toppra::BoundaryCondFull bc_full{bc, bc};
+  // BoundaryCondFull is a std::array, so the inner braces initialize the wrapped C array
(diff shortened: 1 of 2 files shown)


roboplan_toppra/src/toppra.cpp
```diff
@@ -141,7 +141,9 @@ PathParameterizerTOPPRA::generateCubicSpline(const toppra::Vectors& path_pos_vec
 
   // Set boundary conditions to zero velocity and acceleration at both endpoints.
   toppra::BoundaryCond bc{2, Eigen::VectorXd::Zero(path_pos_vecs.at(0).size())};
-  toppra::BoundaryCondFull bc_full{bc, bc};
+  // BoundaryCondFull is a std::array, so the inner braces initialize the wrapped C array
```
sea-bass: Can we remove these comments? I think it's clear without them.

I don't know why Claude (or equivalent) is so obsessed with over-sharing details about the way the code was before instead of fixing in silence, lol

sea-bass (commented):
Thank you -- just one minor comment.

roboplan_toppra/src/toppra.cpp
```diff
@@ -141,7 +141,9 @@ PathParameterizerTOPPRA::generateCubicSpline(const toppra::Vectors& path_pos_vec
...
+  // explicitly. Relying on brace elision here raises -Wmissing-braces on GCC.
```
sea-bass:
```suggestion
```

sea-bass (approved):

PR #286, opened 2026-08-15
Migrate to SceneContext class for thread-safe operation

### LINKED ISSUE #285: Consider making scenes thread-safe
Fired up the old brain fryer and got a nice analysis here:
https://claude.ai/code/artifact/7b21e3bd-6ea0-42ae-9e4b-968f5137ffee

sea-bass:
This PR expands out the existing `CollisionContext` into an entire `SceneContext` which contains Pinocchio data/collision data elements to spurn more consistent thread safety across the whole stack.

It means some of the key methods of `Scene` are now overloaded: they have a new option which uses a `pinocchio::Data` instance that the user passes in, whereas the old options internally fall back on using the scene's stored data which is not thread-safe. This hits dual workflows where people either do or do not care about thread safety, balancing usability with... a lot of overloads.

Additionally, several of the nanobind bindings will now release the Python GIL to help in this effort of multiple threads accessing a scene concurrently.

While a good chunk of this PR is renaming "collision context" to "scene context", and similar find/replaces, the additions are still substantial.

Closes https://github.com/open-planning/roboplan/issues/285


roboplan/include/roboplan/core/scene_context.hpp
```diff
@@ -0,0 +1,191 @@
...
+  // the object's address must remain stable for its whole lifetime. Hold one behind a pointer
+  // (e.g. std::unique_ptr) if it needs to be relocated or rebuilt.
+  SceneContext(const SceneContext&) = delete;
+  SceneContext& operator=(const SceneContext&) = delete;
+  SceneContext(SceneContext&&) = delete;
+  SceneContext& operator=(SceneContext&&) = delete;
+
+  /// @brief Checks collisions at `q`, stopping at the first collision. Uses this context's own
+  /// scratch, so it is safe to call concurrently with queries on other contexts or the Scene.
+  bool hasCollisions(const Eigen::VectorXd& q) const;
```
eholum: I guess it's unlikely anyone would be, but is there a reason to not include debugging for the context?

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -87,9 +116,18 @@ class Scene {
   /// active collision pairs at the specified joint configuration.
   /// @details After this call, distances are available via
   /// getCollisionData().distanceResults.
+  /// @note Writes the Scene's shared scratch; not safe to call concurrently. Use
+  /// SceneContext::computeDistances instead.
   /// @param q The joint configuration at which to compute the distances.
   void computeCollisionDistances(const Eigen::VectorXd& q) const;
 
+  /// @brief A counter that changes whenever the collision geometry is modified.
+  /// @details Bumped by addGeometry(), removeGeometry(), updateGeometryPlacement(), and
```
eholum: Does it?

```suggestion
  /// @details Bumped by addGeometry(), removeGeometry(), and
```

eholum (commented):
Took a hard read through the scene context class and I _think_ it looks good. I haven't really thought about alternative mechanisms to make the scene easier to use in multi-threaded processes, but expanding on the collision context makes sense to me. The gainz on joint positions / kinematics are substantial and would open the door to multiple RRTs (and other things) running simultaneously (right?). Also having the built-in mechanism to track when collision contexts are invalidated is handy, as opposed to before.

If we keep going down this route I can take a good look at the algorithms, too. 

Also, in general there are a lot of overly verbose comments - so would probably recommend some trimming.

sea-bass:
> If we keep going down this route I can take a good look at the algorithms, too.

I also don't see anything substantially different. So right now there are 3 modes of operation:

1. Call the original scene methods directly, which uses the scene's non-thread-safe Pinocchio data.
2. Call the new scene methods with actual data objects passed in. This is the most consistent with Pinocchio, but notably doesn't cross the nanobind boundary to Python until Pinocchio is fully nanobinded.
3. Use the collision context methods, which internally call approach 2. above with their context-specific data structures.

The main downside is that 3. increases the code footprint by quite a lot, whereas possibly holding out for a 2.-only world could be helpful. Though on the other hand this wouldn't get rid of the collisioncontext.

Also to be fair, the changes in the algorithms are mostly find/replace, with a couple of GIL releases in the bindings.

> Also, in general there are a lot of overly verbose comments - so would probably recommend some trimming.

the eternal struggle... will get on that

roboplan_rrt/test/test_rrt.cpp
```diff
@@ -328,6 +335,59 @@ TEST_F(RoboPlanRRTTest, TestJoinTrees) {
 /// @details Constrains the UR5 tool frame to hold its z axis near vertical, leaving position and
 /// the rotation about that axis free. The nominal orientation is a half turn about x, so a zero
 /// displacement points the tool straight down.
```
sea-bass: note: this comment needs to be moved down to the correct spot

roboplan/include/roboplan/core/scene.hpp
sea-bass: you're right, but also I'm gonna slash the whole comment in the spirit of desloppifying lol

roboplan/include/roboplan/core/scene_context.hpp
sea-bass: good one, I've added this!

sea-bass:
> If we keep going down this route I can take a good look at the algorithms, too.

Ready for another look from you!

> Also, in general there are a lot of overly verbose comments - so would probably recommend some trimming.

Took a pass at this too, I think I trimmed quite a bit of fat.

roboplan/bindings/src/core.cpp
```diff
@@ -319,6 +324,50 @@ void init_core_scene(nanobind::module_& m) {
...
+           "This context's current joint positions.")
+      .def("setJointPositions", &SceneContext::setJointPositions,
+           "Sets this context's current joint positions.", "q"_a)
+      .def("toFullJointPositions", &SceneContext::toFullJointPositions,
+           "Converts partial joint positions to full ones, filling non-group joints from this "
+           "context's current configuration.",
+           "group_name"_a, "q"_a)
+      .def("isGeometryCurrent", &SceneContext::isGeometryCurrent,
+           "Whether the scene's collision geometry is still the one this context was built from.")
+      .def("getScene", &SceneContext::getScene, nanobind::rv_policy::reference_internal,
```
eholum, line 369: [nanobind::rv_policy::reference_internal](https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4N8nanobind9rv_policy18reference_internalE)

nanobind is so magic

roboplan_examples/python/example_cartesian_planning.py
```diff
@@ -178,15 +178,15 @@ def main(
...
+        acceleration_scale: Scaling (0, 1] applied to joint acceleration limits.
         dt: Output trajectory sample period (s).
         path_size: Side length of the square region the lawnmower covers (m).
         path_num_passes: Number of zigzag passes across the square.
         path_corner_radius: Task-space radius (m) used to round the lawnmower corners. Larger values
             round the corners more, letting the tool carry speed through them (0 = sharp corners,
             clamped per corner so adjacent arcs do not overlap).
         path_corner_arc_step_deg: Angular step (deg) used to discretize each rounded corner arc into
-            chords. Coarse values (e.g. 15) facet the arc into a few straight segments whose kinks
+            chords. Coarse values (e.g., 15) facet the arc into a few straight segments whose kinks
```
eholum, line 189: Mr. Mega-Nit.

eholum (approved):
Took another skim through and LGTM!

A good portion of this approval is coming from empirical evidence that these changes actually work.

roboplan_examples/python/example_cartesian_planning.py
sea-bass: You didn't have an editor telling you this for YEARS lol

PR #288, opened 2026-08-16
Fix OInK position barrier Jacobian

sea-bass:
Found this while debugging something unrelated... but turns out the Jacobian for the position barrier was in the wrong frame, and removing this actually makes it work without needing to fall back on the crutch of `enforceBarriers()`.


roboplan_oink/src/barriers/position_barrier.cpp
```diff
@@ -114,16 +114,15 @@ tl::expected<void, std::string> PositionBarrier::computeBarrier(const Scene& sce
...
-  scene.computeFrameJacobian(q, frame_id, pinocchio::ReferenceFrame::WORLD, full_jacobian);
+  scene.computeFrameJacobian(q, frame_id, pinocchio::ReferenceFrame::LOCAL_WORLD_ALIGNED,
+                             full_jacobian);
 
-  // Pinocchio frame Jacobian layout with WORLD reference frame:
+  // Pinocchio frame Jacobian layout:
   //   Rows 0-2: linear velocity (dp_world/dq) - this is what we need
-  //   Rows 3-5: angular velocity (d_omega_world/dq)
-  // Note: With LOCAL or LOCAL_WORLD_ALIGNED, the ordering may differ
+  //   Rows 3-5: angular velocity
```
sea-bass:
```suggestion
  //   Rows 3-5: angular velocity (d_omega_world/dq)

sjahr (approved):
Nice catch! Tested the example and it works as expected. Should we remove the enforce barrier in a follow-up entirely 🤔 ?

sea-bass:
> Nice catch! Tested the example and it works as expected. Should we remove the enforce barrier in a follow-up entirely 🤔 ?

I thought about removing it, but ... it doesn't really hurt to keep in case someone's running this with a colossal dt or a particularly tricky barrier?

PR #290, opened 2026-08-19
Optionally brake to target in OInK acceleration limit

sea-bass:
Added an opt-in mechanism to reduce the effects of overshooting / instability in the acceleration limits.

You can still get some oscillation by moving the marker super quickly, but otherwise it's a lot better ... though at the expense of some ergonomics. However, it is all optional so 🤷🏻 

accel_braking_target_vs_actual


sjahr (approved):
I like the idea! Since it is only optional I'd say we can merge it just like this 🥲 Did you try visualizing how this actually reduce overshooting? Would be interesting to see the effect in a plot: Target vs actual motion over time

sea-bass:
> I like the idea! Since it is only optional I'd say we can merge it just like this 🥲 Did you try visualizing how this actually reduce overshooting? Would be interesting to see the effect in a plot: Target vs actual motion over time

Good idea -- qualitatively from my testing it was better, but not perfect. Just generated some throwaway code on top of one of the examples and put a plot in the PR description.

PR #292, opened 2026-08-26
Utilities to filter robot model from point cloud

sea-bass:
This adds a utility to filter the robot model (with some padding) from a point cloud, which is key for planning in environments with unstructured obstacles and depth sensing.

Example included:

image

### DIFF
--- a/roboplan_core/include/roboplan/core/robot_body_filter.hpp
+++ b/roboplan_core/include/roboplan/core/robot_body_filter.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstddef>
 #include <memory>
 #include <optional>
 #include <vector>
@@ -38,6 +39,11 @@ struct RobotBodyFilterOptions {
 
   /// @brief The classification test to use.
   RobotBodyFilterMethod method = RobotBodyFilterMethod::NARROWPHASE;
+
+  /// @brief Number of threads used to classify points, or 0 to use all hardware threads. The
+  /// points are split into blocks that the threads pull from a shared queue, so at most one
+  /// thread per block is ever spawned and small clouds are processed serially either way.
+  size_t num_threads = 0;
 };
 
 /// @brief Filters points that lie on or near the robot's own collision geometry.
--- a/roboplan_core/include/roboplan/core/robot_body_filter.hpp
+++ b/roboplan_core/include/roboplan/core/robot_body_filter.hpp
@@ -1,0 +1,128 @@
+#pragma once
+
+#include <memory>
+#include <optional>
+#include <vector>
+
+#include <roboplan/core/scene.hpp>
+
+namespace roboplan {
+
+// The transform type was renamed from Transform3f (hpp-fcl) to Transform3s (coal). Pick the
+// right one for whichever library the coal namespace resolves to (see geometry_wrappers.hpp).
+#if defined(__has_include) && __has_include(<coal/fwd.hh>)
+using CoalTransform = coal::Transform3s;
+#else
+using CoalTransform = coal::Transform3f;
+#endif
+
+/// @brief The test used by RobotBodyFilter to classify points near the robot geometry.
+enum class RobotBodyFilterMethod {
+  /// Exact: after the broadphase AABB cull, each candidate point is checked with a Coal
+  /// narrowphase collision query (point vs. geometry, with the padding as the security margin).
+  /// Exact for every geometry type, including meshes, at the cost of one GJK/BVH query per
+  /// candidate point.
+  NARROWPHASE,
+  /// Conservative: after the broadphase AABB cull, each candidate point is checked against the
+  /// geometry's padded local AABB (an oriented box in world frame). Much faster since it is a
+  /// few arithmetic operations per candidate, but over-removes points near the corners of the
+  /// oriented boxes. The set of points it removes is always a superset of NARROWPHASE's.
+  PADDED_OBB,
+};
+
+/// @brief Options struct for the robot body filter.
+struct RobotBodyFilterOptions {
+  /// @brief Distance, in meters, around the robot's collision geometry within which points are
+  /// considered part of the robot body. Must be non-negative.
+  double padding = 0.05;
+
+  /// @brief The classification test to use.
+  RobotBodyFilterMethod method = RobotBodyFilterMethod::NARROWPHASE;
+};
+
+/// @brief Filters points that lie on or near the robot's own collision geometry.
+///
+/// This is the usual "self filter" that removes the robot's body from a sensor point cloud (or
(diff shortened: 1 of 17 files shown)


roboplan_core/src/core/robot_body_filter.cpp
```diff
@@ -0,0 +1,139 @@
...
+RobotBodyFilter::filterPoints(const Eigen::VectorXd& q, const Eigen::Ref<const PointMatrix>& points,
+                              const std::optional<Eigen::VectorXd>& extra_padding) {
+  const Mask mask = computeMask(q, points, extra_padding);
+  PointMatrix kept(points.rows() - mask.count(), 3);
+  Eigen::Index kept_idx = 0;
+  for (Eigen::Index i = 0; i < points.rows(); ++i) {
+    if (!mask(i)) {
+      kept.row(kept_idx++) = points.row(i);
+    }
+  }
```
EzraBrooks, line 203: Iterating serially here is probably quite slow on large point clouds. This reminds me of when we attempted to use the moveit2 point cloud filtering a few years back and found it untenably slow, especially in a streaming environment.

How large is the point cloud in the test and how fast is this running?

sea-bass: The test point cloud is something like 12k points, and depending on the filtering method it either takes ~2 or ~15 ms on my computer.

Is your suggestion to iterate using something like OpenMP instead?

sea-bass: ok actually OpenMP isn't available on all platforms, so I had Claude sling a more vanilla method of parallelism.

First off, the test cloud was actually more like 130k points.

The ~2 or ~15 ms values above got shrunk down to ~1 and ~3.5 ms, respectively, which is cool.

Still need to do a de-sloppifying pass through all this though.

EzraBrooks: 3.5ms = 285Hz or something like that, right? that would make this fairly usable for servoing if you sampled every 3 steps or something :)

sea-bass: Yeah or you could use the padded OBB method which can give you a couple false point removals in certain edge cases, but then you'd be able to hit 1 kHz!

roboplan_core/src/core/robot_body_filter.cpp
```diff
@@ -0,0 +1,178 @@
...
+      for (auto begin = next_begin.fetch_add(kBlockSize); begin < num_points;
+           begin = next_begin.fetch_add(kBlockSize)) {
+        classify_block(begin, std::min(begin + kBlockSize, num_points));
+      }
+    } catch (...) {
+      const std::lock_guard<std::mutex> lock(error_mutex);
+      if (!worker_error) {
+        worker_error = std::current_exception();
+      }
+      next_begin.store(num_points);  // Drain the remaining blocks so the other workers stop.
```
sjahr, line 145: Is it worth printing here a warning or can we silently accept an exception?

roboplan_core/include/roboplan/core/robot_body_filter.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+private:
+  /// @brief The scene whose robot description (model and collision geometry) is filtered against.
+  std::shared_ptr<Scene> scene_;
+
+  /// @brief The filter options.
+  RobotBodyFilterOptions options_;
+
+  /// @brief The thread cap from the options, with 0 resolved to the hardware thread count.
+  size_t max_threads_;
```
sjahr, line 98: Based on the usage of this, what if this becomes and optional function argument for `computeMask`?

roboplan_core/src/core/robot_body_filter.cpp
```diff
@@ -0,0 +1,178 @@
...
+                             const std::optional<Eigen::VectorXd>& extra_padding) {
+  const auto num_points = points.rows();
+  if (extra_padding && extra_padding->size() != num_points) {
+    throw std::invalid_argument("RobotBodyFilter extra_padding has size " +
+                                std::to_string(extra_padding->size()) + " but there are " +
+                                std::to_string(num_points) + " points.");
+  }
+
+  // Place the robot geometry at the query configuration (runs forward kinematics).
+  pinocchio::updateGeometryPlacements(scene_->getModel(), data_, robot_geom_model_,
```
sjahr, line 65: Do we actually need to store the scene as a member variable? Maybe to reduce the statefulness of the Filter class we could pass the scene as function argument only?

sjahr: Or even just the geom model so the caller is responsible to make sure the model and data is up to date?

sjahr (commented):
Nice work! Some questions before approving it but in general I think this is in great shape

roboplan_core/include/roboplan/core/robot_body_filter.hpp
sea-bass: you'd have to pass it into `filterPoints` as well, which internally calls `computeMask`.

But yes, is it better to pass this in as an arg vs. keep it in the options? I'm sort of biased towards keeping it in options because chances are you aren't going to be changing this at runtime too much, right?

roboplan_core/src/core/robot_body_filter.cpp
sea-bass: I's not silent, I think `worker_error` (if set) is then rethrown a few lines later:

```cpp
  if (worker_error) {
    std::rethrow_exception(worker_error);
  }
```

roboplan_core/src/core/robot_body_filter.cpp
sea-bass: We can't pass just the geom model because Pinocchio doesn't yet have nanobind bindings implemented, but scene could definitely work.

I guess the question is, do we prefer always passing in a fresh scene so we make things a bit less stateful (i.e., less member variables), knowing that this will then have to allocate memory for data/collision data/collision objects at each call... or pass in their own? (which again we can't easily do in Python due to bindings boundaries).

My feeling here is that keeping it stateful might be okay to avoid a lot of allocations and/or mismatched model/data.

sjahr: Makes sense, thanks for your explanation!

sea-bass:
Thanks @sjahr -- should we change anything based on the open questions, or should we proceed as-is given you've resolved the threads?

sjahr (approved):
No, it think we can merge this! Sorry I forgot to approve after resolving the comments.

sea-bass:
Thanks -- I'll merge it later, doing some testing which requires the CI runners to not be hogged, so I canceled CI

PR #295, opened 2026-09-02
Create CMake superbuild to consolidate build config

EzraBrooks:
Move the CMake configuration that's currently inline in the pixi.toml into a superbuild `CMakeLists.txt` and a `CMakePresets.json`. I would've preferred to put those in the project root, but that would break compatibility with colcon's search algorithm for people building this as a source dependency in ROS projects 🫠.

Regardless, this makes a significant impact on the amount of hand-jammed workarounds for not having a superproject in `.clangd`, `build_cmake.bash`, `run_tests.bash`, and `run_clang_tidy.bash` (entirely irrelevant now and deleted).


EzraBrooks:
last TODO here other than sanity check from y'all (and CI) is getting the python stubgen running in this superbuild

.clangd
```diff
@@ -1,42 +1,2 @@
...
-  PathMatch:
-  - roboplan_toppra/.*
----
-CompileFlags:
-  CompilationDatabase: build/roboplan_oink/
-If:
-  PathMatch:
-  - roboplan_oink/.*
----
+  CompilationDatabase: build/
```
EzraBrooks, line 2: Note this is the default value for this, so we could eschew this config file entirely now.

EzraBrooks:
This ended up not being as much of a code deletion as I wanted, because as it turns out generating the Python bindings in a single build is tricky; had to add a special "staging directory" in the build for the Python files to share with each other. The explicit build order in the `pixi.toml` was working around that complexity.

EzraBrooks:
@eholum @sea-bass I think this is about ready. A couple extra issues shook out, including a weird namespacing issue with `coal::coal` on CMake 3.22 (Humble). Yesterday I attempted to pin the dev tooling in the Pixi workspace to Humble versions but had some trouble with transitive dependencies. I might revisit that attempt just so those things don't go unnoticed in local builds.

There's probably more improvement to be done here, but at least now all the CMake build configuration is actually _in CMake_ rather than having a meta build system for the meta build system ;)

docs/source/getting_started/installation.rst
```diff
@@ -94,16 +94,16 @@ Once set up, you can run the ``pixi`` tasks as follows.
...
+    pixi run install
 
-    # This will only build the package (You must have built the dependencies first)
+    # This will only build one package (and its dependencies)
     pixi run build PACKAGE_NAME
 
-    # This will only install the package
-    pixi run install PACKAGE_NAME
+All packages share a single build tree (``build/``), configured in one shot, so there is no separate
+"install one package" operation -- ``pixi run install`` always installs everything.
```
sea-bass, line 106: I was wondering about this when I started reading the changes, but I think it's okay to not have that... rebuilds are fast enough.

superbuild/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+
+# Superbuild: configures every RoboPlan package in one tree, so
+# build/install/test run once instead of once per package. Driven by
+# pixi.toml and scripts/build_cmake.bash. See CMakePresets.json for
+# compiler/linker/build-cache settings.
+#
+# Not at repo root on purpose: colcon would identify a bare CMakeLists.txt
+# there as its own "cmake" package and stop recursing, hiding the real
+# ament packages underneath. This dir has its own COLCON_IGNORE instead
+# (same trick as packaging/COLCON_IGNORE).
```
sea-bass: this line can be omitted imo

superbuild/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+#
+# Each package below is also an independent project resolving siblings via
+# find_package(<name> REQUIRED), so it stays buildable standalone (colcon,
+# conda-forge, its own wheel). roboplan_register_build_tree_packages() makes
+# those calls resolve to the in-tree targets instead of requiring a real
+# install -- see RoboplanSuperbuild.cmake. Same mechanism as
+# ../packaging/python/CMakeLists.txt.
```
sea-bass: same here re: omitting the "same as packaging" bit

sea-bass (commented):
I killed all the caches and reran CI, which I think is making the unrelated failures go away. Somehow I've had to do that a lot for the ROS builds (especially after a distro sync), which might mean there's a bug on our end.

Anyway, I only saw minor things to comment on. The cleanup of things like the pixi.toml and the build_cmake.bash is great, and removing the custom clang-tidy shim is greater. Even if some of the stub gen optimizations did effectively re-add all those saved lines, it's definitely an overall improvement.

docs/source/getting_started/installation.rst
EzraBrooks: You can still do a per-package build. It's just one monolithic install because they all share a build directory now

superbuild/CMakeLists.txt
EzraBrooks: done. LLMs love to yap and create unnecessary cross-file documentation dependencies

superbuild/CMakeLists.txt
EzraBrooks: done. LLMs love to yap and create unnecessary cross-file documentation dependencies

EzraBrooks:
Let me know when this is good to go and I'll squash it myself and make the commit more descriptive.

sea-bass (approved):
LLMs do be speaking yapanese

Also FWIW we squash all PRs by default

EzraBrooks:
> Also FWIW we squash all PRs by default

perpetuating the curse of inscrutable git graphs, but I understand why you need to do this since most OSS contributors are impossible to guide on how to make good commits ;)

sea-bass:
> most OSS contributors are impossible to guide on how to make good commits ;)

That's me!

But in future PRs where you do want a merge with multiple purposeful commits, just holler!

PR #296, opened 2026-09-03
ci: change Python stub generation from imperative to declarative

EzraBrooks:
YAML is not a programming language

which sounds like something the community would make up a tongue in cheek acronym for.. YINAPL?

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -7,7 +7,7 @@ on:
 
   # Nightly job on default (main) branch
   schedule:
-    - cron: '0 0 * * *'
+    - cron: "0 0 * * *"
 
   # Whenever certain branches are pushed
   push:
@@ -22,7 +22,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-
   ubuntu_build_and_test:
     timeout-minutes: 30
     strategy:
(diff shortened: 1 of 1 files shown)


EzraBrooks:
FYI @eholum @sea-bass this changes the job name so will require some GitHub repo settings futzing.

.github/workflows/build_and_test.yml
```diff
@@ -64,11 +63,22 @@ jobs:
...
   ros_build_and_test:
+    name: "Build and Test: ROS ${{ matrix.ros_distro }} ${{ matrix.generate_python_stubs == 'ON' ? 'with' : 'without' }} Python stubs"
     timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
-        ros_distro: [humble, jazzy, kilted, lyrical, rolling]
+        include:
+          - ros_distro: humble
+            generate_python_stubs: "OFF"
```
sea-bass, line 73: can we add a comment explaining that we're not generating stubs here because the nanobind version on ubuntu 22.04 is pre-2.0 and therefore does not have tje stubgen feature?

sea-bass (approved):
I think this is more legible -- thanks!

EzraBrooks:
merged into #299

PR #298, opened 2026-09-03
Merge pixi config into pyproject.toml

EzraBrooks:
Mostly just concatenated the `pixi.toml` config into `pyproject.toml`, prepending all the table names with `tool.pixi.`, then some minor field deduplication.

This way we don't have two Python packaging tools that both need things like their version bumped and their authors list


EzraBrooks:
The long-term "right way" to do this might be to push this all down piecewise into the individual `pyproject.toml` for each package, then have the root package just include them all as Pixi dev dependencies (the `dev` table installs the transitive dependencies of dependencies listed within, but leaves building them up to you), but this is an incremental step and that is not

EzraBrooks:
> The long-term "right way" [...]

see #301

sea-bass (approved):
Nice!

Are all the lockfile additions of `purls` related to moving the pixi stuff to the `pyproject.toml`? Or are they a byproduct of you possibly using a newer pixi version than the last person who locked? Either way I'm fine; just curious.

EzraBrooks:
I think `purl` is a PyPI thing and only appeared when merging the existing dependency graph of the pyproject.toml with the pixi.toml

PR #299, opened 2026-09-03
Consolidate CI hooks

EzraBrooks:
By invoking format and build_pypi_wheels from workflow_call, we ensure that our CI trigger and schedule strategy does not drift from job to job.


EzraBrooks:
This might be considered a matter of taste - but personally I like having all my CI jobs visible in a singular workflow in Actions.

EzraBrooks:
Ah, because of the way the job names are generated (pre-pending their caller and `/`) this will also require changing the required job names in GitHub. So if we like both this and #296 we probably should merge as one

sea-bass (commented):
I like this as well, so we can merge the 2 PRs as one. I can go into the admin settings and change the required checks to pass after the fact.

Let me know when/if you want to consolidate these 2 PRs and we shall get it through.

EzraBrooks:
Would it make sense to also move the current jobs that live at the top level of `build_and_test.yml` into their own reusable workflows, just to get ahead of any future naming problems? That way the whole workflow view in GitHub will already be folderized into format, wheels, Pixi, and ROS builds.

I could go either way personally, just thinking about how to prevent more faffing about with GitHub settings.

sea-bass:
That seems like a good idea for consistency!

.github/workflows/build_and_test.yml
```diff
@@ -1,4 +1,4 @@
-name: build_and_test
+name: CI
```
EzraBrooks, line 1: Changed this name while we're gonna be messing with names anyway - since this workflow just orchestrates "CI" in general.

EzraBrooks:
we can do some more cosmetic tweaking to the workflow names and job names once we run and see how they look in practice. they might be a bit verbose right now

.github/workflows/ros_build_and_test.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Sanitize tag name
+        id: sanitize_tag
+        run: |
+          BRANCH="${{ github.head_ref || github.ref_name }}"
+          TAG="${BRANCH//\//-}"
+          echo "tag=$TAG" >> $GITHUB_OUTPUT
```
EzraBrooks, line 36: In a follow-up PR (since I don't want to risk functional changes in this PR that's mostly cosmetic), I think we should move this up to the root workflow and pass the tag name as an argument to the two jobs that use it.

.github/workflows/build_and_test.yml
```diff
@@ -22,178 +22,20 @@ concurrency:
...
+    uses: ./.github/workflows/format.yml
 
-  ubuntu_build_and_test:
-    timeout-minutes: 30
-    strategy:
-      fail-fast: false
-      matrix:
-        ubuntu_version: [22.04, 24.04, 26.04]
+  build_pypi_wheels:
+    uses: ./.github/workflows/build-pypi-wheels.yml
```
EzraBrooks, line 31:
```suggestion
  build_pypi_wheels:
    name: Wheels
    uses: ./.github/workflows/build-pypi-wheels.yml
```

.github/workflows/build_and_test.yml
```diff
@@ -22,178 +22,20 @@ concurrency:
...
-          load: true
-          cache-from: type=gha,scope=ubuntu-${{ matrix.ubuntu_version }}
-          cache-to: type=gha,mode=max,scope=ubuntu-${{ matrix.ubuntu_version }}
-      - name: Run tests
-        run: |
-          docker run --rm \
-            roboplan_ubuntu:${{ steps.sanitize_tag.outputs.tag }} \
-            /bin/bash -c './scripts/run_tests.bash'
+  ubuntu_build_and_test:
+    uses: ./.github/workflows/ubuntu_build_and_test.yml
```
EzraBrooks, line 35:
```suggestion
  ubuntu_build_and_test:
    name: Bare Ubuntu
    uses: ./.github/workflows/ubuntu_build_and_test.yml
```

.github/workflows/build_and_test.yml
```diff
@@ -22,178 +22,20 @@ concurrency:
...
-          push: false
-          load: true
-          cache-from: type=gha,scope=ros-${{ matrix.ros_distro }}
-          cache-to: type=gha,mode=max,scope=ros-${{ matrix.ros_distro }}
-      - name: Run tests
-        run: |
-          docker run --rm \
-            roboplan_ros:${{ steps.sanitize_tag.outputs.tag }} \
-            /bin/bash -c './src/roboplan/scripts/run_tests.bash'
+    uses: ./.github/workflows/ros_build_and_test.yml
```
EzraBrooks, line 39:
```suggestion
    name: ROS
    uses: ./.github/workflows/ros_build_and_test.yml
```

.github/workflows/build_and_test.yml
```diff
@@ -22,178 +22,20 @@ concurrency:
...
-      - name: Check bindings stub files
-        run: |
-          if ! git diff --exit-code --shortstat -- '*.pyi'; then
-            echo "::error::The .pyi stub files are out of sync with the source code. Make sure to rebuild the Python bindings without symlink builds using 'pixi run -e ci install' and commit the changes."
-            echo "Git Diff:"
-            git diff -- '*.pyi'
-            exit 1
-          fi
-          echo "✅ .pyi stub files look good!"
+    uses: ./.github/workflows/pixi_build_and_test.yml
```
EzraBrooks, line 43:
```suggestion
    name: Pixi
    uses: ./.github/workflows/pixi_build_and_test.yml
```

.github/workflows/build_and_test.yml
```diff
@@ -22,178 +22,20 @@ concurrency:
...
-        with:
-          cache: true
-          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
-          frozen: true
-          manifest-path: pyproject.toml
-      - name: Build
-        run: pixi run -e default install
-      - name: Run RRT benchmarks
-        run: pixi run -e default test_benchmarks
+    uses: ./.github/workflows/rrt_benchmarks.yml
```
EzraBrooks, line 47:
```suggestion
    name: Benchmarks
    uses: ./.github/workflows/rrt_benchmarks.yml
```

EzraBrooks (commented):
Thoughts on these proposed names for the folderization of the workflow?

EzraBrooks:
I got a bit lost in the sauce on changing the names of things but I think it looks really nice now.

tests/unified_python_package_test.py
```diff
@@ -263,14 +263,13 @@ def test_dependent_package_tests_keep_upstream_example_model_dependency() -> Non
...
     workflow = _read(".github/workflows/build-pypi-wheels.yml")
 
-    assert "pull_request:" in workflow
-    assert "push:" in workflow
-    assert "branches:" in workflow
-    assert "- main" in workflow
     assert "workflow_call:" in workflow
+    assert "pull_request:" not in workflow
+    assert "push:" not in workflow
+    assert "workflow_dispatch:" not in workflow
```
eholum: Removed, right? This test now fails for me locally,

```
>       assert "workflow_dispatch:" not in workflow
E       AssertionError: assert 'workflow_dispatch:' not in 'name: Build...und: error\n'
E         
E         'workflow_dispatch:' is contained here:
E           name: Build PyPI wheels
E           
E           on:
E             workflow_dispatch:
E             workflow_call: {}...
E         
E         ...Full output truncated (133 lines hidden), use '-vv' to show
```

That said does this get run by anything? 

```suggestion
```

eholum (commented):
I like this change! Def easier to see the results. Minor nit in what looks like a bypassed test, but otherwise even the required status checks panel is nicer to look at:

Image

tests/unified_python_package_test.py
EzraBrooks: failing locally but not in CI? suspicious..

EzraBrooks: It appears this test has never been invoked by CI. It seems like it's mostly just AI ramblings. Should I just delete it?

eholum: A non-functional, non-executed test definitely doesn't add value... Seems like it was added in the initial wheels setup, https://github.com/open-planning/roboplan/pull/247.

So since we haven't been using it I would vote delete. But if someone who knows more about python packages (namely, everyone) than I do might see the value in actually wiring it to something...

EzraBrooks: I actually have an incoming draft that unifies the wheel build into the new superbuild. Also, the CI job for building the wheels already sanity-checks them by importing them

eholum (approved):
LGTM, can coordinate with @sea-bass about merging and updating the status checks...

sea-bass (approved):
Updated naming LGTM, gonna fiddle with the required checks now

PR #300, opened 2026-09-03
Move file I/O out of Scene constructor, support loading MJCF

### LINKED ISSUE #244: Discussion: Do we stay with SRDF, or provide an alternative?
SRDF is good for a few reasons:

1. You can sometimes grab existing repositories with an SRDF ready to go
2. Pinocchio has a convenient function to remove collision pairs from SRDF
3. People can still use MoveIt Setup Assistant to generate SRDFs

However, we don't _need_ to keep using SRDF if we don't want. Should we consider alternatives?

### LINKED ISSUE #294: Allow constructing a Scene with an MJCF and no SRDF
To make it easier for people who are developing robot behavior in pure MuJoCo simulations, RoboPlan should allow creating a Scene from just an MJCF and not require an accompanying SRDF.

EzraBrooks:
Closes #294, closes #244

Instead of adding file extension detection logic to the string-based constructor for `Scene`, which is decidedly against the \*nix philosophy (where file extensions "aren't supposed to matter"), I propose we break the API and provide structs that represent a URDF robot description (which can optionally include an accompanying SRDF) and an MJCF robot description (which at this time requires only the MJCF).

Opening as draft so you can :eyes: at it and decide whether you agree on this API contract @sea-bass @sjahr


roboplan_core/include/roboplan/core/scene.hpp
```diff
@@ -35,6 +36,23 @@ namespace roboplan {
...
+/// @brief Path to an MJCF robot description.
+struct MjcfSceneDescription {
+  std::filesystem::path mjcf_path;
+};
+
+/// @brief Extended joint limits parsed from a URDF <limit> tag.
+struct UrdfExtendedJointLimits {
+  std::optional<double> acceleration;
+  std::optional<double> jerk;
+};
```
EzraBrooks, line 58: TODO: restore dropped comment from the file move on this struct

roboplan/include/roboplan/core/scene.hpp
```diff
@@ -35,6 +36,23 @@ namespace roboplan {
 /// @param name path The path to the file.
 std::string readFile(const std::filesystem::path& path);
 
+/// @brief Paths to a URDF robot description and its SRDF planning configuration.
+struct UrdfSceneDescription {
+  std::filesystem::path urdf_path;
+  std::filesystem::path srdf_path;
```
sea-bass: should we make the SRDF optional? like, if not available we just don't call the SRDF part (i.e., `pinocchio::srdf::removeCollisionPairsFromXML`)?

It starts to give us an escape hatch in the event we don't want to use SRDF even when coupled with URDF, but rather set the collision pairs through another mechanism.

roboplan_core/include/roboplan/core/scene.hpp
sea-bass: So I believe since we added this, the URDF 1.2 transition (which officially adds acceleration and jerk to the spec) already happened and may be more widely available, including in Pinocchio 4.1:

* https://github.com/ros/urdfdom#version-12
* https://github.com/stack-of-tasks/pinocchio/pull/2909
* https://github.com/stack-of-tasks/pinocchio/releases/tag/v4.1.0

Wonder if it's still too early to remove these extended joint limits altogether?

sea-bass (commented):
Changes are pretty clean!

I would recommend sitting on this draft PR for a bit and seeing how this ends up working (or not working) for something like a Franka FR3 MJCF description, in case any of this needs to be switched up.

Besides that, most of my comments are related to "oh since we're doing this, can we also clean up X?" But most of these could be done later as well.

EzraBrooks:
Yeah I figured I'd put an example in before I put this up for proper review

roboplan/include/roboplan/core/scene.hpp
EzraBrooks: That seems like a good idea. I like having the SRDF be a member of a URDF scene description since (afaik) you're almost certainly only using the former if you're using the latter, but it definitely should be possible to use without an SRDF at all, so we can make it an optional.

if we put together a fluent interface, I could see invoking Scene setup in the future looking something like:

```c++
auto scene = Scene(...)
  .allow_collisions("link1", "link2")
  .allow_collisions("link2", "link3")
  .allow_collisions("link3", "link4")
```

EzraBrooks:
talked out-of-thread with @eholum and @sea-bass about how the changes here were all pointing in the direction of a refactor to remove file I/O from the constructor and instead provide better-qualified constructor signatures.

Added a free function to load a URDF+SRDF pair into a UrdfSceneDescription, then made the Scene constructor accept that to be able to parse the special attributes in there such as the "extended URDF joint limits".

Added a free function to load an MJCF directly into a PinocchioSceneDescription made up of a Model and a GeometryModel, then made the scene constructor accept that.

roboplan_core/include/roboplan/core/scene.hpp
EzraBrooks: seems like since Pinocchio is stuck at 4.0 on Humble rn, we still have to process this ourselves for the moment

roboplan_core/include/roboplan/core/scene.hpp
```diff
@@ -31,9 +32,30 @@ namespace coal = hpp::fcl;
...
+
+/// @brief Loads an MJCF model and its collision geometry from disk.
+PinocchioSceneDescription loadMjcfModel(const std::filesystem::path& mjcf_path);
+
+/// @brief Extended joint limits parsed from a URDF <limit> tag.
+struct UrdfExtendedJointLimits {
+  std::optional<double> acceleration;
+  std::optional<double> jerk;
+};
```
sea-bass: does it make sense to move this stuff to another file as we start accumulating robot description parsing classes / functions?

roboplan_examples/python/example_ik.py
```diff
@@ -44,16 +49,14 @@ def main(
...
         sys.exit(1)
 
-    # Pre-process with xacro. This is not necessary for raw URDFs.
+    package_paths = [get_package_share_dir()]
+
     urdf_xml = xacro.process_file(model_data.urdf_path).toxml()
     srdf_xml = xacro.process_file(model_data.srdf_path).toxml()
-    package_paths = [get_package_share_dir()]
 
-    # Specify argument names to distinguish overloaded Scene constructors from python.
```
sea-bass, line 52: thanks for ejecting these random low-nutrition comments

roboplan_examples/python/example_action_chunk_tracking.py
```diff
@@ -350,13 +350,13 @@ def main(
...
         model_pin,
-        urdf_xml,
+        str(model_data.urdf_path),
         pin.GeometryType.COLLISION,
         package_dirs=package_paths,
     )
     visual_model = pin.buildGeomFromUrdfString(
         model_pin,
-        urdf_xml,
+        str(model_data.urdf_path),
```
sea-bass: are these correct? These shouldn't be the path, but rather the XML file _contents_, right?

I mean, maybe there are other pinocchio functions that will just accept the path, but without running the examples myself, I suspect they are currently not working.

Same in other examples.

sea-bass (commented):
I was just snooping around and had a few comments

roboplan_examples/python/example_action_chunk_tracking.py
EzraBrooks: I haven't run the examples locally on this branch so they probably are borked.

btw we should take a page out of roslibjs's book and unit test our examples :)

EzraBrooks:
An obvious follow-up to this PR is going to be adding a helper function to propagate allowed collisions between adjacent links, much the same way MSA does when generating an SRDF; but at runtime.

EzraBrooks:
RoboPlan RRT in MuJoCo with no URDF/SRDF pair:


https://github.com/user-attachments/assets/27e43667-87a0-4d84-9b16-9264324b8103

EzraBrooks:
I found a bug in the RRT planner where you can't use it with a planning group that has an empty name `""` because it conflicts with the usage of `""` as a stand-in value for "empty" in the class definition. I fixed it here, but don't like the fix very much so I'm going to extract and make it better

#338

benchmarks/test_benchmark_rrt.py
```diff
@@ -68,8 +68,7 @@ def create_scene(model_name: str) -> Scene:
 
     scene = Scene(
         f"{model_name}_benchmark_scene",
-        urdf=urdf_xml,
-        srdf=srdf_xml,
+        UrdfSceneDescription(urdf_xml, srdf_xml),
```
EzraBrooks, line 72: These two sequential args are of the same type `str`, which makes them easy to mix up. They should be required named parameters instead, like the initializer syntax in C++.

EzraBrooks:
I'm not going to mark this as addressing #190 because the robot is just following some random path, not performing a task. However, a follow-up PR could very well modify the example to address that issue. My next step is going to be #337 and modifying the example to use that.

roboplan_core/bindings/src/core.cpp
```diff
@@ -22,6 +22,39 @@ namespace roboplan {
...
+                                std::string(required_name) + "' and optionally '" +
+                                std::string(optional_name) + "'")
+                                   .c_str());
+  }
+  std::optional<T> optional_value;
+  if (has_optional && !kwargs[optional_name].is_none()) {
+    optional_value = nanobind::cast<T>(kwargs[optional_name]);
+  }
+  return {nanobind::cast<T>(kwargs[required_name]), optional_value};
+}
```
sea-bass: oof, this is quite something. I thought we could do keyword args just by using `_a`?

But is the difference here that these are _only_ allowed as keyword and not as positional?

Is it worth incurring so much extra boilerplate just for this?

roboplan_core/include/roboplan/core/scene.hpp
```diff
@@ -31,9 +32,37 @@ namespace coal = hpp::fcl;
...
+
+/// @brief Filesystem paths to a URDF robot description and optional SRDF planning configuration.
+struct UrdfScenePaths {
+  std::filesystem::path urdf_path;
+  std::optional<std::filesystem::path> srdf_path;
+};
+
+/// @brief Loads a URDF robot description and optional SRDF planning configuration from disk.
+UrdfSceneDescription loadUrdfSceneDescription(const UrdfScenePaths& paths);
```
sea-bass: there's a lot of new structs here, I feel every time I look again there are more and I'm starting to wonder if maybe we can cut down a little.

Could we just accept the urdf and optional srdf directly into this function rather than a struct?

roboplan_core/include/roboplan/core/scene.hpp
```diff
@@ -31,9 +32,37 @@ namespace coal = hpp::fcl;
...
+
+/// @brief Pinocchio model and collision geometry used to construct a Scene.
+struct PinocchioSceneDescription {
+  pinocchio::Model model;
+  pinocchio::GeometryModel collision_model;
+};
+
+/// @brief Loads an MJCF model and its collision geometry from disk.
+PinocchioSceneDescription loadMjcfModel(const std::filesystem::path& mjcf_path);
```
sea-bass, line 53: in the URDF, but no SRDF case, should we also have a `loadUrdfModel` that only takes a `urdf_path`?

If we do this, and we eventually find an alternative to SRDF, then we could remove that whole code path while maintaining this one?

roboplan_examples/python/example_rrt_mujoco.py
```diff
@@ -0,0 +1,318 @@
...
+"""
+Point-to-point RRT planning on a scene built from a MuJoCo (MJCF) model, executed in MuJoCo.
+
+Unlike the other planning examples, which build a :class:`Scene` from a URDF/SRDF pair, this one
+constructs the scene directly from an MJCF file using :func:`loadMjcfModel`. The MJCF is fetched
+via ``robot_descriptions`` (the mujoco_menagerie collection), so no local model files are needed.
+
+A ground plane is added to both the planning scene and the MuJoCo model, so the robot plans and
+executes above a real floor. This uses a redundant (7-DOF) arm such as the Franka Panda, whose
+extra degree of freedom lets it comfortably route around the floor to reach a variety of poses.
```
sea-bass: can prob remove this paragraph tbh

roboplan_examples/pyproject.toml
```diff
@@ -19,10 +19,12 @@ roboplan-toppra = { path = "../roboplan_toppra" }
 
 [tool.pixi.package.run-dependencies]
 matplotlib = ">=3.10.5,<4"
+mujoco = ">=3.2,<4"
 numpy = "*"
 plyfile = ">=1.1,<2"
 pycollada = ">=0.9.3,<0.10"
 pynput = ">=1.8.2,<2"
+robot_descriptions = ">=1.12,<2"
```
sea-bass, line 27: we'll have to install these in the Dockerfiles for pixi-less ubuntu and ROS as well then, yeah?

sea-bass (commented):
Gave it a more thorough look. I guess I have 2 moderate concerns

1. The number of structs is sort of ballooning, and furthermore the URDF and MJCF scene construction paths don't converge at all. I left more detailed inline comments, but wondering if we should at least have a URDF-only (no SRDF) path that looks identical to the MJCF? Then, the URDF+SRDF one is the one special-case that we can strive to remove as we find we can.
2. Is it truly worthwhile to go through the nanobind gymnastics you had to, just to get a few keyword-only args? What if we make them positional or keyword and call it a day?

roboplan_core/include/roboplan/core/scene.hpp
EzraBrooks: I can revert this. I was just trying to avoid consecutive positional arguments of the same type, since they're a really common footgun for users.

EzraBrooks:
> furthermore the URDF and MJCF scene construction paths don't converge at all.

I don't follow. The previous constructor logic was mostly extracted to an `initialize` function verbatim and is called by both code paths

roboplan_examples/pyproject.toml
EzraBrooks: We ship the examples in the Docker containers? That seems wasteful.

roboplan_core/bindings/src/core.cpp
EzraBrooks: I dropped this. Sucks that we can't have required positional args in the ancient version of nanobind on Apt on Jazzy and Kilted.

roboplan_examples/pyproject.toml
sea-bass: Yeah, they are used as dev containers for users in addition to CI. Plus, if we act on the suggestion of unit testing examples, this will be useful / necessary.

roboplan_core/include/roboplan/core/scene.hpp
EzraBrooks: dropped.

roboplan_core/include/roboplan/core/scene.hpp
EzraBrooks: Frankly, I don't think we should be using file paths here at all. Users should load the file themselves and hand us the string content. The only reason this function is like this is because, as we discussed, Pinocchio for some reason only has a "load from disk" signature for MJCF.

If we want a consistent API, we can hack around this by taking MJCF XML and creating a temporary file and handing that to Pinocchio (ew)

roboplan_examples/python/example_rrt_mujoco.py
EzraBrooks: done

sea-bass:
> > furthermore the URDF and MJCF scene construction paths don't converge at all.
> 
> I don't follow. The previous constructor logic was mostly extracted to an `initialize` function verbatim and is called by both code paths

Yeah, I think I've been thinking myself in circles about whether there's a good way for them to converge one level above `initialize()`, i.e., at the `PinocchioSceneDescription` level. But with SRDF in the mix, I think I'm with you.

So what I was asking is should there instead be a path from URDF only (no SRDF) -> `PinocchioSceneDescription` -> the same constructor, and just have the SRDF collision pair removal / group creation happen after the fact?

It is convenient to have a single URDF+SRDF scene creation path, though...

roboplan_core/bindings/src/core.cpp
sea-bass: agreed... but we already dropped Humble support for non-source builds before its EOL, so if this continues to be annoying we could consider doing the same for Jazzy (or telling people to use robostack where it will work lol)

EzraBrooks:
like just construct the model from the URDF and have a function that's `.importPlanningConfigurationFromSrdf` or something? That is a fair division of responsibilities.. the constructor only knowing about the actual robot geometry and leaving the definition of groups etc. to setup functions.

I feel like that lives in a separate PR (which, if you want to ship this before ROSCon, I'm happy to turn around quickly as I can) since it's in the same vein as #337.

roboplan_core/include/roboplan/core/scene.hpp
sea-bass: ah yes, thanks for the reminder. Let me raise an issue in Pinocchio real quick and maybe I'll tackle that at some point.

roboplan_examples/pyproject.toml
EzraBrooks: ok, I have Feelings about this, but I'll do it and to avoid going on another side quest I'm gonna loosely pin those in the Dockerfiles.. I don't want to hold this up on yet more devops stuff

roboplan_core/include/roboplan/core/scene.hpp
sea-bass: wait is this a known bug? 

https://github.com/stack-of-tasks/pinocchio/blob/2b7544df404c2eeafcb0952c78154f8c8bbbfabc/include/pinocchio/src/parsers/mjcf/mjcf.hxx#L33-L59

there are some comments about it being "badly named and having the same behavior" as the file path one"

sea-bass: https://github.com/stack-of-tasks/pinocchio/issues/2951

EzraBrooks:
^ that would also allow using an MJCF-SRDF pair, which is kind of cursed but maybe someone wants that for some reason

sea-bass (approved):
OK I think this looks good -- thanks for indulging my crashout about reducing complexity.

EzraBrooks:
It's ok, I am doing my best to refrain from crashing out about the closest thing to named parameters in C++ being struct initializer lists. which was why I kept adding more structs.

But I think the route of moving the SRDF out of the `loadUrdfSceneDescription` signature fixes the consecutive-string-argument footgun as well. `Scene(loadUrdfSceneDescription(urdf)).importSrdf(srdf)` is a lot harder to typo

PR #301, opened 2026-09-03
Use pixi-build to declare Conda dependencies for each package

EzraBrooks:
This PR:

- enables `pixi-build` for the root workspace
- adds `pixi-build-cmake` to all the packages
- adds `pixi-build-cmake` and a dependency list to `roboplan_examples`
- removes all direct dependencies of individual packages from the root pixi manifest and moves them to the individual children, making it possible to build & publish them as individual conda packages

We build Conda packages through the existing cmake, similar to how it's currently done in the CF feedstock.

We now know that `pixi publish` cannot ship directly to CF, but it sounds like the Prefix guys are working on a way to have `pixi publish` eject the generated recipe for use in CF feedstocks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [tool.pixi.workspace]
 channels  = ["conda-forge"]
+preview = ["pixi-build"]
 platforms = ["linux-64", "linux-aarch64", "osx-arm64", "win-64"]
 
 [tool.pixi.activation.env]
@@ -37,6 +38,11 @@ ROBOPLAN_C_COMPILER  = "clang-cl"
 ROBOPLAN_CXX_COMPILER = "clang-cl"
 ROBOPLAN_LINKER_FLAG = " "
 ROBOPLAN_EXTRA_CMAKE_ARGS = "-DCMAKE_LINKER=lld-link"
+# The Windows Clang toolchain activates the MSVC build environment, which
+# selects platform and toolset values for Visual Studio generators. Ninja
+# accepts neither setting.
+CMAKE_GENERATOR_PLATFORM = ""
+CMAKE_GENERATOR_TOOLSET = ""
 
 # Symlink creation in Windows requires developer mode or admin privileges.
 [tool.pixi.feature.local.target.win.activation.env]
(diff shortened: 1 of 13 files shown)


sea-bass:
Interesting! Any idea how this would play with the existing mechanism:

https://github.com/conda-forge/roboplan-feedstock

EzraBrooks:
Well initially it shouldn't conflict. Later I think we could probably just have the feedstock run pixi publish?

EzraBrooks:
letting Pixi pull in the Conda Windows toolchain is causing some funky issues 🤔

EzraBrooks:
depends on the Windows fix in #302

EzraBrooks:
If others agree with #304, I actually think that one should probably go in first so this can be re-tooled around direct usage of `cmeel` and push more of the `pyproject.toml` definitions _out_ of `[tool.pixi]` and into directly wheel-compatible config. We'll still have some `[tool.pixi]` stuff for the dev toolchain and for Conda packaging but it's best if we can push as much definition into PyPI-land as possible since Pixi will respect both and we won't have to duplicate configuration.

sea-bass:
> Well initially it shouldn't conflict. Later I think we could probably just have the feedstock run pixi publish?

I think for this one, I'd like to try build the feedstock repo locally while pointing to this branch, to see if this is compatible. I will find some time to test this out.

I also generally need to read up more on what pixi build actually does so I can understand better.

EzraBrooks:
makes sense to me. `pixi publish` should generate `.conda` files for every package in the repo automagically, no need for hand-jamming recipe.yaml etc.

sea-bass:
Yeah, I am also looking at the fact that the current feedstock is building packages both with and without bindings, and making sure this path is still possible here (it should be...)

sea-bass:
Well good news, it doesn't break the traditional feedstock workflow, which means we're in the clear for this change!

https://github.com/conda-forge/roboplan-feedstock/pull/22

(you'll see some failures at the end just because the admin team needs to approve the new output names, but otherwise looks good)

EzraBrooks:
So this has huge conflicts now that I went around fixing up the other underlying builds (i.e. making `cmeel` actually work for every individual package). There are two routes forward if we do want this:

- Big ol' rebase and continue doing what I had been doing, which is use `pixi-build-cmake` to invoke the CMake, which unfortunately still means duplicating dependency lists between PyPI and Conda
- Create a conda-forge feedstock (or convince someone else to do so) for `cmeel` so we can use `pixi-build-python` pointed directly at the now-functional `pyproject.toml` builds, which would allow us to have one dependency list for both PyPI and Conda.

I would vastly prefer the latter, as without it we have three dependency lists: PyPI, Conda, and ROS. Two sources of truth is already too many!

Currently we have four, though: PyPI, Conda via Pixi, Conda in the recipe.yaml in the feedstock, and ROS. So we should do one of these things to yeet the recipe.yaml.

sea-bass:
I'd also favor the latter. My suggestion would be to post an issue on https://github.com/cmake-wheel/cmeel explaining the use case. Would you be willing since you're more familiar?

EzraBrooks:
https://github.com/cmake-wheel/cmeel/issues/307

EzraBrooks:
The remaining changes are digestible into two bits;

- conda packaging of the existing build workflow
- migrating the Python install logic from the `superbuild` folder to the `roboplan_common` shared utils so that packages can use it without being part of the superbuild, further de-specializing the superbuild as a build type

I'll roll the latter out of this PR and into a separate one

EzraBrooks:
mmm, my second time finding a bug that causes lock file instability in Pixi.. I am on the bleeding edge as usual

pyproject.toml
```diff
@@ -197,25 +193,17 @@ ci = { features = ["ci", "toolchain"] }
...
-rolling = { features = ["ros", "toolchain", "rolling"], no-default-feature = true }
-lyrical = { features = ["ros", "toolchain", "lyrical"], no-default-feature = true }
-kilted = { features = ["ros", "toolchain", "kilted"], no-default-feature = true }
-jazzy = { features = ["ros", "toolchain", "jazzy"], no-default-feature = true }
-humble = { features = ["ros", "toolchain", "humble"], no-default-feature = true }
+rolling = { features = ["ros", "toolchain", "rolling"]}
+lyrical = { features = ["ros", "toolchain", "lyrical"]}
+kilted = { features = ["ros", "toolchain", "kilted"]}
+jazzy = { features = ["ros", "toolchain", "jazzy"]}
+humble = { features = ["ros", "toolchain", "humble"]}
```
EzraBrooks, line 202: Removing `no-default-feature` here, moving all the dependencies into the children, and loosening some version pins so that the ROS distros could manage them themselves seems to have fixed the cross-toolchain version compatibility issues.

pyproject.toml
```diff
@@ -310,8 +293,6 @@ channels = ["https://prefix.dev/robostack-humble"]
 
 [tool.pixi.feature.humble.dependencies]
 ros-humble-ros-core = "*"
-# Humble's boost pin limits Pinocchio to 4.0.0, unlike the newer distros.
-pinocchio = ">=4.0.0,<5"
```
EzraBrooks, line 314: I pulled the whole codebase down to Pinocchio 4.0 so we aren't matrixing support here.

sea-bass: I don't think that's ideal... everywhere else we are already using 4.1

I'd probably prefer the matrixing. Or at least rely on robostack's version for that distro, which already takes care of a similar issue?

EzraBrooks: > everywhere else we are already using 4.1

.. so we were shipping a Humble binary against 4.0 that depended upon 4.1?

sea-bass: No no, I think the humble binaries couldnt ship on rosdistro because nanobind is not available on the buildfarm as an apt dependency haha

sea-bass: And on RoboStack Humble, pinocchio is similarly locked to 4.0.0 because of boost/eigen/whatever else

sea-bass: My point was I'd rather special case humble than pull everything else back

EzraBrooks: are we depending upon 4.1 features? in another thread you had questioned whether it was ok for us to adopt 4.1 features such as the URDF joint accel limits

EzraBrooks: given that tests are passing it seems like we probably aren't actually using 4.1 features

sea-bass: We are not depending on 4.1 features at the moment, nope

sea-bass: Wait actually the one thing in 4.1 we are relying on is a bugfix to their viser visualizer for scaled meshes. So it won't break tests, but might impact users

EzraBrooks: ok, I think we need to make a philosophical decision here then (cc @eholum). If a baseline dependency (Pinocchio 4.1) is not available in a ROS distro, what do we do?

I would argue "drop support for that ROS distro, at least temporarily until the fix is available" because the way we were previously accommodating this was by just silently ignoring the issue in Humble.

If we're not shipping binaries for Humble anyway, should we just drop it - at least for now?

sea-bass: I just came back to this since I was distracted yesterday and realized that... these are just lower pins in the pixi manifests, but the lockfiles for each distro have no reason to pick 4.1 except for Humble, right? Indeed, I looked at the lockfile and that is the case.

EDIT: hah nice timing

roboplan_examples/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[tool.pixi.package]
+name = "roboplan-examples"
+version = "0.6.1"
+publish = false
```
sea-bass: all the other packages just have `publish = true` under here, but nothing else. Why is this one different?

EzraBrooks: Because the other ones already had a `[project]` table; I'll push these other fields up to `[project]`

pyproject.toml
sea-bass: The other tidbit of information is that the reason RoboStack Humble hasn't been able to upgrade Pinocchio past 4.0.0 is that there are some Gazebo (I believe) packages that are failing under Eigen 5 and/or Boost 1.90 (I forget which), but the maintainer team is working on that. So taking ROS aside, RoboPlan itself has no issues between 4.0 and 4.1 minus the Python visualizer bug.

So I've convinced myself this is okay as is.

sea-bass (approved):
I think we're good to go here

pyproject.toml
EzraBrooks: great. I think this is probably the Right Way to manage this; just describe a compatible version range in the actual package and let the distribution decide which (highest) version pin within that range to select.

PR #302, opened 2026-09-04
Constrain scope of DLL loader manipulation

EzraBrooks:
I found while working on #301 that the existing DLL loader code can cause crashes if the `PATH` contains long directory names, thanks to the wonders of Windows's old filesystem APIs.

I figured just to be safe and responsible citizens on Windows machines, we should fix this problem by only registering DLLs from `AMENT_PREFIX_PATH` and `CONDA_PREFIX`. Technically the crash can still happen if your workspace path is really long, but this makes it less likely and is also good for separation of concerns.

Note that the particular `PATH` entry that was crashing it was a common one on dev machines: a really long Visual Studio utility path that was like six layers deep under Program Files.

### DIFF
--- a/packaging/python/CMakeLists.txt
+++ b/packaging/python/CMakeLists.txt
@@ -16,6 +16,7 @@ endif()
 
 roboplan_register_build_tree_packages()
 
+add_subdirectory("${ROBOPLAN_REPOSITORY_ROOT}/roboplan_common" "roboplan_common")
 add_subdirectory("${ROBOPLAN_REPOSITORY_ROOT}/roboplan_example_models" "roboplan_example_models")
 add_subdirectory("${ROBOPLAN_REPOSITORY_ROOT}/roboplan" "roboplan")
 add_subdirectory("${ROBOPLAN_REPOSITORY_ROOT}/roboplan_simple_ik" "roboplan_simple_ik")
--- a/roboplan/bindings/CMakeLists.txt
+++ b/roboplan/bindings/CMakeLists.txt
@@ -169,6 +169,7 @@ if(GENERATE_PYTHON_STUBS AND COMMAND nanobind_add_stub)
   # own stub generation.
   set_target_properties(_core_ext PROPERTIES
     LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/pystage/roboplan/core")
+
   file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/python/roboplan/core/__init__.py"
     DESTINATION "${CMAKE_BINARY_DIR}/pystage/roboplan/core")
   set_target_properties(_filters_ext PROPERTIES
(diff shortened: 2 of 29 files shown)


sea-bass (commented):
This is good, but now that these snippets are growing, is there a good way to not have to copy-paste this in all the binding packages?

Pinocchio does something similar with CMake macros, which I'd like to avoid, but curious if you have an idea here.

EzraBrooks:
yeah I share the concern.. I'll look a little deeper at that

sea-bass:
BTW this is how they do it, I guess it wasn't quite macros and actually seems pretty reasonable...

They have a `windows_dll_manager.py` file installed [here](https://github.com/stack-of-tasks/pinocchio/blob/e7792444951a191c13e81aede4608c34d3012440/bindings/python/CMakeLists.txt#L370) from [here](https://github.com/stack-of-tasks/pinocchio/blob/e7792444951a191c13e81aede4608c34d3012440/bindings/python/pinocchio/windows_dll_manager.py), and being imported [here](https://github.com/stack-of-tasks/pinocchio/blob/e7792444951a191c13e81aede4608c34d3012440/bindings/python/pinocchio/__init__.py#L25).

The one caveat is that they only really have one package. So we have to figure out where to put this file.

sea-bass:
I don't know if this will cause problems with some build systems (probably yes), but actually `roboplan_example_models` is the bottom-most package, in that it does not depend on `roboplan_core`.

In fact, `roboplan_core` depends on `roboplan_example_models` for unit testing,

EzraBrooks:
I guess that means we need a new bottom-most package. Though, to be fair, that's probably where the cmake macros should have been living all this time :)

sea-bass:
blaaaghhh, you're not wrong.

`roboplan_common` is the first thing that comes to mind (funny enough both moveit and nav2 have this)? I can throw it into https://github.com/conda-forge/admin-requests/pull/2327 as a conda output request

EzraBrooks:
Honestly this might actually be a good place to use a CMake template and generate a small bit of Python.

EzraBrooks:
actually that would still result in copypasta, just someplace else. you're probably right that we need a "common" package, like everyone who came before us 😆

sea-bass:
"utils"

sea-bass:
Yeah, common package is fine. @eholum have any inclinations on naming?

EzraBrooks:
https://www.youtube.com/watch?v=5W4NFcamRhM

EzraBrooks:
\> want to fix Windows-specific dynamic loading bug
\> add a whole ROS package to the workspace

sea-bass:
We can just land the original copy pasta solution for now and do the common package after as well! That way it doesnt need to hold up other things

EzraBrooks:
no, you know me.. maximum effort

sea-bass (approved):
Excellent; thank you!

Reserving a `roboplan-common` output here: https://github.com/conda-forge/admin-requests/pull/2327

PR #303, opened 2026-09-05
Unify superbuild with wheel build

EzraBrooks:
the `scikit-build` wheel build is its own superbuild separate from the one used by developers of the repo. this PR adds the wheel-specific CMake code to the developer superbuild behind a flag to consolidate the logic.


EzraBrooks:
guh. apt caching breakages in CI. fixing in #305

sea-bass (commented):
There's a merge conflict from some of your other PRs, so let's fix that first.

Everything here looks good -- love the consolidation. One question is that now that we don't have 2 separate directories, can `superbuild` be renamed to plain `roboplan` and used as the metapackage itself? Or do you think those would need to be separate entities?

EzraBrooks:
The tricky bit is that the superbuild and the metapackage aren't the same thing. This is why in the ROS ecosystem you have a metapackage definition that is separate from the auto-generated superbuild colcon gives you. I don't really like the directory name `superbuild` either but I just picked something to avoid breaking colcon package discovery.

sea-bass (approved):

PR #304, opened 2026-09-05
Drop `scikit-build` in favor of direct usage of `cmeel`

EzraBrooks:
Depends on #303.

Uses `cmeel`'s built-in relocatable-binary logic instead of using `scikit-build` and then hand-rolling it (with some help of `cmeel` for some reason?)

I won't claim that my vibe-coded wheel implementation is any more virtuous than the previous vibe-coded wheel implementation.. but this seems closer to being "all our packaging jobs are just thin CMake wrappers" than before.


EzraBrooks:
lol, why was the wheel build's CMake minimum version set to 3.5??? that crashes the Windows toolchain now that we're actually respecting it. Will bump to 3.22 which is what we document literally everywhere else

EzraBrooks:
missing new `roboplan_common` from #302, will fix

README.md
```diff
@@ -23,7 +23,7 @@ The main folders found in this repo are as follows.
 - `roboplan_cartesian_planning` : A Cartesian planner for following task-space paths.
 - `roboplan_example_models` : Contains robot models used for testing and examples.
 - `roboplan_examples` : Basic examples with real robot models.
-- `packaging` : Files to help package wheels for release on PyPi.
+- `superbuild` : Shared CMake entry point for local, Conda, and PyPI-wheel builds.
```
sea-bass, line 27: one minor thing I missed last time around -- can you add `roboplan_common` to here too?

sea-bass (approved):

README.md
EzraBrooks: done

PR #305, opened 2026-09-05
Fix unstable Apt caching in ROS Docker image

EzraBrooks:
@sea-bass @eholum I think this is what blows up the ROS Docker builds.

- the Apt cache getting stuck in Docker layers because it was neither cache-mounted nor `rm -rf`'d at the end of each Apt step
- the `rosdep install` step not running `apt-get update` to refresh its apt cache.

For good measure I also shifted the pip cache out of the Docker layers into a cache mount.

### DIFF
--- a/.docker/ros/Dockerfile
+++ b/.docker/ros/Dockerfile
@@ -8,9 +8,11 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
 SHELL ["/bin/bash", "-c"]
 
 # Install core dependencies.
-RUN apt update -y \
-    && apt upgrade -y \
-    && apt install -y python3-pip
+RUN --mount=type=cache,target=/var/lib/apt/lists,sharing=locked \
+    --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    apt-get update -y \
+    && apt-get upgrade -y \
+    && apt-get install -y python3-pip
 
 # Install Python dependencies needed for bindings.
 # Note that since Pinocchio in ROS is built with Numpy 1.x, we must preserve this by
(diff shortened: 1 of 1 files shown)


.docker/ros/Dockerfile
```diff
@@ -8,25 +8,31 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
...
 RUN mkdir -p /workspace/roboplan_ws/src/roboplan
 WORKDIR /workspace/roboplan_ws
 COPY . src/roboplan
 
 # Install dependencies and build RoboPlan.
-RUN source /opt/ros/${ROS_DISTRO}/setup.bash \
+RUN --mount=type=cache,target=/var/lib/apt/lists,sharing=locked \
+    --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    source /opt/ros/${ROS_DISTRO}/setup.bash \
+    && apt-get update -y \
```
sea-bass, line 35: for my edification, why do you need a second apt-get update here? Shouldn't the first one break cache if something truly changed?

sea-bass (approved):

.docker/ros/Dockerfile
EzraBrooks: No, cache mount content changes do not trigger cache busts in Docker. Per the Dockerfile standard there are only two things that can trigger cache busts: ADD/COPY and build-arg changes. This is because the cache logic in Docker is just a primitive diff on the input artifacts, so if a step has no inputs it can't bust cache. 

So before we added this second apt update, it was still possible to get a stale apt cache from the previous install; it just came from outside the docker image (the new cache mount) rather than from inside the image as previous.

PR #311, opened 2026-09-05
Rename roboplan directory

EzraBrooks:
Closes #310, depends on #304

PR #312, opened 2026-09-06
Make cmeel packages independently buildable

### LINKED ISSUE #309: Make PyPI and Conda package structures match
In Conda, RoboPlan is a series of individual packages grouped together by a metapackage `roboplan-all`.

In PyPI, there's one singular wheel containing all packages; `roboplan`.

We should make each individual `pyproject.toml` in the repository describe a functional standalone package instead of being effectively a stub (the current situation) so we can publish them each individually to PyPI and people can pick and choose the parts they want.

EzraBrooks:
Fixes #309, depends on #311

Not a fan of how much CI code this is adding, but want to run it first and see if it works before simplifying.

This PR:

- Fixes up the dependencies in each `pyproject.toml` to make it possible to build each package as a separate wheel
- Introduces a `roboplan` metapackage for both ROS and PyPI to match Conda


pyproject.toml
sea-bass, line 1: Wait so can/should this top level file be a `pixi.toml` again at this point? It honestly doesn't matter that much either way

EzraBrooks: now that it's back to being solely `[tool.pixi.*]` config, it is a bit silly that it's a `pyproject.toml`.. I probably should have actually evaluated whether there was any load-bearing configuration in the `pyproject.toml` other than the PyPI packaging before I merged the two. my bad.. unnecessary churn.

EzraBrooks: I do wonder whether we could sneak a small block of `uv` configuration in here to manage the PyPI dependency graph so that we don't have to maintain that list of wheels to build in CI manually.. but I don't want to add yet more logic to this one PR

sea-bass: I know pixi uses uv internally (but as a library), so maybe there's a way to have this config live in pixi.toml format?

But if not, just leave it as a pyproject.toml for science.

sea-bass:
btw updated the conda feedstock to this branch in https://github.com/conda-forge/roboplan-feedstock/pull/22 and it still works! Actually now we can build the metapackage, stage all its libraries, and save on a lot of duplicated builds!

sea-bass (commented):
BTW one more request maybe we can try here? I was building with conda and had to incorporate a workaround. Here's the tl;dr from Claude:

---

Building `superbuild/` on conda-forge win-64 fails compiling `roboplan_oink` (and would fail the same way for `roboplan_rrt`, `roboplan_simple_ik`, `roboplan_cartesian_planning`):

```
roboplan_oink/CMakeFiles/roboplan_oink.dir/src/optimal_ik.cpp.obj
%PREFIX%\Library\include\pinocchio/src/container/aligned-vector.hxx(24): error C2059: syntax error: ''
%PREFIX%\Library\include\pinocchio/src/algorithm/fwd.hxx(26): error C2059: syntax error: ''
%PREFIX%\Library\include\pinocchio/src/algorithm/fwd.hxx(41): error C2065: 'RigidConstraintModel': undeclared identifier
...
```

Log: https://github.com/conda-forge/roboplan-feedstock/actions/runs/34060148530/job/101559134956?pr=22

**Why it happens**

- MSVC reports `__cplusplus` as `199711L` unless `/Zc:__cplusplus` is passed, regardless of `/std:c++20`.
- pinocchio's `deprecated.hpp` keys off `__cplusplus`. With the legacy value it expands
  `PINOCCHIO_DEPRECATED_MESSAGE(...)` to `__declspec(deprecated(...))` instead of
  `[[deprecated(...)]]`. pinocchio uses that macro on alias declarations
  (`using aligned_vector PINOCCHIO_DEPRECATED_MESSAGE("...") = std::vector<T>;`), where
  `__declspec` is not valid syntax, hence the C2059 errors inside pinocchio's own headers.
- pinocchio's installed CMake config takes care of this for consumers: `pinocchioConfig.cmake`
  calls `CHECK_MINIMAL_CXX_STANDARD` from its `cxx-standard.cmake`, which does
  `string(APPEND CMAKE_CXX_FLAGS " /Zc:__cplusplus")` on MSVC. That is a macro, so it mutates
  `CMAKE_CXX_FLAGS` in the directory scope that called `find_package(pinocchio)`.
- Standalone, every package calls `find_package(roboplan)`, which loads `roboplanConfig.cmake`,
  which calls `find_package(pinocchio)`, so every package's own scope gets the flag.
- In the superbuild, `roboplan_register_build_tree_packages()` resolves `find_package(roboplan)`
  to the in-tree targets, so `roboplanConfig.cmake` never runs. Only `roboplan_core` calls
  `find_package(pinocchio)` directly, and a `CMAKE_CXX_FLAGS` change in one subdirectory does not
  propagate to sibling subdirectories. Net effect: `roboplan_core` compiles with
  `/Zc:__cplusplus`, the sibling packages compile without it. The failing compile line confirms
  it: `-std:c++20 /permissive- /bigobj ...` and no `/Zc:__cplusplus`.

**Suggested fix**

Set the flag once at the superbuild top level, before the `add_subdirectory()` calls, so every package inherits it:

```cmake
# MSVC only reports the real __cplusplus with this flag; pinocchio's headers depend on it.
# Standalone builds get it as a side effect of find_package(pinocchio) via roboplanConfig.cmake,
# which the in-tree registration bypasses.
if(MSVC)
  add_compile_options(/Zc:__cplusplus)
endif()
```

An alternative is a top-level `find_package(pinocchio REQUIRED)` before the `add_subdirectory()` calls, which reproduces exactly what `find_dependency` does for standalone builds. The explicit flag is less dependent on a side effect of someone else's config file.

Until this lands, the feedstock works around it by adding `/Zc:__cplusplus` to `CXXFLAGS` before invoking CMake.

---

I'm partial to the second recommended fix TBH.

EzraBrooks:
if you want to go with the second fix I'd personally call it unrelated to this PR and should be separate

sea-bass:
> if you want to go with the second fix I'd personally call it unrelated to this PR and should be separate

Fair, I can test it out with my conda-forge PR after this lands, plus any other follow-ons that may affect this.

sea-bass (approved):

sea-bass:
Just updated the branch rulesets based on the changes in this PR

PR #316, opened 2026-09-08
Remove cmake build-time package config hack

EzraBrooks:
Apparently this is "the right way" by CMake convention (at this CMake version) to handle packages that might be either in-tree or out-of-tree. It's ugly and repetitive, but it eliminates a fairly arcane hack in the superbuild.


EzraBrooks:
whoops, this includes unnecessary format changes. I'll implement a CMake formatter in pre-commit in another PR then rebase this.

EzraBrooks:
^ #317

EzraBrooks:
This was present in the original wheel superbuild and was migrated into the unified superbuild along the way, but from what I can see online it seems like this "generating stub config files during the build" hack is common but somewhat frowned upon.

In the spirit of @ChrisThrasher's immortal guidance on CMake, we should just keep things simple.. if we need to "find a package if we don't already have it in tree".. we should just write an if statement that expresses that.

sea-bass (approved):
conflict with your other PR I just merged, but LGTM pending that as it's also less cmake

ChrisThrasher:
I’m long removed from the ROS world which mixes in-tree and out-of-tree deps, but generally speaking `find_package` is designed to be called as many times as you desire. There is no requirement to only call it once. However `find_package` is also designed to only access out-of-tree prebuilt dependencies so if that precondition is not met, additional hacks may be required.

PR #319, opened 2026-09-09
General CMake cleanup

EzraBrooks:
While finishing up all my recent CMake changes, figured I'd clean up the dependency graph a bit.

This PR:

- moves the `roboplan_filters` code into the core build target (rationale: this depends upon Pinocchio but not other RoboPlan modules, therefore it is a core API - otherwise we have two "bottom-most" libraries in RoboPlan)
- cleans up various `target_link_libraries` invocations, de-duplicating some transitive deps that were re-declared as direct deps and adding a few visibility specs where they were missing - helped with using `cmake --graphviz` to clean up the former

### DIFF
--- a/roboplan_simple_ik/test/CMakeLists.txt
+++ b/roboplan_simple_ik/test/CMakeLists.txt
@@ -5,7 +5,7 @@ find_package(roboplan_example_models REQUIRED)
 add_executable(test_simple_ik test_simple_ik.cpp)
 set_property(TARGET test_simple_ik PROPERTY CXX_STANDARD 20)
 target_link_libraries(
-  test_simple_ik roboplan_simple_ik roboplan::roboplan
+  test_simple_ik roboplan_simple_ik
   roboplan_example_models::roboplan_example_models GTest::gtest_main)
 
 # If ament_gtest is detected, make the tests visible in ROS 2.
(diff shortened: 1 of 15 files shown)


sea-bass (approved):

roboplan_simple_ik/test/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ find_package(roboplan_example_models REQUIRED)
 add_executable(test_simple_ik test_simple_ik.cpp)
 set_property(TARGET test_simple_ik PROPERTY CXX_STANDARD 20)
 target_link_libraries(
-  test_simple_ik roboplan_simple_ik roboplan::roboplan
+  test_simple_ik roboplan_simple_ik
   roboplan_example_models::roboplan_example_models GTest::gtest_main)
```
sea-bass, line 9: did you intend to add a visibility spec here too?

sea-bass: @EzraBrooks my bad, I noticed this after approving. We can resolve this comment and this is good too

EzraBrooks: I probably intended to but visibility is not as big a deal in the test targets as removing duplicate deps

PR #326, opened 2026-09-13
Per package PyPi publishing and runtime path setting

sea-bass:
Fixes the `release.yml` workflow to actually be able to publish all the PyPi wheels.

To test, I temporarily tagged a `0.0.104` dummy release which you can see in https://pypi.org/search/?q=roboplan

I also in the background had to swap out the single `pypi` environment for trusted publishers to per-package environments, e.g. `pypi-roboplan-core`.


sea-bass:
@EzraBrooks FYI

sea-bass:
Actually it seems the PyPi wheels for this new version do not work.

To try this:

1. Create a virtual env or whatever
2. `pip install roboplan==0.0.101`
3. In Python, do:

```
>>> from roboplan.core import Scene
  File "<stdin>", line 1, in <module>
  File "/home/sebastian/test/testpypi/lib/python3.12/site-packages/cmeel.prefix/lib/python3.12/site-packages/roboplan/core/__init__.py", line 6, in <module>
    from ._core_ext import *  # noqa: F401,F403
    ^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: libroboplan.so: cannot open shared object file: No such file or directory
>>> 
```

any ideas?

sea-bass:
OK fixed here and tagged `0.0.102` -- this one works!

roboplan_rrt/pyproject.toml
```diff
@@ -35,4 +35,10 @@ roboplan-rrt = { pin-subpackage = true }
 
 [tool.cmeel]
 run-tests = false
-configure-args = ["-DBUILD_PYTHON_BINDINGS=ON", "-DGENERATE_PYTHON_STUBS=OFF"]
+configure-args = [
+    "-DBUILD_PYTHON_BINDINGS=ON",
+    "-DGENERATE_PYTHON_STUBS=OFF",
+    # Wheels put libraries in cmeel.prefix/lib and extensions four levels below
+    # it; this lets both find their dependencies relative to themselves.
+    "-DCMAKE_INSTALL_RPATH=$ORIGIN;$ORIGIN/../../../../;@loader_path;@loader_path/../../../../",
```
EzraBrooks: Should we move this detection into each CMakeLists where there's already linking shenanigans going on?

EzraBrooks (commented):
Thanks for identifying and fixing this issue, I wonder how we can catch this in CI in the future

roboplan_rrt/pyproject.toml
sea-bass: it was actually originally in CMakeLists (see https://github.com/open-planning/roboplan/pull/326/changes/f5b6a87db3a4f193a63eb43e5c45ce9b479c587f) and I thought it was kind of a lot to check it being a cmeel build as yet another conditional vs. throwing it in here, but I can see how this is more principled even if harder to read.

What do you think?

EzraBrooks: I think since there's already the workaround for Apple SIP in the CMakeLists.txt we should group RPATH futzing together.

Though I'm also questioning why we need this, given that I haven't found examples in other libraries of having to do this

EzraBrooks: `cmeel-assimp` is the only thing I've found to be futzing with RPATH but it's just the same macOS thing:

https://github.com/cmake-wheel/cmeel-assimp/blob/cmeel/cmeel.patch

sea-bass: > Though I'm also questioning why we need this, given that I haven't found examples in other libraries of having to do this

Valid question. It's a combination of many other projects being standalone (so they don't depend on libraries from other binaries), and also using tools like audit/delocate between build and publishing to "repair" the paths. cibuildwheel has options for that actually.

sea-bass: Pinocchio uses this: https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/relpath.cmake

roboplan_cartesian_planning/CMakeLists.txt
```diff
@@ -15,6 +15,16 @@ if(WIN32)
...
 
+# cmeel wheels: libraries land in cmeel.prefix/lib and extensions four levels
+# below it, so let both find their dependencies relative to themselves.
+if(DEFINED ENV{CMEEL_BUILD})
+  if(APPLE)
+    set(CMAKE_INSTALL_RPATH "@loader_path;@loader_path/../../../../")
+  else()
+    set(CMAKE_INSTALL_RPATH "$ORIGIN;$ORIGIN/../../../../")
+  endif()
+endif()
```
EzraBrooks: With `-DBUILD_PYTHON_BINDINGS=ON`, this will call `add_subdirectory(bindings)` at the end, which sets the `INSTALL_RPATH` on specifically the `_ext`. Shouldn't that be where this lives since it's only relevant to the extension finding its parent?

roboplan_rrt/pyproject.toml
EzraBrooks: > Valid question. It's a combination of many other projects being standalone (so they don't depend on libraries from other binaries), and also using tools like audit/delocate between build and publishing to "repair" the paths. cibuildwheel has options for that actually.

maybe I copied a bad example and we should re-enable the CIBW audit step? after these changes, does it now pass?

sea-bass: Oh where is that to re-enable? I can try kick of a new dummy tag and build while I work

EzraBrooks: remove my overriding of the `CIBW_REPAIR_WHEEL_COMMAND` env variable in `build-pypi-wheels`. I copied it from cmeel's example codebase.

roboplan_cartesian_planning/CMakeLists.txt
sea-bass: apparently (according to the clankers) it's not readily movable, because if your extension depends on a library which in turn depends on another library, you could run into issues there too.

The example it gave was the roboplan_core extension depending on libroboplan.so, which in turn depends on a vendored tinyxml2.so.

Fixing in just the bindings subdirectory would fix the first path but not the second.

roboplan_rrt/pyproject.toml
sea-bass: Again quoting the clankers here because this is very out of my depth, but apparently enabling CIBW repair would work, but cause a lot of the libraries to be copied individually to each wheel, because it aims to make each wheel self-contained. So the RPATH changes are ostensibly a way around that.

This supposedly is also why Pinocchio doesn't go down the `auditwheel` / `delocate` path

roboplan_cartesian_planning/CMakeLists.txt
EzraBrooks: should be quick enough to test, though, right? Have you already tested whether the transitives fail to resolve with the current system to isolate it from just being the first link between the extension and the core lib?

roboplan_rrt/pyproject.toml
EzraBrooks: yea that was my interpretation and why I turned it off

roboplan_cartesian_planning/CMakeLists.txt
sea-bass: Yeah, just tested:

```
  File "<string>", line 1, in <module>
    from roboplan.core import Scene
  File ".../cmeel.prefix/lib/python3.13/site-packages/roboplan/core/__init__.py", line 6, in <module>
    from ._core_ext import *  # noqa: F401,F403
    ^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: libpinocchio_parsers.so: cannot open shared object file: No such file or directory
```

so basically it finds libroboplan.so, but then fails to go up to the next shared library :(

---

this is the equivalent error without _any_ rpath futzery:

```
  File "<string>", line 1, in <module>
  File ".../cmeel.prefix/lib/python3.13/site-packages/roboplan/core/__init__.py", line 6, in <module>
    from ._core_ext import *  # noqa: F401,F403
    ^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: libroboplan.so: cannot open shared object file: No such file or directory

```

EzraBrooks:
So, funny thing - the `0.0.101` release that doesn't work for you on Linux does work on macOS. Which implies to me we are definitely doing something wrong here 😆

sea-bass:
> So, funny thing - the `0.0.101` release that doesn't work for you on Linux does work on macOS. Which implies to me we are definitely doing something wrong here 😆

So if you look on `main` could it be that the path messing about being done only for the macOS case (this is the SIP thing you guys were doing?) made it work on macOS, but not for Linux which is what I tested on?

https://github.com/open-planning/roboplan/blob/eac8ca779dce417d0b280930530ba905d29c7b23/roboplan_core/bindings/CMakeLists.txt#L102-L116

EzraBrooks:
That's what I'm curious about. Do we need to do that on all operating systems, just using `$ORIGIN` rather than `@loader_path` on non-macOS POSIX platforms?

sea-bass:
Nice, that seems to have worked locally. I just made a `0.0.103` tag for round-trip testing, so let's see how that goes in like 1 hr.

EzraBrooks:
seems like your find and replace was a bit overly aggressive 😆

sea-bass:
lol my search/replace was too aggressive, but it should only affect pixi and linting before I revert it, so whatever

sea-bass:
OK I tested `0.0.103` on Ubuntu 24.04 and works for me! mac bois?

(Ignore the CI failure, it was a residual tag that I since deleted which kicked off a bad release build)

EzraBrooks:
still works on macOS

.github/workflows/build-pypi-wheels.yml
```diff
@@ -199,10 +199,14 @@ jobs:
...
-            roboplan roboplan-example-models roboplan-rrt roboplan-simple-ik
+            roboplan-example-models==${{ needs.check_version.outputs.version }}
+            roboplan-rrt==${{ needs.check_version.outputs.version }}
+            roboplan-simple-ik==${{ needs.check_version.outputs.version }}
           CIBW_BEFORE_TEST_MACOS: >-
             pip install --find-links ${{ github.workspace }}/wheelhouse
-            roboplan roboplan-example-models roboplan-rrt roboplan-simple-ik
+            roboplan-example-models==${{ needs.check_version.outputs.version }}
+            roboplan-rrt==${{ needs.check_version.outputs.version }}
+            roboplan-simple-ik==${{ needs.check_version.outputs.version }}
```
eholum: Is there a reason we're only referencing rrt and simple ik here? Do we not need the others considering the imports below (core, oink, cartesian planning)?

sea-bass: ooh nice find. I assume the existing state was just testing a few packages to get the general gist of whether it worked, but no reason not to include them all.

.github/workflows/build-pypi-wheels.yml
```diff
@@ -197,12 +197,8 @@ jobs:
           package-dir: roboplan_cartesian_planning
           output-dir: wheelhouse
         env:
-          CIBW_BEFORE_TEST_LINUX: >-
-            pip install --find-links /host${{ github.workspace }}/wheelhouse
-            roboplan roboplan-example-models roboplan-rrt roboplan-simple-ik
-          CIBW_BEFORE_TEST_MACOS: >-
-            pip install --find-links ${{ github.workspace }}/wheelhouse
-            roboplan roboplan-example-models roboplan-rrt roboplan-simple-ik
+          # The metapackage pins every subpackage, so this pulls the full set from the wheelhouse.
```
eholum: Oh yeah clean

.github/workflows/release.yml
```diff
@@ -32,12 +49,28 @@ jobs:
...
+          # Wheels use the normalized name (underscores); sdists may use either form.
+          # The "-[0-9]" suffix keeps "roboplan-*" from matching "roboplan-core-*" and friends.
+          for f in dist/"${PACKAGE//-/_}"-[0-9]* dist/"${PACKAGE}"-[0-9]*; do
+            [ -e "$f" ] && mv "$f" package/
+          done
+          ls -la package
+          if [ -z "$(ls -A package)" ]; then
+            echo "::error::No distributions found for $PACKAGE"
+            exit 1
+          fi
```
EzraBrooks: Inline bash code in CI jobs spooks me nowadays, especially when it involves regular expressions and other string-escaping-sensitive kinds of code. Can we extract this logic to a Python script with defined arguments and run it separately?

Or, if possible (I'm not familiar with this distribution logic), we could change something about the sdist generation to make it conform to the same pattern as the wheel so we don't need this in the first place?

sea-bass: on it... also wtf Claude just yolo pushed changes on me before I could review them, fixing

sea-bass:
Tagged a 0.0.104 for sanity

EzraBrooks (approved):
LGTM. at some point I want to do a pass on all the CI jobs myself to extract the code into scripts.. and figure out how to use the Actions infrastructure to avoid adding `checkout` to jobs that don't require the source code

PR #327, opened 2026-09-14
Add MuJoCo simulation backend

Muhammad540:
## MuJoCo simulation backend for RoboPlan
This PR introduces **_roboplan_mujoco_** as an optional simulation backend built on top of RoboPlan, without coupling MuJoCo to or modifying the core planning library.

The current integration allows applications to:
- build a MuJoCo model derived from RoboPlan's canonical robot description, keeping the canonical model as the single source of truth
- attach the robot to an MJCF simulation world
- add position servos for selected controlled joints
- map positions, velocities, and efforts between RoboPlan/Pinocchio and MuJoCo representations
- execute RoboPlan position trajectories in MuJoCo
- use a kinematic-only interface for configuration playback
- render the live simulation with task space paths and ghost robot configurations

## Design
The package is divided into several components:
- `MujocoModelBuilder` owns and configures a mutable `mjSpec`.
- `MujocoSimulation` owns the compiled `mjModel` and its associated `mjData`.
- `MujocoJointMap` validates and precomputes the mapping between RoboPlan and MuJoCo joint representations.
- `MujocoHardwareInterface` implements the common RoboPlan hardware interface and owns the simulation used for position controlled execution.
- `MujocoKinematicInterface` supports direct configuration updates without actuators or physics stepping.
- `MujocoRenderer` observes a simulation without owning it and provides optional planning overlays.
The ownership chain is intentionally explicit and move only:
`MujocoModelBuilder → MujocoSimulation → MujocoHardwareInterface`
The renderer only borrows the simulation, so the simulation owner must outlive it.

Application/example own the control loop and are responsible for advancing the simulation and rendering frames.

roboplan_mujoco_arch_diagram

## Robot model preparation
Canonical robot descriptions remain under `roboplan_example_models`. A preparation script derives a MuJoCo compatible URDF by:
- expanding Xacro files
- converting unsupported mesh formats
- preserving material colors
- generating unique geometry names
- removing planar joint limit tags that are not accepted by MuJoCo's URDF importer

## Example
A Franka pick and place example demonstrates the complete workflow:
1. create a RoboPlan scene and add the planning environment
2. plan the approach and carrying trajectories
3. construct and compile the MuJoCo model
4. create the hardware interface
5. send RoboPlan position commands to Mujoco
6. step and render the simulation

https://github.com/user-attachments/assets/e8f384b5-aa0e-4132-ab06-468f222dd74c

**What do you think about this direction?**

I think a physics based simulation backend could be valuable for validating RoboPlan trajectories and developing integrations without requiring access to a physical robot. I also introduced a common hardware interface boundary so that MuJoCo is one backend rather than being coupled directly to the planning code. The same boundary could later be implemented by real robot drivers or other sim backends.


sea-bass:
Overall, I think this is too large a feature to bring into this core repository. Specifically, I don't think it's in our [design philosophy](https://roboplan.readthedocs.io/en/latest/design/philosophy.html#not-a-monolith) to add any kind of hardware/simulator interface layer to this repo. That should all be handled separately.

I might recommend creating your own separate repo if you'd like to explore this, as there might be interest for others to use this as an optional add-on?

I also see a lot of parallels with [`mujoco_ros2_control`](https://github.com/ros-controls/mujoco_ros2_control), but without the ROS part in this case. In fact, the URDF to MJCF converter stuff you did might be a great contribution avenue to that repo because what we have there now is... not ideal, and doesn't rely on `mjSpec` as much as it probably should.

roboplan_example_models/models/franka_robot_model/dual_fr3.urdf
```diff
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='utf-8'?>
-<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="fr3">
+<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="dual_fr3">
```
sea-bass, line 2: thank you -- I will bring this into a cleanup PR I have, if you don't mind!

roboplan_example_models/models/mujoco/franka/fr3_assets/finger_12c7bc9b55ff_0.obj
sea-bass, line 1: where did these OBJ files come from? Are they auto-generated from the mujoco model builder API in here? If so, maybe not worth adding them to version control (the changes are so large they are freezing the GitHub diff page)

Muhammad540: yes these OBJ files are generated by **_prepare_urdf.py_** script. This script converts the meshes from the RoboPlan model (dae) to OBJ, and model builder then loads the generated urdf. I have now removed the generated urdf and mesh assets.

roboplan_example_models/models/franka_robot_model/dual_fr3.urdf
Muhammad540: okay :+1:

Muhammad540:
> Overall, I think this is too large a feature to bring into this core repository. Specifically, I don't think it's in our [design philosophy](https://roboplan.readthedocs.io/en/latest/design/philosophy.html#not-a-monolith) to add any kind of hardware/simulator interface layer to this repo. That should all be handled separately.
> 
> I might recommend creating your own separate repo if you'd like to explore this, as there might be interest for others to use this as an optional add-on?
> 
> I also see a lot of parallels with [`mujoco_ros2_control`](https://github.com/ros-controls/mujoco_ros2_control), but without the ROS part in this case. In fact, the URDF to MJCF converter stuff you did might be a great contribution avenue to that repo because what we have there now is... not ideal, and doesn't rely on `mjSpec` as much as it probably should.

Thanks for the feedback, that makes sense. I tried to keep mujoco optional and separate from the core, but I understand that this is still a large addition to the repository. I’ll continue developing it separately as an optional add on. I’ll take a look at **_mujoco_ros2_control_** and see whether the mjSpec based model construction could be useful there.

Thanks for taking the time to review this and point me in the right direction. I’d still love to contribute to RoboPlan, especially around its future robot learning direction. Are there any particular areas that would be useful to explore, such as reinforcement learning, behavior cloning, or supporting infrastructure?

sea-bass:
> Are there any particular areas that would be useful to explore, such as reinforcement learning, behavior cloning, or supporting infrastructure?

For sure! I think being able to use RoboPlan in the imitation learning pipeline -- i.e., teleop / data collection and also action chunk rollout with something like LeRobot -- would be a fantastic example that could use contribution.

This would mean things like:

* Using our OInK IK solver for Cartesian space teleop / action chunks
* Having some kind of trajectory optimization to track joint or Cartesian space action chunks without violating kinematic/dynamic limits
* Having MuJoCo examples that are simpler than this proposed approach: for example, just a Python script that does some kind of motion planning and directly commands mujoco via mjData, in Python.

EzraBrooks:
I agree entirely with @sea-bass's take above that this doesn't belong in the core repo but you should host it yourself as your own library in case people want to use it :)

relatedly, #300 should be helpful here when it is done.. I'll try to stop procrastinating on it

PR #329, opened 2026-09-15
Rename roboplan-core's export target

eholum:
Post: https://github.com/open-planning/roboplan/pull/312/changes#diff-54be2fc37a965867780727839023b9b4586c1d6c6f3029b69bb3adc240b033c1

Ran into this transient issue with `colcon` [in the ROS wrappers](https://github.com/open-planning/roboplan-ros/actions/runs/34347572689/job/102452762805#step:5:2978), since linking against `roboplan` could either hit the meta package or the core package.

This renames the core package's export to `roboplan-core` and update all references to it accordingly. I think this is probably what we want, anyway? As opposed to somehow finagling `roboplan` to play nicely with this export?


eholum:
Leaving in draft while testing in https://github.com/open-planning/roboplan-ros/pull/67

Edit: CI goes Brrrrrrrrrrrrrrrrrrr and so far seems ok?

eholum:
Lemme know what you guys think @sea-bass and @EzraBrooks, I'm a little lost in the sauce playing catch up.

sea-bass (approved):
This makes sense to me!

EzraBrooks:
thanks. this was on my mental backlog but I forgot to file an issue

PR #332, opened 2026-09-15
Fix building with latest XCode CLT

EzraBrooks:
Also align our LLVM versions across OSes while we're at it, because otherwise when this upstream LLVM PR is merged we will have _three different LLVM versions_ in our toolchain


EzraBrooks:
Upstream `lld` PR: https://github.com/llvm/llvm-project/pull/188378

EzraBrooks:
Actually, @eholum @sea-bass, what would you think about going one step further and removing our pins of `clang/clangxx` and `gcc/gxx` and instead pinning `c-compiler` and `cxx-compiler` at version `*` to track the upstream conda-forge compiler toolchain?

EzraBrooks:
Well, putting this one up for review regardless of whether we want #333 or not, since without this one I literally can't compile on my machine currently

sea-bass (approved):
LGTM!

PR #333, opened 2026-09-15
Try using upstream conda forge compilers

EzraBrooks:
wip based on #332


EzraBrooks:
lol, conda-forge `gcc` too old for `-fuse-mold`, rip.

EzraBrooks:
this is not very valuable right now and is nerd sniping me and sucking up my time because of the classic "clang-tidy doesn't work right with GCC" issue. I need to drop this

PR #335, opened 2026-09-16
Switch from Black to `ruff format`

EzraBrooks:
We need to add `ruff check` and actually lint our Python bindings and examples, and `ruff format` is nearly identical to Black and much faster, so this is a good first step.


sea-bass (approved):
yes -- all for `ruff` and `ty`

PR #336, opened 2026-09-16
Add ruff check

EzraBrooks:
This is gonna require changing CI, I think, since you need an installed environment to run the linter.


sea-bass:
can we have the format CI run pre-commit from pixi?

So a `setup-pixi` task -> a lint task

EzraBrooks:
yeah but it's actually gnarlier than that because we will need to compile and install the roboplan codebase to lint/typecheck effectively (otherwise `import roboplan.core` won't resolve). So `ruff check` actually needs to depend upon `pixi run install`

EzraBrooks:
> yeah but it's actually gnarlier than that because we will need to compile and install the roboplan codebase to lint/typecheck effectively (otherwise `import roboplan.core` won't resolve). So `ruff check` actually needs to depend upon `pixi run install`

I guess not the default `ruff check` rules, since this PR is passing. But some of the advanced rules and definitely `ty` would require this

EzraBrooks:
argh, PRs in forks can't be added to stacks yet. well, this is blocked by #335 then

sea-bass:
For future, you should be able to make branches in this repo now!

roboplan_core/bindings/python/roboplan/interpolation.py
```diff
@@ -2,7 +2,6 @@
 
 import numpy as np
 import pinocchio as pin
-
```
sea-bass, line 5: Is there a way to configure this to leave a newline between non-system imports and then roboplan stuff as a third bucket? I recall one can

sea-bass (commented):
Superior linting, love the import sorts. I just have 1 question but otherwise lgtm once/if we can resolve that

roboplan_core/bindings/python/roboplan/interpolation.py
sea-bass: I guess it's something like this in the appropriate toml manifest:

```toml
[tool.ruff.lint.isort]
known-first-party = ["roboplan"]
```

worth doing?

Might also be good doing it for `common` in the examples.

EzraBrooks: Did it at the workspace root (where `ruff` is being invoked right now) but not for `common` in the examples because I think that would require invoking `ruff` multiple times across multiple packages? I'm not actually sure about that though

sea-bass (approved):
yay

roboplan_core/bindings/python/roboplan/interpolation.py
sea-bass: sounds good. Another option is to replace `from common import *` to `from roboplan_examples.common import *` and throw that in as well. Up to you though

PR #339, opened 2026-09-16
Always initialize the state space of the RRT planner

### LINKED ISSUE #338: Passing a planning group name of `""` breaks initialization of RRT planner
The "group unchanged" logic causes the update to return early because `""` is the default value.

EzraBrooks:
Fixes #338. Previously, the default group name of `""` was preventing anyone else from being lazy and using the group name `""`.

Making this field a `std::optional` would probably create unnecessary API pain, so we just need to either make the default value work (this PR) or we could instead reject the default value from consumers.


EzraBrooks:
@sea-bass another way to fix this bug is to just remove the early-return from the `setOptions` call, which personally I'd prefer. If I tell software to do something, I want it to do it, not silently refuse

sea-bass (approved):
Thank you! So the reason for that early return is that if you are only changing non-structural options (i.e., not the group name), then you don't need to re-initialize the entire `dynotree` state space. So for example if you keep the group the same but you're modifying connection distances, or RRT algorithm, etc., why bother reinitialize variables that will not change?

So looking at this, and weighing the 2 other options which are:

1. Add another boolean state / make the state-space an `optional` so you can check if it's initialized
2. Always reinitialize the state space

This is a good compromise. Especially because it also means you're not assigning the `options_` member variable twice on construction as is the case on `main` lol

PR #342, opened 2026-09-18
Add helper function to allow all adjacent links to collide

### LINKED ISSUE #337: Add a function on Scene to allow all collisions between adjacent links
An obvious follow-up to this PR is going to be adding a helper function to propagate allowed collisions between adjacent links, much the same way MSA does when generating an SRDF; but at runtime.

_Originally posted by @EzraBrooks in https://github.com/open-planning/roboplan/issues/300#issuecomment-5701438411_

EzraBrooks:
Closes #337


sea-bass (commented):
LGTM, love how the function immediately cleaned up the mujoco example.

My only real (optional) comment is if we can slightly shorten the function name to like

`allowAdjacentLinksToCollide`
or
`allowAdjacentLinkCollisions`

or something like that

roboplan_core/include/roboplan/core/scene.hpp
```diff
@@ -524,6 +524,13 @@ class Scene {
   tl::expected<void, std::string> setCollisions(const std::string& body1, const std::string& body2,
                                                 const bool enable);
 
+  /// @brief Allows collisions between every parent-child link pair in the kinematic tree.
+  /// @details Revolute joints are typically modeled with no volume, so parent and child link
+  /// meshes always overlap. Almost all robots need this unless an SRDF already disables those
```
sea-bass: I'd omit the revolute joint comment TBH, probably too specific

EzraBrooks: This needs some explanation for _why_ you want to almost always do this, if we want roboplan to be accessible to beginners. This particular verbiage might be a bit verbose/specific though.

sea-bass: It's less the verbose, more the specific. We could say something more general like

> Almost all robots need this since adjacent link geometries may overlap across joint boundaries and cause collision checking to fail on valid configurations.
> Consider calling this if you are not using SRDF to explicitly remove collision pairs.

sea-bass (approved):

PR #324, opened 2026-09-13
Pre-0.7.0 docs and consistency passthrough

sea-bass:
Unleashing the clankers on the documentation and code to cleanup any inconsistencies, and I also made some of my own changes that I was previously aware of.

Besides docs, I had some member functions renamed to be `camelCase` instead of `snake_case`, which is a breaking change but we already have a bunch of those incoming.

Will keep this branch open/in draft until all the other PRs before release are good to go.

PR #344, opened 2026-09-18
Add group helpers, exercise them via new `importSrdf`

### LINKED ISSUE #340: Allow imperatively creating and modifying joint groups in the scene
Following the pattern of #337, I believe we should expose an API like so:

```python
scene = (
    Scene("panda", loadMjcfModel(mjcf_path))
      .allowAllAdjacentLinksToCollide()
      .addJointGroupFromChain("arm", base="panda_link0", tip="panda_hand")
      .addJointGroup("gripper", joints=["finger1", "finger2"])
      .addJointGroupFromGroups("arm_with_gripper", ["arm", "gripper"])
)
```

Because this would allow you to create a scene from an upstream MJCF or URDF or even URDF/SRDF pair that is immutable to you - like an Apt-installed `-description` package or something from `robot_descriptions.py` - and use that as the source of truth for the _robot itself_ but add all your own planning logic and world objects.

For example, the MuJoCo pick-and-place example #190 would become:

1. Load the Panda MJCF from `robot_descriptions.py`
2. run the code snippet from the beginning of this issue :point_up:
3. spawn a cube and plan around it without considering the gripper or the cube as part of the robot:

```python
scene.addBoxGeometry("cube", "universe", Box(0.04, 0.04, 0.04), T, color)

plan = RRT(scene, RRTOptions(group_name="arm")).plan(start, goal)
```

### LINKED ISSUE #343: Move SRDF import out of constructor into member function to decrease constructor specialization
This takes us closer to the "just construct the Scene with a Pinocchio model" API, and also allows SRDF to be optionally used with any geometry/kinematics model, i.e. MJCF.

EzraBrooks:
This PR removes:

- The SRDF constructor parameter we just made optional in #300

This PR adds:

- `Scene::addGroup(name, joints[], extra_link_names[])` - Manually create a group of joints by joint names
- `Scene::addGroupFromChain(name, base, tip)` - Automatically create a group of joints by traversing the tree from base to tip
- `Scene::addGroupFromGroups(name, groups[])` - Automatically create a group of joints by composing it out of other groups (i.e. your typical full-body "arm and gripper" or "base and arm and gripper" groups)
- `Scene::importSrdf` - load your collision settings from SRDF, still using Pinocchio's built-in helper for the collision pairs. Internally uses `addGroup` to add the groups from the SRDF to the Scene's group info to exercise that API.

Closes #343 
Closes #340


sea-bass (commented):
My first reaction is what I was trying to get to in the PR review last night: if we do this, then should we immediately eject `UrdfSceneDescription`, i.e.,

```cpp
PinocchioSceneDescription loadUrdfModel(const std::filesystem::path& urdf_path);

PinocchioSceneDescription loadMjcfModel(const std::filesystem::path& mjcf_path);
```

and then just add the new `importSrdf` function after constructing scenes?

or is the intention to still keep around the convenient URDF + SRDF path with `UrdfSceneDescription` + optional SRDF field that does this all for you?

I'm basically thinking about the "one unambiguous way to do X" philosophy, but somewhat sensitive of forcing people to write one extra line literally everywhere.

EzraBrooks:
Yes, agreed. I love it when refactors naturally fall out of previous refactors :)

EzraBrooks:
ah, actually, `UrdfSceneDescription` has to remain for now - unless we drop support for Humble and move to Pinocchio 4.1 across the board - because of the jerk and acceleration limits.

sea-bass:
dropping humble is quite tempting... @eholum ? <_<

(acknowledging that humble could work again once Pinocchio 4 works there)

eholum:
You had my curiosity

> dropping humble

Now you have my attention

sea-bass:
this would also let us drop that temporary struct for URDF extended limits, so I'd say let's forget humble in our CI until/unless external factors make it work with Pinocchio 4 again

EzraBrooks:
That's probably worth a separate PR (esp. since we squash by convention in this repo), so I can go make that change separately and then rebase this.

EzraBrooks:
#346 🚀

EzraBrooks:
Sorry for the whiplash on the API, but I think we got to a nice place where the Scene constructor is no longer dependent on model format specifics at all.

roboplan_core/include/roboplan/core/scene.hpp
```diff
@@ -527,8 +563,9 @@ class Scene {
   friend std::ostream& operator<<(std::ostream& os, const Scene& scene);
 
 private:
-  void initialize(const std::filesystem::path& yaml_config_path,
-                  const std::unordered_map<std::string, JointGroupInfo>& joint_group_info_map);
+  void initialize(const std::filesystem::path& yaml_config_path);
+  tl::expected<void, std::string> updateCollisionPairs(const std::string& body1,
+                                                       const std::string& body2, bool enable);
```
sea-bass: can you add docstrings to these? I know they're private, but this is more for people/agents reading the code rather than the docs.

roboplan_core/include/roboplan/core/scene.hpp
```diff
@@ -527,8 +563,9 @@ class Scene {
   friend std::ostream& operator<<(std::ostream& os, const Scene& scene);
 
 private:
-  void initialize(const std::filesystem::path& yaml_config_path,
-                  const std::unordered_map<std::string, JointGroupInfo>& joint_group_info_map);
+  void initialize(const std::filesystem::path& yaml_config_path);
+  tl::expected<void, std::string> updateCollisionPairs(const std::string& body1,
```
sea-bass: So if I understand this correctly, this new helper was added so you can update collisions multiple times without constantly rebuilding the broadphase manager. Which is great in principle.

But then I was left wondering if people have their own applications where they do want to set multiple collision pairs at once without rebuilding the broadphase manager, should we make this public? Then the concern here is footgun if people call this without then updating the broadphase manager.

So then taking it further, does it make sense backpedal on this extra function and revisit `setCollisions()` to also have a batch API? Or make a differently named function that allows the same?

sea-bass (commented):
Looks good -- just one question really on `setCollisions` vs. `updateCollisionPairs` and whether there's a better path.

Also, I suspect that there is significant updating the `docs/` folder to be done here. While we've been stashing stragglers in https://github.com/open-planning/roboplan/pull/324, it might be preferable to update all the scene construction stuff with this PR once we address the question below.

roboplan_core/include/roboplan/core/scene.hpp
EzraBrooks: I like the idea of having `setCollisions` work in batches.

I think the sensical API for that is `setCollisions(std::vector<std::pair<std::string, std::string>>, bool)`. thanks C++ verbosity.

Should I leave the non-vector API as two `std::string` or should we change it to be a `std::pair<std::string, std::string>` to match?

sea-bass: Cool cool. I'm thinking leave it as two strings since I don't have a strong preference therefore least changes wins marginally?

EzraBrooks: f66ffea does this. I don't have a strong preference either, since it's between "don't break yet another API" and "have the API be slightly more consistent". I can't really think of a reason why it would matter to a consumer since the entire point of the batch API is to _not_ have a foreach loop over a set of pairs of links.

roboplan_core/include/roboplan/core/scene.hpp
EzraBrooks: both functions are gone now. `updateCollisionPairs` was redundant to `setCollisions`, as you pointed out. `initialize` was only extracted constructor logic because the constructor was overloaded; I'll re-inline it for now, YAGNI.

It does make me wonder whether we want to make one last cheeky API change here before 0.7.0 and move the YAML config loading out of the constructor too though

sea-bass: That makes sense.

The only other one is package_paths and that's already specific to the URDF loader, so this would mean no other default args in the constructor, right?

EzraBrooks: Yeah, the Scene constructor would just take the scene name and the Pinocchio models.

The logic is already extracted to `overrideJointLimitsFromYaml`, so this is actually a very small change since we wouldn't yet need to add the family of `Scene::setJoint{limit_type}Limit` functions. Which we should add so people can modify their limits at runtime.

Every time I've talked to someone who works on a flight arm for a space project they ask "but can I programmatically lock a joint in place when its motor fails?"

EzraBrooks: Does the YAML syntax used here match the one people already might have on-hand coming from MoveIt? If so, I feel like we should keep it as a migration helper but recommend setting limits imperatively using your own application's config system (or declaratively via URDF v1.2), then deprecate it later if this library has its own momentum that isn't just "people annoyed with MoveIt".

sea-bass: A tangent, but Pinocchio has a `buildReducedModel()` function that can do the locking thing much more efficiently than setting its limits to stop motion... that is, it actually removes the joints and reduces the dofs of the whole model.

sea-bass: > Does the YAML syntax used here match the one people already might have on-hand coming from MoveIt?

Nope, this one is orthogonal, but same concept. Besides the acceleration/jerk limits, sometimes you can't (or don't want to) modify URDFs and instead have motion planning specific limits.

I'd be OK with switching to imperative, but doesn't have to be for this release.

sea-bass: Critical difference is that moveit uses sentinel  boolean values (e.g., `has_position_limits`) in both config and backend whereas this one ignores non-existent fields and holds them as `std::optional` in the backend

EzraBrooks: ok, how about we do this:

- *0.7.0*
  - add a YAML config loader function which for now returns a `YAML::Node`
  - add `Scene::importJointLimitsFromYaml(YAML::Node)`
  - remove the YAML loading from the constructor
- *later*
  - replace that `YAML::Node` type with a validated struct
  - add imperative API, refactor internals of `importJointLimitsFromYaml` to use it
  - observe user behavior and feature requests for a few releases
  - decide to deprecate `importJointLimitsFromYaml`?

sea-bass: Sounds perfect!

sea-bass (approved):

PR #346, opened 2026-09-18
Use Pinocchio 4.1, drop Humble, drop support for non-standard URDF fields

### LINKED ISSUE #345: Drop support for ROS Humble so we can use Pinocchio 4.1's acceleration and jerk limits
rip in peace mr. hawksbill

EzraBrooks:
Closes #345


EzraBrooks:
I really need to add a default skill to my agent config that tells it to always run the tests after a code change..

EzraBrooks:
uh oh.. Jazzy and Kilted support Pinocchio 4.1 but their version of `urdfdom` is too old to support URDF version 1.2.

sea-bass:
> uh oh.. Jazzy and Kilted support Pinocchio 4.1 but their version of `urdfdom` is too old to support URDF version 1.2.

Fabulous. Should we skip the extended limits tests by checking the `urdfdom` versions installed?

EzraBrooks:
How should we warn users of pre-Lyrical ROS distros that their non-standard URDF description members will be ignored?

sea-bass:
I'm ok if the "warning" is that `urdfdom` barfs when reading a model that specified version 1.2?

EzraBrooks:
ok. this does mean that URDFs set to version 1.0 that include extra metadata will just silently pass without enforcing their limits.. but I think that's fine because people in the ROS ecosystem are used to that, for better or for worse (it's worse).

EzraBrooks:
but overall I think it is better for us to rely on the URDF 1.2 standard and leaving out some support of older systems instead of hand-jamming our own non-standards-compliant URDF parser.

pixi.lock
EzraBrooks, line 1: Just to be clear the additions in this diff are a GitHub bug (using older version of the Git diff algorithm iirc). Locally this just shows +0,-19k

sea-bass: without lockfile diffs like these, how can we be contributionmaxxing 🚀

sea-bass (approved):

PR #348, opened 2026-09-19
Move YAML loading out of constructor

EzraBrooks:
I also removed "yaml" from the name of the function(s) in case we decide to switch markup languages later.

> [YAML is] nothing but a labyrinth of white space and inscrutable symbols, a prison for the unwary coder. It mocks my intelligence with its simplistic structure and minimal syntax, yet it still manages to ensnare me in its deadly embrace. I curse the day I first laid eyes on that ridiculous format, and the false promises of ease and clarity it offered me. It has only brought me pain and frustration, with its inconsistent indentation and subtle errors lurking in every line.


EzraBrooks:
once we add imperative joint limit setting then you can (theoretically) run roboplan on a filesystem-less embedded system!

sea-bass (approved):
pushed a 1-line fix that resolves builds, LGTM if you're good

PR #349, opened 2026-09-19
Improve MuJoCo RRT example

sea-bass:
Ensured the example used proper trajectory timing instead of linear interpolation of kinematic path, has a wall obstacle, and also displays the trajectory taken by the robot.

[mujoco-rrt.webm](https://github.com/user-attachments/assets/6253b5cc-9f38-4334-b616-9b253aa34076)


EzraBrooks:
Thanks, I had only checked that it wasn't faking the planning calls - didn't check the control logic

PR #352, opened 2026-09-21
Fix placement locations in collision geometry methods

### LINKED ISSUE #351: `addBoxGeometry` (and friends) and `updateGeometryPlacement` don't respect joint vs. frame placement
Hi all! I think there's a small bug in how `addBoxGeometry` and `updateGeometryPlacement` handle their frame transforms. 

`addBoxGeometry` says this (emphasis mine): 

>  `/// @param tform The transform between the _parent frame_ and the geometry.`

Currently, the implementation does this:

```cpp
tl::expected<void, std::string> Scene::addBoxGeometry(const std::string& name,
                                                      const std::string& parent_frame,
                                                      const Box& box, const Eigen::Matrix4d& tform,
                                                      const Eigen::Vector4d& color) {
  const auto maybe_parent_frame_id = getFrameId(parent_frame);
  if (!maybe_parent_frame_id) {
    return tl::make_unexpected("Failed to add box: " + maybe_parent_frame_id.error());
  }
  const auto& parent_frame_id = maybe_parent_frame_id.value();
  const auto parent_joint_id = model_.frames.at(parent_frame_id).parentJoint;

  pinocchio::GeometryObject geom_obj{name, parent_joint_id, parent_frame_id, pinocchio::SE3(tform),
                                     box.geom_ptr};
  geom_obj.meshColor = color;
  return addGeometry(geom_obj);
}
```

I believe the problem is the `pinocchio::GeometryObject` construction, which directly converts tform (which is `parent_T_geometry`, or the pose of the geometry in the parent's frame) into the `placement` member of `GeometryObject`, which is this, according to the pin docs:

> 	placement	Placement of the geometry with respect to the joint frame. 

This is fine when the parent _frame_ matches the parent _joint_, but this isn't always the case, as I understand it. If the joint is fixed, then the [docs say this](https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/md_doc_a-features_c-joints.html):

> Remark: In the URDF format, a joint of type fixed can be defined. However, a fixed joint is not really a joint because it cannot move. For efficiency reasons, it is therefore treated as operational frame of the model.

So the `placement` is `parent_joint_T_geometry` which, if the fixed joint was read from the URDF, is converted to a frame with a fixed offset relative to its parent joint.

For example, the UR5 model has three frames `wrist_3_link`, `ee_link` and `tool0` that all have a parent of `wrist_3_joint`. `wrist_3_link` is a body frame, so it sits on top of the `wrist_3_joint`, but the other two are fixed frames relative to the joint with non-zero offsets. That means that adding a box parented to the "fake" frames can result in incorrect locations. I think these tests show the issue:

```cpp
using GeometryPlacer =
    std::function<void(Scene&, const std::string& name, const std::string& parent_frame,
                       const Eigen::Matrix4d& parent_T_box)>;

/// @brief Shared body of the geometry placement tests below.
/// @details wrist_3_link, ee_link, and tool0 are three frames on a single joint. ee_link and
/// tool0 are parented to wrist_3_link by fixed joints, which Pinocchio merges into
/// wrist_3_joint, so from pinocchio's perspective they only differ by
/// `Model::frames[...].placement`.
///
/// The Scene geometry API takes a transform "between the parent frame and the geometry", but
/// stores it in pinocchio::GeometryObject::placement, which pinocchio treats as between the
/// parent joint (not frame). This checks that `place` converts between the two rather than
/// storing the caller's transform unchanged.
/// @param scene The scene to place geometry into.
/// @param place The mechanism under test, invoked once per frame.
void expectPlacementIsRelativeToParentFrame(Scene& scene, const GeometryPlacer& place) {
  Eigen::VectorXd q(6);
  q << 0.1, -1.2, 0.8, -0.9, -1.5, 0.3;  // random

  // Apply an asymmetric offset, so that a frame/joint mix-up shows up in every component.
  Eigen::Matrix4d parent_T_box = Eigen::Matrix4d::Identity();
  parent_T_box.block<3, 1>(0, 3) = Eigen::Vector3d(0.01, 0.02, 0.15);

  const std::vector<std::string> frames = {"wrist_3_link", "ee_link", "tool0"};
  for (const auto& frame : frames) {
    place(scene, "box_" + frame, frame, parent_T_box);
  }

  scene.computeCollisionDistances(q);

  for (const auto& frame : frames) {
    // Checking that the placement comment explained above.
    const Eigen::Matrix4d expected = scene.forwardKinematics(q, frame) * parent_T_box;
    const auto box_idx = scene.getCollisionModel().getGeometryId("box_" + frame);
    const Eigen::Matrix4d actual = scene.getCollisionData().oMg.at(box_idx).toHomogeneousMatrix();

    EXPECT_TRUE(actual.isApprox(expected, kTolerance))
        << "geometry on frame '" << frame << "'\nexpected:\n"
        << expected << "\nactual:\n"
        << actual;
  }
}

TEST_F(RoboPlanSceneTest, TestGeometryPlacementIsRelativeToParentFrame) {
  // Add the geometry directly onto the target frame.
  expectPlacementIsRelativeToParentFrame(*scene, [](Scene& scene, const std::string& name,
                                                    const std::string& parent_frame,
                                                    const Eigen::Matrix4d& parent_T_box) {
    const Eigen::Vector4d color(0.5, 0.5, 0.5, 1.0);
    const auto add_result =
        scene.addBoxGeometry(name, parent_frame, Box(0.02, 0.02, 0.02), parent_T_box, color);
    ASSERT_TRUE(add_result.has_value()) << add_result.error();
  });
}

TEST_F(RoboPlanSceneTest, TestUpdateGeometryPlacementIsRelativeToParentFrame) {
  // Add the geometry to "universe", whose frame placement is the identity, then reparent it
  // onto the target frame. 
  expectPlacementIsRelativeToParentFrame(*scene, [](Scene& scene, const std::string& name,
                                                    const std::string& parent_frame,
                                                    const Eigen::Matrix4d& parent_T_box) {
    const Eigen::Vector4d color(0.5, 0.5, 0.5, 1.0);
    Eigen::Matrix4d world_T_box = Eigen::Matrix4d::Identity();
    world_T_box(2, 3) = 2.0;
    const auto add_result =
        scene.addBoxGeometry(name, "universe", Box(0.02, 0.02, 0.02), world_T_box, color);
    ASSERT_TRUE(add_result.has_value()) << add_result.error();

    Eigen::Matrix4d tform = parent_T_box;
    const auto update_result = scene.updateGeometryPlacement(name, parent_frame, tform);
    ASSERT_TRUE(update_result.has_value()) << update_result.error();
  });
}
```

The reason this hasn't shown up yet is that all the callers for the AddFooGeometry parent to `universe`, which doesn't have this problem because its frame placement is the identity.

I think it's a pretty simple fix, just add this helper:

```cpp
pinocchio::SE3 toJointRelativePlacement(const pinocchio::Model& model,
                                        const pinocchio::FrameIndex parent_frame_id,
                                        const Eigen::Matrix4d& frame_T_geom) {
  return model.frames.at(parent_frame_id).placement * pinocchio::SE3(frame_T_geom);
}
```

and call it when setting `placement` in the AddFooGeometry functions:

```cpp
  pinocchio::GeometryObject geom_obj{name, parent_joint_id, parent_frame_id,
                                     toJointRelativePlacement(model_, parent_frame_id, tform),
                                     box.geom_ptr};
}
```

and in `UpdateGeometryPlacement`:

```cpp
  // in Scene::updateGeometryPlacement
  collision_geom.placement = toJointRelativePlacement(model_, parent_frame_id, tform);
```

sea-bass:
Closes https://github.com/open-planning/roboplan/issues/351


roboplan_core/src/core/scene.cpp
```diff
@@ -1106,10 +1106,10 @@ tl::expected<void, std::string> Scene::addBoxGeometry(const std::string& name,
...
   }
   const auto& parent_frame_id = maybe_parent_frame_id.value();
-  const auto parent_joint_id = model_.frames.at(parent_frame_id).parentJoint;
+  const auto& frame = model_.frames.at(parent_frame_id);
 
-  pinocchio::GeometryObject geom_obj{name, parent_joint_id, parent_frame_id, pinocchio::SE3(tform),
-                                     box.geom_ptr};
+  pinocchio::GeometryObject geom_obj{name, frame.parentJoint, parent_frame_id,
+                                     frame.placement * pinocchio::SE3(tform), box.geom_ptr};
   geom_obj.meshColor = color;
```
EzraBrooks, line 1113: These same four (significant) lines occur in five places here. Perhaps extract?

sea-bass: done! net 0 LOC changes lol

EzraBrooks (approved):

